PR #63, opened 2014-12-10
Fix replicator ID uniqueness:

tomblench:
- Source and Destination are inputs to the hash
- Filter and Query string are inputs to the hash for pull
  replications, if set

This implementation of replicator IDs is closer in spirit to the CDTDatastore one. Note that we don't need to include the 'push' flag as the input to the hash because source and destination are given, with their own keys.

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
+++ b/sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
@@ -25,11 +25,15 @@ import com.cloudant.sync.datastore.DocumentRevsList;
 import com.cloudant.sync.datastore.PreparedAttachment;
 import com.cloudant.sync.datastore.UnsavedStreamAttachment;
 import com.cloudant.sync.util.JSONUtils;
+import com.cloudant.sync.util.Misc;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Multimap;
 import com.google.common.eventbus.EventBus;
 
+import org.apache.commons.codec.binary.Hex;
+
+import java.io.ByteArrayInputStream;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -377,12 +381,18 @@ class BasicPullStrategy implements ReplicationStrategy {
         return changesProcessed;
     }
 
-    private String getReplicationId() {
-        if(filter == null) {
-            return this.sourceDb.getIdentifier() ;
-        } else {
-            return this.sourceDb.getIdentifier() + "?" + filter.toQueryString();
+    public String getReplicationId() {
+        HashMap<String, String> dict = new HashMap<String, String>();
+        dict.put("source", this.sourceDb.getIdentifier());
+        dict.put("target", this.targetDb.getIdentifier());
+        if(filter != null) {
+            dict.put("filter", this.filter.toQueryString());
         }
+        // get raw SHA-1 of dictionary
+        byte[] sha1Bytes = Misc.getSha1(new ByteArrayInputStream(JSONUtils.serializeAsBytes(dict)));
+        // return SHA-1 as a hex string
+        byte[] sha1Hex = new Hex().encode(sha1Bytes);
+        return new String(sha1Hex);
     }
 
     private ChangesResultWrapper nextBatch() {
--- a/sync-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java
+++ b/sync-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java
@@ -24,11 +24,16 @@ import com.cloudant.sync.datastore.BasicDocumentRevision;
 import com.cloudant.sync.datastore.DocumentRevisionTree;
 import com.cloudant.sync.datastore.MultipartAttachmentWriter;
 import com.cloudant.sync.datastore.RevisionHistoryHelper;
+import com.cloudant.sync.util.JSONUtils;
+import com.cloudant.sync.util.Misc;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import com.google.common.eventbus.EventBus;
 
+import org.apache.commons.codec.binary.Hex;
+
+import java.io.ByteArrayInputStream;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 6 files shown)


sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
```diff
@@ -377,12 +381,18 @@ private int processOneChangesBatch(ChangesResultWrapper changeFeeds)
...
+        dict.put("source", this.sourceDb.getIdentifier());
+        dict.put("target", this.targetDb.getDbCore().getPublicIdentifier());
+        if(filter != null) {
+            dict.put("filter", this.filter.toQueryString());
         }
+        // get raw SHA-1 of dictionary
+        byte[] sha1Bytes = Misc.getSha1(new ByteArrayInputStream(JSONUtils.serializeAsBytes(dict)));
+        // return SHA-1 as a hex string
+        byte[] sha1Hex = new Hex().encode(sha1Bytes);
+        return new String(sha1Hex);
```
mikerhodes: This looks right. Is this the same as the iOS code, out of curiousity?

sync-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java
```diff
@@ -312,14 +317,25 @@ private ItemsToPush missingRevisionsToJsonDocs(
         return allOpenRevisions;
     }
 
+    public String getReplicationId() {
+        HashMap<String, String> dict = new HashMap<String, String>();
+        dict.put("source", this.sourceDb.getIdentifier());
```
mikerhodes: In pull, you use `this.targetDb.getDbCore().getPublicIdentifier()` for the local identifier, why not here?

sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
tomblench: It's in the same spirit but not an exact copy. We use "source" and "target" rather than "localUUID" and "remoteURL".

The iOS code is this:
```
    NSMutableDictionary* spec =
        $mdict({ @"localUUID", _db.privateUUID }, { @"remoteURL", _remote.absoluteString },
               { @"push", @(self.isPush) }, { @"filter", _filterName },
               { @"filterParams", _filterParameters });
    return TDHexSHA1Digest([TDCanonicalJSON canonicalData:spec]);
```

sync-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java
tomblench: Fixed that - in pull it now uses the getIdentifier() convenience method which is equivalent

sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
mikerhodes: Thoughts as to why they include `push` in the dict?

mikerhodes: Oh, I see. When you use the local/remote rather than source/target as the identifiers, you need to include push to differentiate between the push and the pull case, whereas with source/target you get that automatically.

sync-core/src/test/java/com/cloudant/sync/replication/ReplicatorIdTest.java
```diff
@@ -0,0 +1,112 @@
...
+ * Created by tomblench on 10/12/14.
+ */
+public class ReplicatorIdTest extends ReplicationTestBase {
+
+    // source/target switched, so ids should be different
+    @Test
+    public void pullNotEqualToPush() throws URISyntaxException {
+        PullReplication pull = this.createPullReplication();
+        PushReplication push = this.createPushReplication();
+        Assert.assertNotEquals(pull.createReplicationStrategy().getReplicationId(), push.createReplicationStrategy().getReplicationId());
```
mikerhodes: Run the formatter on this file to get rid of the long lines.

sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
tomblench: I think they include push because you can't infer the direction from local/remote. We do source/target so you can infer whether it's a push because the source would be a database UUID and the target would be an http URL.

I can change it to match the iOS dictionary if you think it makes more sense.

sync-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java
mikerhodes: I don't think it matters, let's leave it.

sync-core/src/test/java/com/cloudant/sync/replication/BasicPushStrategyTest2.java
```diff
@@ -39,6 +39,9 @@
     String id2;
     String id3;
 
+    BasicPushStrategy push;
+    BasicPullStrategy pull;
```
mikerhodes: For me, `push()` and `pull()` should return the `Strategy` objects, rather than bunging them in class level variables, as they are created within a test method (as opposed to `setUp`).

sync-core/src/test/java/com/cloudant/sync/replication/ReplicatorIdTest.java
```diff
@@ -0,0 +1,119 @@
...
+        PushReplication push2 = this.createPushReplicationDifferingTarget();
+        Assert.assertNotEquals(push1.createReplicationStrategy().getReplicationId(),
+                push2.createReplicationStrategy().getReplicationId());
+    }
+
+    // two pull reps with differing source not equal
+    @Test
+    public void pullsDifferingSourceNotEqual() throws URISyntaxException {
+        PullReplication pull1 = this.createPullReplication();
+        PullReplication pull2 = this.createPullReplicationDifferingSource();
```
mikerhodes: I think the differing properties should be made explicit in the test rather than hidden in methods like `DifferingSource()`. The abstraction makes it more difficult to read the tests as information relevant to the test is somewhere outside the test method.

```java
    @Test
    public void pullsDifferingSourceNotEqual() throws URISyntaxException {
        PullReplication pull1 = this.createPullReplication();
        PullReplication pull2 = this.createPullReplication();

        pull1.source = new URI("http://blah.com/");
        pull2.source = new URI("http://blah2.com/");

        Assert.assertNotEquals(pull1.createReplicationStrategy().getReplicationId(),
                pull2.createReplicationStrategy().getReplicationId());
    }
```

... and similarly for the other tests.

sync-core/src/test/java/com/cloudant/sync/replication/ReplicatorIdTest.java
```diff
@@ -0,0 +1,119 @@
+package com.cloudant.sync.replication;
```
mikerhodes: Licence header

mikerhodes:
I think this looks great. Squash and +1.

PR #73, opened 2015-01-05
Add logging messages

rhyshort:
- Log HTTP Request status and headers
- Log successful insertions of revisions during replication

FB 42694 also mentions logging the response content as well as the header for verbose logging level, From my research it is a pain to implement, any thoughts?

### DIFF
--- a/sync-core/src/main/java/com/cloudant/mazha/HttpRequests.java
+++ b/sync-core/src/main/java/com/cloudant/mazha/HttpRequests.java
@@ -28,7 +28,7 @@ import com.google.common.io.Resources;
 
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.io.IOUtils;
-import org.apache.http.HttpHost;
+import org.apache.http.Header;
 import org.apache.http.HttpRequest;
 import org.apache.http.HttpRequestInterceptor;
 import org.apache.http.HttpResponse;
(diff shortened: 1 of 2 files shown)


mikerhodes:
Be more specific in your commit subject on the changes.

tomblench:
@rhyshort this level of logging is probably sufficient. Logging response content might have side effects on performance even with logging switched off and it's normally time to break out Wireshark if you need to do that level of troubleshooting.

sync-core/src/main/java/com/cloudant/mazha/HttpRequests.java
```diff
@@ -226,10 +226,17 @@ private HttpResponse executeRequest(HttpRequestBase request) {
...
                 request.addHeader("Authorization", authHeaderValue);
             }
+            //log the request
+            logger.info(request.toString());
             HttpResponse response = httpClient.execute(request, context);
             validate(request, response);
+            logger.info(response.getStatusLine().toString());
+            for(Header h: response.getAllHeaders()){
+                logger.finer(h.toString());
+            }
```
mikerhodes: Perhaps wrap in a `logger.getLevel()` check to avoid the loop if the level is higher than `finer`?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -935,6 +935,7 @@ public void forceInsert(BasicDocumentRevision rev,
                 }
             }
             if (ok) {
+                logger.finer("Inserted revision: "+rev);
```
mikerhodes: Is there a logging method which takes a formatting string and the object so we can avoid the rev.toString and concat calls for most runs through this loop?

sync-core/src/main/java/com/cloudant/mazha/HttpRequests.java
rhyshort: I've opted to use `logger.isLoggable(Level.FINER)`  in 41adfeb I think it makes it easier to read

mikerhodes:
This looks good, fixup all the commits into 03dd0d5 and merge, +1.

PR #75, opened 2015-01-07
Move begin transaction statements to outside try block

rhyshort:
Move start of database transactions to outside of try block.


tomblench:
Looks like you've caught them all, +1

PR #76, opened 2015-01-07
Add gradle wrapper to files for build and update .travis.yml

rhyshort:
- Add gradle wrapper files for use with the build
- Update .travis.yml file to use gradlew instead of gradle


mikerhodes:
Do the contributing docs need updating to get everyone to use the wrapper? (and how to make sure IDEA/Studio is?)

rhyshort:
The import works exactly the same as far as I can tell, the command line examples should probably be updated to use gradlew

mikerhodes:
Yes, they need to be. Using `gradlew` seems to be the accepted way to do this.

CONTRIBUTING.md
```diff
@@ -8,7 +8,7 @@ Cloudant Sync - Android is written in Java and uses
 The main requirements are:
 
 * Java 1.6
-* Gradle 1.8+
+* Gradle 2.2.1+
```
mikerhodes: Is this still a requirement for most developers if they are instructed to use `gradlew`?

CONTRIBUTING.md
```diff
@@ -48,6 +48,8 @@ $ brew install gradle
 
 Other platforms TDB.
 
+You can also use the gradle wrapper to download the needed gradle distribution at build time.
+
```
mikerhodes: I suggest we remove the gradle section if we're moving to `gradlew`.

CONTRIBUTING.md
```diff
@@ -65,7 +67,7 @@ Follow the instructions provided on the android developer site.
 The project should build out of the box with:
 
 ```bash
-$ gradle build
+$ ./gradlew build
```
mikerhodes: The first time we use gradle, we should mention how to run it on Windows.

mikerhodes:
+1, after squashing.

PR #77, opened 2015-01-08
Add simple AND query find functionality

alfinkel:
This pull request replaces PR #74. The single valid commit from that PR was cherry picked from the 42942-query-simple-AND-query branch and placed into this new branch 42942-query-find-simple-AND.  That commit was commented on by @mikerhodes and the subsequent commit 1c1ba14 to this branch addressed all issues raised with the exception of the naming and content of the TranslatorState class.  See https://github.com/cloudant/sync-android/commit/38f37b43ec52273943c9a0b8301cdcc21e83f0f6#commitcomment-9185831.

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
@@ -1,0 +1,284 @@
+//  Copyright (c) 2014 Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+
+package com.cloudant.sync.query;
+
+import com.cloudant.sync.datastore.Datastore;
+import com.cloudant.sync.sqlite.Cursor;
+import com.cloudant.sync.sqlite.SQLDatabase;
+import com.cloudant.sync.util.DatabaseUtils;
+import com.google.common.collect.Sets;
+
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ *  Handles querying indexes for a given datastore.
+ */
+class QueryExecutor {
+
+    private final SQLDatabase database;
+    private final Datastore datastore;
+    private final ExecutorService queue;
+
+    private static final Logger logger = Logger.getLogger(QueryExecutor.class.getName());
+
+    /**
+     *  Constructs a new CDTQQueryExecutor using the indexes in 'database' to find documents from
+     *  'datastore'.
+     */
+    QueryExecutor(SQLDatabase database, Datastore datastore, ExecutorService queue) {
+        this.database = database;
+        this.datastore = datastore;
+        this.queue = queue;
+    }
+
+    /**
+     *  Execute the query passed using the selection of index definition provided.
+     *
+     *  The index definitions are presumed to already exist and be up to date for the
+     *  datastore and database passed to the constructor.
+     *
+     *  @param query query to execute.
+     *  @param indexes indexes to use (this method will select the most appropriate).
+     *  @param skip how many results to skip before returning results to caller
+     *  @param limit number of documents the result should be limited to
+     *  @param fields fields to project from the result documents
+     *  @param sortDocument document specifying the order to return results, null to have no sorting
+     *  @return the query result
+     */
+    public QueryResult find(Map<String, Object> query,
+                            final Map<String, Object> indexes,
+                            long skip,
+                            long limit,
+                            List<String> fields,
+                            final List<Map<String, String>> sortDocument) {
+        //
+        // Validate inputs
+        //
+
+        if (!validateSortDocument(sortDocument)) {
+            return null;  // validate logs the error if doc is invalid
+        }
+
+        fields = normaliseFields(fields);
+
+        if (!validateFields(fields)) {
+            return null;  // validate logs error message
+        }
+
+        // normalise and validate query by passing into the executors
+
+        query = QueryValidator.normaliseAndValidateQuery(query);
+
+        if (query == null) {
+            return null;
+        }
+
+        //
+        // Execute the query
+        //
+
+        Boolean[] indexesCoverQuery = new Boolean[]{ false };
+        final ChildrenQueryNode root = translateQuery(query, indexes, indexesCoverQuery);
+
+        if (root == null) {
+            return null;
+        }
+
+        Future<List<String>> result = queue.submit(new Callable<List<String>>() {
+            @Override
+            public List<String> call() throws Exception {
+                Set<String> docIdSet = executeQueryTree(root, database);
+                List<String> docIdList;
+
+                // sorting
+                if (sortDocument != null && !sortDocument.isEmpty()) {
+                    docIdList = sortIds(docIdSet, sortDocument, indexes, database);
+                } else {
+                    docIdList = docIdSet != null ? new ArrayList<String>(docIdSet) : null;
+                }
+
+                return docIdList;
+            }
+        });
+
+        List<String> docIds;
+        try {
+            docIds = result.get();
+        } catch (ExecutionException e) {
+            logger.log(Level.SEVERE, "Execution error encountered:", e);
+            return null;
+        } catch (InterruptedException e) {
+            logger.log(Level.SEVERE, "Execution interrupted error encountered:", e);
+            return null;
+        }
+
+        if (docIds == null) {
+            return null;
+        }
+
+        UnindexedMatcher matcher = matcherForIndexCoverage(indexesCoverQuery, query);
+
+        if (matcher != null) {
+            String msg = "Query could not be executed using indexes alone; falling back to ";
+            msg += "filtering documents themselves. This will be VERY SLOW as each candidate ";
+            msg += "document is loaded from the datastore and matched against the query selector.";
+            logger.log(Level.WARNING, msg);
+        }
+
+        return new QueryResult(docIds, datastore, fields, skip, limit, matcher);
+    }
+
+    protected ChildrenQueryNode translateQuery(Map<String, Object> query,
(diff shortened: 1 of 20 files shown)


sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -0,0 +1,386 @@
...
+                String msg = String.format("Expected single operator per predicate map, got %s",
+                                           predicate.toString());
+                logger.log(Level.SEVERE, msg);
+                return null;
+            }
+
+            String operator = (String) predicate.keySet().toArray()[0];
+
+            // $not specifies the opposite operator OR NULL documents be returned
+            if (operator.equals("not")) {
```
mikerhodes: `$not`?

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -0,0 +1,386 @@
...
+                   predicateValue instanceof Integer ||
+                   predicateValue instanceof Long ||
+                   predicateValue instanceof Short ||
+                   predicateValue instanceof Double ||
+                   predicateValue instanceof Float) {
+            sqlClause = String.format("%s %s", clausePrefix, predicateValue);
+        }
+
+        return sqlClause;
+    }
```
mikerhodes: I think `addPredicateValueToSqlClause` is completing the statement that the objc code generates [here][1], that is it returns something like `= "mike"`. In that case, the `String.format` statements in this should have placeholders (`?`) for the `predicateValue` rather than using string substitution.

[1]: https://github.com/cloudant/CloudantQueryObjc/blob/master/Pod/Classes/CDTQQuerySqlTranslator.m#L390

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
```diff
@@ -0,0 +1,112 @@
+//  Copyright (c) 2014 Cloudant. All rights reserved.
```
mikerhodes: Given this class is a no-op, does it need to be in this commit?

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
```diff
@@ -0,0 +1,282 @@
...
+        if (node instanceof AndQueryNode) {
+            Set<String> accumulator = null;
+
+            AndQueryNode andNode = (AndQueryNode) node;
+            for (QueryNode qNode: andNode.children) {
+                Set<String> childIds = executeQueryTree(qNode, db);
+                if (accumulator == null) {
+                    accumulator = new HashSet<String>(childIds);
+                } else {
+                    accumulator.addAll(childIds);
```
mikerhodes: This should be an [intersection](https://github.com/cloudant/CloudantQueryObjc/blob/5432ddc8b68dc1ca7fc45225d84529df3614066f/Pod/Classes/CDTQQueryExecutor.m#L232). Union is for `$or`.

sync-core/src/main/java/com/cloudant/sync/query/QueryNode.java
```diff
@@ -0,0 +1,19 @@
...
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+
+package com.cloudant.sync.query;
+
+interface QueryNode {
+
+    // This interface is used to provide a general categorization for all types of query nodes.
```
mikerhodes: These types of comments belong as class-level javadoc.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -0,0 +1,386 @@
...
+                                           TranslatorState state) {
+        // At this point we will have a root compound predicate, AND or OR, and
+        // the query will be reduced to a single entry:
+        // { "$and": [ ... predicates (possibly compound) ... ] }
+        // { "$or": [ ... predicates (possibly compound) ... ] }
+
+        ChildrenQueryNode root = null;
+        List<Object> clauses = new ArrayList<Object>();
+
+        if (query.keySet().toArray()[0].equals(AND)) {
```
mikerhodes: Why not use the fact that `Map.get(blah)` returns `null` for key not present, as the [objc code](https://github.com/cloudant/CloudantQueryObjc/blob/5432ddc8b68dc1ca7fc45225d84529df3614066f/Pod/Classes/CDTQQuerySqlTranslator.m#L123) does and you have elsewhere?

sync-core/src/test/java/com/cloudant/sync/query/QueryExecutorTest.java
```diff
@@ -0,0 +1,351 @@
...
+        Assert.assertEquals(3, queryResult.documentIds().size());
+        Assert.assertTrue(queryResult.documentIds().containsAll(checkList));
+        List<String> docCheckList = new ArrayList<String>();
+        for (DocumentRevision rev: queryResult) {
+            Assert.assertNotNull(rev.getId());
+            Assert.assertNotNull(rev.getBody());
+            docCheckList.add(rev.getId());
+        }
+        Assert.assertEquals(3, docCheckList.size());
+        Assert.assertTrue(docCheckList.containsAll(checkList));
```
mikerhodes: Don't duplicate the check that the right results are returned.

Instead, remove the check in lines 130 to 137 for all the tests here. Add a separate set of tests that check that the list of documents in `documentIds()` matches the returned documents. Then you have the same checking in far fewer lines of code:

1. The list returned by `documentIds()` is correct for lots of queries.
2. The list returned by `documentIds()` is the same as the results returned by the iterator.

sync-core/src/test/java/com/cloudant/sync/query/QueryExecutorTest.java
```diff
@@ -0,0 +1,351 @@
...
+        query.put("name", nameOperator);
+        Map<String, Object> petOperator = new HashMap<String, Object>();
+        petOperator.put("$eq", "cat");
+        query.put("pet", petOperator);
+        QueryResult queryResult = im.find(query);
+        Assert.assertNotNull(queryResult);
+        Assert.assertEquals(2, queryResult.size());
+        List<String> checkList = new ArrayList<String>();
+        checkList.add("mike12");
+        checkList.add("mike72");
```
mikerhodes: This would seem clearer as `List<String> checkList = Arrays.asList("mike12", "mike72");`

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
alfinkel: I'd like to leave it since it is already referenced in the code https://github.com/cloudant/sync-android/pull/77/files#diff-80ec4d1d12847da53428d6c14a88dbb7R139 .  I was adding TODO comments everywhere and I missed that here.  If its OK with you I'd like to add a TODO comment here and leave it in place since it will likely be the very next thing to be worked on.  The alternative is to remove the code referenced above.

mikerhodes: I guess as this is a porting piece of work it's fine to leave in.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: This is fixed in 186736e

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: Yeah this was something I planned on addressing (fixing) at some point.  Initially, I diverged from the iOS implementation and used string substitution instead of placeholders and values when constructing SQL.  It is all fixed now.  SQL construction is being handled through the use of a SqlParts class similar to the iOS implementation.  Fixed in 186736e.

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
alfinkel: This was a mistake on my part and is fixed in 186736e. Unfortunately, I cannot write a test for the condition above yet.  The SQL translator needs to be completed so that it can handle multi-level queries.  As of now the query will fail with null during query translation and won't reach this section of code.

sync-core/src/main/java/com/cloudant/sync/query/QueryNode.java
alfinkel: Cleaned up class-level javadoc throughout index/query in 186736e.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: Makes sense.  Done in 186736e.

sync-core/src/test/java/com/cloudant/sync/query/QueryExecutorTest.java
alfinkel: Added a test to validate the iterator against documentIds and removed same validation from all other tests.  Done in 186736e.

sync-core/src/test/java/com/cloudant/sync/query/QueryExecutorTest.java
alfinkel: Done throughout index/query tests in 186736e

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
alfinkel: I had to make a small fix since in the previous commit version the accumulator in the else block was not actually being set to the intersection of itself and the childIds.  https://github.com/cloudant/sync-android/commit/f0ea756ce7ce7131381fc8d84c8b0f6ac16ff57c#diff-80ec4d1d12847da53428d6c14a88dbb7L208
Like mentioned before I should be able to exercise this code once the SQL translator class is capable of handling more complex queries.  Fixed in f0ea756.

mikerhodes:
This looks good to me. fixup the PR response commits into the original commit and merge. +1.

sync-core/src/test/java/com/cloudant/sync/query/QueryExecutorTest.java
```diff
@@ -0,0 +1,283 @@
...
+    public void canQueryOverOneStringField() {
+        // query - { "name" : "mike" }
+        Map<String, Object> query = new HashMap<String, Object>();
+        query.put("name", "mike");
+        QueryResult queryResult = im.find(query);
+        Assert.assertNotNull(queryResult);
+        Assert.assertEquals(3, queryResult.size());
+        List<String> checkList = Arrays.asList("mike12", "mike34", "mike72");
+        Assert.assertEquals(3, queryResult.documentIds().size());
+        Assert.assertTrue(queryResult.documentIds().containsAll(checkList));
```
alfinkel: Done in f9ecc68.  Included in PR:  https://github.com/cloudant/sync-android/pull/80

Issue #51, opened 2014-11-21
Allow setting of custom HTTP headers for replications


alfinkel:
After some investigation I believe that we should be able to follow a similar paradigm as is done in CDTDatastore and detailed in [Task 36445][Task 36445] to implement the setting of custom headers for replications in sync-android.

Specifically:

- Update the abstract class [Replication.java][Replication.java] by adding an optionalHeaders map as a property and adding a method that gets/checks the contents of the map similar to dictionaryForReplicatorDocument: in [CDTAbstractReplication][CDTAbstractReplication].
- Update [HttpRequests.java][HttpRequests.java] by adding a requestHeaders map property and all necessary code to manage requestHeaders and set the the new headers on each request accordingly.
- Minor updates to [BasicPullStrategy.java][BasicPullStrategy.java], [BasicPushStrategy.java][BasicPushStrategy.java], [CouchClientWrapper.java][CouchClientWrapper.java], and [CouchClient.java][CouchClient.java] to accommodate the new change and make the optionalHeaders map available to HttpRequests.
- These proposed changes should have no impact on any other set of functionality outside of replication.

@mikerhodes, @tomblench, @gadamc let me know if this makes sense.  I can add greater detail as well if necessary.

[Task 36445]: https://cloudant.fogbugz.com/f/cases/36445/Expose-TDReplicator-requestHeaders-to-CDTPull-PushReplication-configuration-objects

[Replication.java]: https://github.com/cloudant/sync-android/blob/master/sync-core/src/main/java/com/cloudant/sync/replication/Replication.java

[CDTAbstractReplication]: https://github.com/cloudant/CDTDatastore/blob/master/Classes/common/CDTReplicator/CDTAbstractReplication.m

[HttpRequests.java]: https://github.com/cloudant/sync-android/blob/master/sync-core/src/main/java/com/cloudant/mazha/HttpRequests.java

[BasicPullStrategy.java]: https://github.com/cloudant/sync-android/blob/master/sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java

[BasicPushStrategy.java]: https://github.com/cloudant/sync-android/blob/master/sync-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java

[CouchClientWrapper.java]: https://github.com/cloudant/sync-android/blob/master/sync-core/src/main/java/com/cloudant/sync/replication/CouchClientWrapper.java

[...]

PR #78, opened 2015-01-08
Support custom http headers as a CouchConfig option

tomblench:


sync-core/src/main/java/com/cloudant/mazha/CouchConfig.java
```diff
@@ -144,6 +151,23 @@ public void setMaxConnections(int maxConnections) {
...
+        return customHeaders;
+    }
+
+    public void setCustomHeaders(Map<String, String> customHeaders) {
+        this.customHeaders = new HashMap<String, String>();
+        List<String> prohibitedHeaders = Arrays.asList("authorization", "www-authenticate",
+                "host", "connection", "content-type", "accept", "content-length");
+        for (Map.Entry<String, String> header : customHeaders.entrySet()) {
+            if (prohibitedHeaders.contains(header.getKey().toLowerCase())) {
+                throw new IllegalArgumentException("Bad optional HTTP header: " + header);
```
rhyshort: I think we should have a better exception message, it should state that the header is not allowed to be set by the user

sync-core/src/test/java/com/cloudant/mazha/HttpRequestsTest.java
```diff
@@ -57,6 +59,37 @@ public void get_resourceExist_success() throws Exception {
         Assert.assertTrue(s.contains(TEST_DB));
     }
 
+    @Test(expected = IllegalArgumentException.class)
+    public void prohibitedCustomHeaderFails() throws Exception {
+        CouchConfig customCouchConfig = getCouchConfig();
+        Map<String, String> customHeaders = new HashMap<String, String>();
+        customHeaders.put("WWW-Authenticate", "test");
```
mikerhodes: This should be a loop with all the prohibited ones.

sync-core/src/main/java/com/cloudant/mazha/HttpRequests.java
```diff
@@ -110,11 +112,29 @@ public int getMaxForRoute(HttpRoute route) {
...
+                }
+            } else {
+                if (authHeaderValue != null) {
+                    // if we're doing basic auth, make a copy of the custom headers and add it in
+                    this.requestHeaders = new HashMap<String, String>();
+                    customHeaders.putAll(customHeaders);
+                    customHeaders.put("authorization", authHeaderValue);
+                }
+            }
+        }
```
mikerhodes: It looks like here the Authorization header will not be set if there are no already-existing custom headers (`customHeaders != null`).

mikerhodes:
Looks good, squash and merge.

tomblench:
Fixes #51

PR #79, opened 2015-01-09
26049 run on api 14 android

rhyshort:
Code changes to ensure tests can run against all android platform's we support.

These are:

Lower the minimum API level to 14 for the AndroidTest app
Add the system time in mills as a suffix for remote db names to avoid collisions

### DIFF
--- a/AndroidTest/app/build.gradle
+++ b/AndroidTest/app/build.gradle
@@ -13,7 +13,7 @@ android {
 
     defaultConfig {
         applicationId "cloudant.com.androidtest"
-        minSdkVersion 15
+        minSdkVersion 14
         targetSdkVersion 20
         versionCode 1
         versionName "1.0"
@@ -88,6 +88,8 @@ dependencies {
         // the current version of objenesis which mockito depends on is buggy on android, so force a newer version
         force true
     }
+
+    compile 'com.google.code.findbugs:jsr305:3.0.0' //this is needed for some versions of android
 
 }
 
--- a/sync-core/src/test/java/com/cloudant/mazha/CouchClientTestBase.java
+++ b/sync-core/src/test/java/com/cloudant/mazha/CouchClientTestBase.java
@@ -23,7 +23,7 @@ import org.junit.Before;
 
 public abstract class CouchClientTestBase {
 
-    static final String TEST_DB = "mazha-test";
+    String TEST_DB = "mazha-test"+System.currentTimeMillis();
 
     public static final Boolean TEST_WITH_CLOUDANT = Boolean.valueOf(
             System.getProperty("test.with.cloudant",Boolean.FALSE.toString()));
--- a/sync-core/src/test/java/com/cloudant/sync/replication/CompactedDBReplicationTest.java
+++ b/sync-core/src/test/java/com/cloudant/sync/replication/CompactedDBReplicationTest.java
@@ -81,7 +81,7 @@ public class CompactedDBReplicationTest extends ReplicationTestBase {
         CouchDbInfo info = couchClient.getDbInfo(getDbName());
 
         while(info.isCompactRunning()) {
-            Thread.sleep(10);
+            Thread.sleep(1000);
             info = couchClient.getDbInfo(getDbName());
         };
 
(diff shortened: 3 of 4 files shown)


sync-core/src/test/java/com/cloudant/sync/replication/CompactedDBReplicationTest.java
```diff
@@ -81,7 +81,7 @@ public void replicationFromCompactedDB() throws Exception{
         CouchDbInfo info = couchClient.getDbInfo(getDbName());
 
         while(info.isCompactRunning()) {
-            Thread.sleep(10);
+            //Thread.sleep() causes the tests to hang for long periods
```
tomblench: @rhyshort we have other bits of test code which use Thread.sleep(), and RetriableTask uses it. So I wonder why it works in those places and not here?

AndroidTest/app/build.gradle
```diff
@@ -89,6 +89,8 @@ dependencies {
         force true
     }
 
+    compile 'com.google.code.findbugs:jsr305:3.0.0' //this is needed for some versions of android
```
mikerhodes: Can this just be enabled for the test target?

sync-core/src/test/java/com/cloudant/sync/replication/CompactedDBReplicationTest.java
```diff
@@ -97,6 +98,7 @@ public void replicationFromCompactedDB() throws Exception{
         Assert.assertEquals(Replicator.State.STARTED, replicator.getState());
 
         while(replicator.getState() != Replicator.State.COMPLETE) {
+            //Thread.sleep(1000); this fails on older versions of android,
```
mikerhodes: Comment out of date?

sync-core/src/test/java/com/cloudant/sync/replication/ReplicationTestBase.java
```diff
@@ -37,6 +37,8 @@
     protected CouchClientWrapper remoteDb = null;
     protected CouchClient couchClient = null;
 
+    private long randomDBSuffix = System.currentTimeMillis();
```
mikerhodes: This isn't a random suffix right now. Why not just actually generate a random suffix rather than relying on the current system time?

sync-core/src/test/java/com/cloudant/mazha/CouchClientTestBase.java
```diff
@@ -23,7 +23,7 @@
 
 public abstract class CouchClientTestBase {
 
-    static final String TEST_DB = "mazha-test";
+    String TEST_DB = "mazha-test"+System.currentTimeMillis();
```
mikerhodes: Again, generate a proper random suffix here.

sync-core/src/test/java/com/cloudant/sync/replication/ReplicationTestBase.java
rhyshort: After talking with @tomblench we came to conclusion that the system time was unlikely to be the same for the tests running in parallel in jenkins.

mikerhodes: Okay, well at least correct the variable name and give it a comment on why it's here and is ~safe.

mikerhodes:
@rhyshort can you address my comments here so we can merge?

rhyshort:
oops, I didn't notice your last comment, I'll get on that

AndroidTest/app/build.gradle
rhyshort: Nope, the test app uses standard compilation and brings in all the tests classes that way

mikerhodes:
Your PR says we're targeting API level 14 but your commit subject mentioned level 15 -- is that right?

rhyshort:
oops the headline for that commit is a typo it should be 14. There a few issues with running on 14 and currently experiencing them on 15, both are issues with the emulators from what I can tell

mikerhodes:
So it looks like we're basically done if you rewrite the commit message to be correct? If so, +1 once the squashing and correction is done.

rhyshort:
Yup that's correct, I'll go do that now

PR #81, opened 2015-01-14
Add unindexed matcher capability to query

alfinkel:
This pull request provides simple, single level AND-able post hoc matcher functionality to complement the current set of Cloudant query SQL engine matching functionality.

@mikerhodes - points of interest in this PR are:

- The "match" implementation in QueryResult 
https://github.com/cloudant/sync-android/commit/f59f7bf3f8432fd0a1c51ced7286b03f6b8ce0f0#diff-5f5fe154c6c97af9fc67a7b9f8ae9c87R39

- The query rejection in SQLQueryTranslator when floats are found in a query object.  
https://github.com/cloudant/sync-android/commit/f59f7bf3f8432fd0a1c51ced7286b03f6b8ce0f0#diff-10b697b5119c2ac1b73822d862ec16dcR44
https://github.com/cloudant/sync-android/commit/f59f7bf3f8432fd0a1c51ced7286b03f6b8ce0f0#diff-10b697b5119c2ac1b73822d862ec16dcR172

The rest of the code is pretty much a straight port from iOS.

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/OperatorExpressionNode.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/OperatorExpressionNode.java
@@ -1,0 +1,31 @@
+//  Copyright (c) 2015 Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+
+package com.cloudant.sync.query;
+
+import java.util.Map;
+
+/**
+ *  An OperatorExpressionNode object is used by the
+ *  {@link com.cloudant.sync.query.UnindexedMatcher}.  Since some methods within the
+ *  {@link com.cloudant.sync.query.UnindexedMatcher} class that utilize an OperatorExpressionNode
+ *  object are static in nature, the OperatorExpressionNode class cannot be implemented as an
+ *  inner class to {@link com.cloudant.sync.query.UnindexedMatcher} and must be
+ *  implemented this way instead.
+ *
+ *  @see com.cloudant.sync.query.UnindexedMatcher
+ */
+class OperatorExpressionNode implements QueryNode {
+
+    public Map<String, Object> expression;
+
+}
--- a/sync-core/src/main/java/com/cloudant/sync/query/QueryResult.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QueryResult.java
@@ -14,37 +14,37 @@ package com.cloudant.sync.query;
 
 import com.cloudant.sync.datastore.BasicDocumentRevision;
 import com.cloudant.sync.datastore.Datastore;
+import com.cloudant.sync.datastore.DocumentRevision;
 import com.google.common.collect.Lists;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.Iterator;
-import java.util.LinkedList;
 import java.util.List;
+import java.util.NoSuchElementException;
 
 /**
- *  Iterable result of a query executed with {@link com.cloudant.sync.query.IndexManager}.
+ *  Iterable result of a query executed with {@link IndexManager}.
  *
- *  @see com.cloudant.sync.query.IndexManager
+ *  @see IndexManager
  */
-public class QueryResult implements Iterable<BasicDocumentRevision> {
+public class QueryResult implements Iterable<DocumentRevision> {
 
     private final static int DEFAULT_BATCH_SIZE = 50;
 
-    private final List<String> docIds;
+    private final List<String> originalDocIds;
     private final Datastore datastore;
     private final List<String> fields;
     private final long skip;
     private final long limit;
     private final UnindexedMatcher matcher;
 
-    public QueryResult(List<String> docIds,
+    public QueryResult(List<String> originalDocIds,
                        Datastore datastore,
                        List<String> fields,
                        long skip,
                        long limit,
                        UnindexedMatcher matcher) {
-        this.docIds = docIds;
+        this.originalDocIds = originalDocIds;
         this.datastore = datastore;
         this.fields = fields;
         this.skip = skip;
@@ -55,109 +55,129 @@ public class QueryResult implements Iterable<BasicDocumentRevision> {
     /**
      *  Returns the number of documents in this query result.
      *
-     *  @return the number of the {@code DocumentRevision} in this query result
+     *  @return the number of documents {@code DocumentRevision} in this query result.
      */
     public long size() {
-        // TODO - skip, limit w.r.t. size
-
-        return docIds.size();
+        return documentIds().size();
     }
 
     /**
      *  Returns a list of the document ids in this query result.
      *
+     *  This method is implemented this way to ensure that the list of document ids is
+     *  consistent with the iterator results.
+     *
      *  @return list of the document ids
      */
     public List<String> documentIds() {
-        // TODO - skip, limit w.r.t. document id list
-
-        return docIds;
+        List<String> documentIds = new ArrayList<String>();
+        List<DocumentRevision> docs = Lists.newArrayList(iterator());
+        for (DocumentRevision doc : docs) {
+            documentIds.add(doc.getId());
+        }
+        return documentIds;
     }
 
     @Override
-    public Iterator<BasicDocumentRevision> iterator() {
+    public Iterator<DocumentRevision> iterator() {
+        return new QueryResultIterator();
+    }
 
-        /**
-         * Partitions a set of document IDs into batches of DocumentRevision
-         * objects, and provides an iterator over the whole, un-partitioned set
-         * of revision objects (as if they were not batched).
-         */
-        return new Iterator<BasicDocumentRevision>() {
+    private class QueryResultIterator implements Iterator<DocumentRevision> {
 
-            // TODO - skip, limit, field projection, apply post-hoc matcher
+        private Range range;
+        private int nSkipped;
+        private int nReturned;
+        private boolean limitReached;
+        private Iterator<DocumentRevision> documentBlock;
 
-            /** List containing lists of partitions document IDs */
-            private final List<List<String>> subLists = this.partition(docIds, DEFAULT_BATCH_SIZE);
-            /** The current partition's iterator of document objects */
-            private Iterator<BasicDocumentRevision> subIterator = null;
+        private QueryResultIterator() {
+            range = new Range(0, DEFAULT_BATCH_SIZE);
+            nSkipped = 0;
+            nReturned = 0;
+            limitReached = false;
+            documentBlock = populateDocumentBlock();
+        }
 
-            @Override
-            public boolean hasNext() {
-                if(subIterator == null) {
-                    return subLists.size() > 0;
-                } else {
-                    return this.subIterator.hasNext() || subLists.size() > 0;
-                }
+        @Override
+        public boolean hasNext() {
+            return documentBlock.hasNext() ||
+                   (!limitReached && range.location < originalDocIds.size());
+        }
+
+        @Override
+        public DocumentRevision next() {
+            if (!this.hasNext()) {
+                throw new NoSuchElementException();
             }
-
-            @Override
-            public BasicDocumentRevision next() {
-                if(subIterator == null || !subIterator.hasNext()) {
-                    List<String> ids = subLists.remove(0);
-                    subIterator = this.nextSubIterator(ids);
-                }
-                
-                return subIterator.next();
+            if (!documentBlock.hasNext()) {
+                documentBlock = populateDocumentBlock();
             }
+            return documentBlock.next();
+        }
 
-            @Override
-            public void remove() {
-                throw new UnsupportedOperationException();
-            }
+        @Override
+        public void remove() {
+            throw new UnsupportedOperationException();
+        }
 
-            /**
-             * Partition a list of document IDs into batches of batchSize.
-             *
-             * Return a mutable list of consecutive sublists.
-             * Same as Guava's "Lists.partition" except the result list is mutable.
-             * It is needed because this iterator removes sublist from the partitions
-             * as it goes.
-             *
-             * @see <a href="http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/
-             * common/collect/Lists.html#partition(java.util.List, int)">Lists.partition</a>
-             */
-            private List<List<String>> partition(List<String> documentIds, int batchSize) {
-                List<List<String>> partitions = Lists.partition(documentIds, batchSize);
-                List<List<String>> res = new LinkedList<List<String>>();
-                for(List<String> p : partitions) {
-                    res.add(p);
-                }
-                return res;
-            }
+        private Iterator<DocumentRevision> populateDocumentBlock() {
+            List<DocumentRevision> docList = new ArrayList<DocumentRevision>();
+            while (range.location < originalDocIds.size()) {
+                range.length = Math.min(DEFAULT_BATCH_SIZE, originalDocIds.size() - range.location);
+                List<String> batch = originalDocIds.subList(range.location,
+                                                            range.location + range.length);
+                List<BasicDocumentRevision> docs = datastore.getDocumentsWithIds(batch);
+                for (DocumentRevision rev : docs) {
+                    DocumentRevision innerRev;
+                    innerRev = rev;  // Allows us to replace later if projecting
 
-            /**
-             * Load the next partition of DocumentRevision objects for the
-             * iterator.
-             *
-             * @param ids the IDs of the revisions to load.
-             * @return an iterator over the DocumentRevision objects for `ids`.
-             */
-            private Iterator<BasicDocumentRevision> nextSubIterator(List<String> ids) {
-                HashMap<String, BasicDocumentRevision> map = new HashMap<String, BasicDocumentRevision>();
-                for(BasicDocumentRevision revision : datastore.getDocumentsWithIds(ids)) {
-                    map.put(revision.getId(), revision);
-                }
-                List<BasicDocumentRevision> revisions = new ArrayList<BasicDocumentRevision>(ids.size());
-                // return list of DocumentRevision that in the same order as input "ids"
-                for(String id : ids) {
-                    BasicDocumentRevision revision = map.get(id);
-                    if(revision != null ) {
-                        revisions.add(revision);
+                    // Apply post-hoc matcher
+                    if (matcher != null && !matcher.matches(innerRev)) {
+                        continue;
                     }
+
+                    // Apply skip (skip == 0 means disable)
+                    if (skip > 0 && nSkipped < skip) {
+                        nSkipped = nSkipped + 1;
+                        continue;
+                    }
+
+                    // TODO - Add projection logic
+
+                    docList.add(innerRev);
+
+                    // Apply limit (limit == 0 means disable)
+                    nReturned = nReturned + 1;
+                    if (limit > 0 && nReturned >= limit) {
+                        limitReached = true;
+                        break;
+                    }
+
                 }
-                return revisions.iterator();
+
+                range.location = range.location + range.length;
+
+                if (limitReached) {
+                    break;
+                }
+
+                if (!docList.isEmpty()) {
+                    break;
+                }
             }
-        };
+            return docList.iterator();
+        }
+    }
+
+    private class Range {
+        public int location;
+        public int length;
+
+        private Range(int location, int length) {
+            this.location = location;
+            this.length = length;
+        }
     }
 
 }
--- a/sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
@@ -1,4 +1,4 @@
-//  Copyright (c) 2014 Cloudant. All rights reserved.
+//  Copyright (c) 2015 Cloudant. All rights reserved.
 //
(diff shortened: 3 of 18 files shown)


alfinkel:
@mikerhodes when you get a chance can you take a look at this PR?  Thanks.

sync-core/src/main/java/com/cloudant/sync/query/QueryResult.java
```diff
@@ -44,23 +42,19 @@ public QueryResult(List<String> docIds,
                        long skip,
                        long limit,
                        UnindexedMatcher matcher) {
-        this.docIds = docIds;
+        this.docIds = matcher == null ? processDocIds(docIds, skip, limit) :
+                                        processDocIds(docIds, datastore, skip, limit, matcher);
```
mikerhodes: This processing needs to happen during the iteration process to avoid false positives/negatives as the contents of the document store could change between this point and the iterator.

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
```diff
@@ -98,14 +106,221 @@ public static UnindexedMatcher matcherWithSelector(Map<String, Object> selector)
...
+            // just flipping the result at the end.
+            //
+            // This is because { "$not": { "$eq": "white_cat" } } needs
+            // to be taken as an atomic check, meaning:
+            //   "there's an item in the array that matches `!= "white_cat"`"
+            // rather than:
+            //   "not (there's an item that matches white_cat)"
+            // The latter is satisfied using the $nin operator.
+            if (actual instanceof List) {
+                boolean currentItemPassed = false;
```
mikerhodes: Not used outside `for`, declare inside, avoid the compiler warning.

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
```diff
@@ -98,14 +106,221 @@ public static UnindexedMatcher matcherWithSelector(Map<String, Object> selector)
...
+    private boolean compareEq(Object l, Object r) {
+        if (l instanceof String && r instanceof String) {
+            return l.equals(r);
+        }
+        if (l instanceof Boolean && r instanceof Boolean) {
+            return l == r;
+        }
+        return l instanceof Number && r instanceof Number &&
+                !(l instanceof Float || r instanceof Float) &&
+                ((Number) l).longValue() == ((Number) r).longValue();
```
mikerhodes: Seems like this will compare `1.0` and `1.1` as equal. I think this needs making protected and some tests written.

sync-core/src/test/java/com/cloudant/sync/query/AbstractQueryExecutorTestBase.java
```diff
@@ -0,0 +1,280 @@
...
+ *  within this class using {@link com.cloudant.sync.query.QueryExecutorTest}
+ *  which exercises the SQL query engine matching functionality.  We then test
+ *  the same test execution pipeline contained within this class using
+ *  {@link com.cloudant.sync.query.QueryExecutorUnindexedTest} which in turn
+ *  exercises the post hoc matcher matching functionality.
+ */
+public abstract class AbstractQueryExecutorTestBase extends AbstractIndexTestBase {
+
+    @Override
+    public void setUp() throws SQLException {
```
mikerhodes: I don't think this works correctly, because you appear to create the same class of IndexManager object (in `AbstractIndexTestBase` `setUp()`) for all of the query test classes, which will use both SQL indexes and the unindexed matcher?

You need sub-classes of IndexManager specific to testing, which override:

* the `QueryTranslator#translateQuery()` to always return a query which matches all document (so only the unindexed matcher is being tested).
* the `QueryTranslator#matcherForIndexCoverage()` to always return `null` (so only the SQL query engine is tested).

sync-core/src/main/java/com/cloudant/sync/query/QueryResult.java
alfinkel: I re-wrote the QueryResult processing so that it more closely resembles what is being done in the iOS version.  I also added skip and limit testing since skip and limit are now implemented.  Done in 3f1b4c2.  

https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-5f5fe154c6c97af9fc67a7b9f8ae9c87L31
https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-1a5092ac6698384ba572368c278436fdR311

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
alfinkel: Done in 3f1b4c2.

https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-71032f1442a4b45142f92307d7418a22L236

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
alfinkel: You are absolutely right.  I changed Number comparison from .longValue() to .doubleValue() and added tests to verify conditions.  Done in 3f1b4c2.
https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-71032f1442a4b45142f92307d7418a22L284
https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-9845e00b021fbb4b1f9cd0cb43e43b23R59

sync-core/src/test/java/com/cloudant/sync/query/AbstractQueryExecutorTestBase.java
alfinkel: I changed the processing of these tests to now use mock IndexManager and QueryExecutor subclasses to force the tests to go in two directions.  Same set of tests are executed once against the SQL engine and once against the post hoc matcher.  Done in 3f1b4c2.
https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-1a5092ac6698384ba572368c278436fdL47
https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-0eaf4604dd4e002edfe0d41fdea2c7e3R20
https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-162b0c09f735c14ea05abac3dd234d9eR27
https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-c4ac637a25d1bf100c90b6b658737b56R20
https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-253a3500c1d06e06260c5b09832b8986R21
https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-c8d8ef5f4948719512f62057f5359599L29
https://github.com/cloudant/sync-android/commit/3f1b4c2992cab575b6d999ec536957bb10d04f57#diff-162fe9eab99c2b46b7150404a5da366dL28

alfinkel:
@mikerhodes please take a look when you have a chance at 3f1b4c2.  I believe that I addressed all of your comments for https://github.com/cloudant/sync-android/pull/81.

sync-core/src/main/java/com/cloudant/sync/query/OperatorExpressionNode.java
```diff
@@ -0,0 +1,32 @@
+//  Copyright (c) 2014 Cloudant. All rights reserved.
```
mikerhodes: It's now 2015 :)

sync-core/src/test/java/com/cloudant/sync/query/QueryExecutorMatcherTest.java
```diff
@@ -0,0 +1,39 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+
+import com.cloudant.sync.util.SQLDatabaseTestUtils;
+
+import java.sql.SQLException;
+import java.util.Arrays;
+
+public class QueryExecutorMatcherTest extends AbstractQueryTestBase {
```
mikerhodes: I think it's worth adding a comment to each of the four classes `MockSQLOnlyQueryExecutor`, `MockSQLOnlyIndexManager`, `MockMatcherQueryExecutor` and `MockMatcherIndexManager` to explain what each one does -- or you could point to explanations you add to `QueryExecutorMatcherTest` and `QueryExecutorSQLOnlyTest`.

This set up is rather odd -- perhaps we can figure out a way to clean up both the objc and this one sometime! -- and so needs documenting.

sync-core/src/test/java/com/cloudant/sync/query/UnindexedMatcherTest.java
```diff
@@ -0,0 +1,394 @@
...
+        assertThat(UnindexedMatcher.compareEq(2l, 1l), is(false));
+    }
+
+    @Test
+    public void compareEqLongToInt() {
+        assertThat(UnindexedMatcher.compareEq(1l, 1), is(true));
+        assertThat(UnindexedMatcher.compareEq(1, 1l), is(true));
+        assertThat(UnindexedMatcher.compareEq(1l, 2), is(false));
+        assertThat(UnindexedMatcher.compareEq(2, 1l), is(false));
+    }
```
mikerhodes: Do you think you need to do the boxing manually instead of relying on autoboxing for the compareEq tests? Given that `compareEq` will never autobox because all its inputs will come from dictionaries where the boxing has already happened? I'd say you want to be passing Integers and Doubles in, and ensuring they work, and passing in Floats and ensuring they don't.

I'd also statically import compareEq if that's possible.

sync-core/src/main/java/com/cloudant/sync/query/QueryResult.java
```diff
@@ -55,109 +54,132 @@ public QueryResult(List<String> docIds,
...
-            }
+                    // Apply skip (skip == 0 means disable)
+                    if (skip > 0 && nSkipped < skip) {
+                        nSkipped = nSkipped + 1;
+                        continue;
+                    }
+
+                    // Apply projection if result matches
+                    if (fields != null && !fields.isEmpty()) {
+                        // TODO - projection
```
mikerhodes: For things that are not yet done, I suggest throwing NotImplementedException. Same for the operators and so on.

mikerhodes: Though perhaps not here, now I recall where we are in the code. This TODO needs more context really. Or just remove this code until projection is implemented, I don't think it's needed even as a placeholder?

sync-core/src/main/java/com/cloudant/sync/query/QueryResult.java
```diff
@@ -55,109 +54,132 @@ public QueryResult(List<String> docIds,
...
+        List<String> documentIds = new ArrayList<String>();
+        List<BasicDocumentRevision> docs = Lists.newArrayList(iterator());
+        for (BasicDocumentRevision doc : docs) {
+            documentIds.add(doc.getId());
+        }
+        return documentIds;
     }
 
     @Override
     public Iterator<BasicDocumentRevision> iterator() {
```
mikerhodes: Public API should use `DocumentRevision` interface not the specific class.

sync-core/src/main/java/com/cloudant/sync/query/OperatorExpressionNode.java
```diff
@@ -0,0 +1,32 @@
...
+ *  An OperatorExpressionNode object is used by the
+ *  {@link com.cloudant.sync.query.UnindexedMatcher}.  Since some methods within the
+ *  {@link com.cloudant.sync.query.UnindexedMatcher} class that utilize an OperatorExpressionNode
+ *  object are static in nature, the OperatorExpressionNode class cannot be implemented as an
+ *  inner class to {@link com.cloudant.sync.query.UnindexedMatcher} and must be
+ *  implemented this way instead.
+ *
+ *  @see com.cloudant.sync.query.UnindexedMatcher
+ */
+public class OperatorExpressionNode implements QueryNode {
```
mikerhodes: Private

sync-core/src/test/java/com/cloudant/sync/query/AbstractQueryTestBase.java
```diff
@@ -105,13 +132,38 @@ public void setUp() throws SQLException {
...
+
+    @Test
+    public void returnsNullWhenQueryContainsInvalidValue() {
+        // query - { "name" : { "eq" : "mike" }, "age" : { "$eq" : 12.0f } }
+        Map<String, Object> nameOperator = new HashMap<String, Object>();
+        nameOperator.put("$eq", "mike");
+        Map<String, Object> query = new HashMap<String, Object>();
+        query.put("name", nameOperator);
+        Map<String, Object> ageOperator = new HashMap<String, Object>();
+        ageOperator.put("$eq", 12.0f);
```
mikerhodes: If `12.0f` is the invalid value (because it's a float) note that somewhere in the test.

sync-core/src/test/java/com/cloudant/sync/query/AbstractQueryTestBase.java
```diff
@@ -17,27 +17,55 @@
...
+ *  same as the SQL query engine.
+ *
+ *  To accomplish this, we test the entire test execution pipeline contained
+ *  within this class using {@link com.cloudant.sync.query.QueryExecutorSQLOnlyTest}
+ *  which exercises the SQL query engine matching functionality.  We then test
+ *  the same test execution pipeline contained within this class using
+ *  {@link com.cloudant.sync.query.QueryExecutorMatcherTest} which in turn
+ *  exercises the post hoc matcher matching functionality.
+ */
+public abstract class AbstractQueryTestBase {
```
mikerhodes: You still need a QueryExecutorTest class which tests the default implementations of IndexManager and QueryExecutor. Right now you only have the two classes which test each separately, you need the test for the use of both. The `QueryExecutorTest` might need some extra tests to cover the cases where a combination of both SQL and unindexed matcher is needed -- because the Abstract class probably only has tests which can be solved by either in isolation (depending on the indexes defined).

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
```diff
@@ -98,14 +106,221 @@ public static UnindexedMatcher matcherWithSelector(Map<String, Object> selector)
...
+        // AND and OR subclauses are handled identically whatever the parent is.
+        // We go through the query twice to order the OR clauses before the AND
+        // clauses, for predictability.
+        //
+
+        // Add subclauses that are OR
+        // TODO - Handle OR subclauses
+
+        // Add subclauses that are AND
+        // TODO - Handle AND subclauses
```
mikerhodes: I think the query should fail if the user passed in something with `and/or` subclauses -- or do you check for this in validation?

sync-core/src/main/java/com/cloudant/sync/query/OperatorExpressionNode.java
alfinkel: Updated all copyright notices on classes written in 2015.  Done in https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3

sync-core/src/test/java/com/cloudant/sync/query/QueryExecutorMatcherTest.java
alfinkel: Added comments where appropriate explaining what each class does.  I took a note to revisit the process later on in both iOS and Android once the smoke clears.
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-0eaf4604dd4e002edfe0d41fdea2c7e3R20
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-162b0c09f735c14ea05abac3dd234d9eR27
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-c4ac637a25d1bf100c90b6b658737b56R20
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-253a3500c1d06e06260c5b09832b8986R21
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-c8d8ef5f4948719512f62057f5359599R24
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-162fe9eab99c2b46b7150404a5da366dR24

sync-core/src/test/java/com/cloudant/sync/query/UnindexedMatcherTest.java
alfinkel: Added UnindexMatcher.compareEq as a static import to the test class and added manual boxing to all Numbers being used during the compare equality tests.

https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-9845e00b021fbb4b1f9cd0cb43e43b23R15
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-9845e00b021fbb4b1f9cd0cb43e43b23R61

sync-core/src/main/java/com/cloudant/sync/query/QueryResult.java
alfinkel: I believe that the resolution we came to today was to leave this PR as it is wrt the handling of unimplemented functionality and that in the next set of commits and subsequent new PR to account for unimplemented functionality more gracefully there.  With that said, I've removed the if placeholder logic for projection and replaced it with a single TODO comment.
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-5f5fe154c6c97af9fc67a7b9f8ae9c87L143

sync-core/src/main/java/com/cloudant/sync/query/QueryResult.java
alfinkel: Made QueryResult implement Iterable\<DocumentRevision> and used DocumentRevision throughout the code.  The exception being the case where we are documents are retrieved from the datastore since the datastore returns List\<BasicDocumentRevision>.
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-5f5fe154c6c97af9fc67a7b9f8ae9c87L29

sync-core/src/main/java/com/cloudant/sync/query/OperatorExpressionNode.java
alfinkel: I made the class package protected instead of public.  In order to make it private it would have to be an inner class.
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-cac498342a06ebddae3efd7af8bbbf5eL28
I did consider moving all of the QueryNode classes to be inner classes of the QueryExecutor.  @mikerhodes what are your thoughts on that?  Would it be better to leave them as top level classes or does it make sense to make them inner to the executor?  I go back and forth with what I think is right.

sync-core/src/test/java/com/cloudant/sync/query/AbstractQueryTestBase.java
alfinkel: Added a comment to the test description.
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-1a5092ac6698384ba572368c278436fdR157

sync-core/src/test/java/com/cloudant/sync/query/AbstractQueryTestBase.java
alfinkel: Added a `QueryExecutorStandardTest` which uses the regular `IndexManager` and `QueryExecutor`. 
https://github.com/cloudant/sync-android/commit/cf527fd540554a34338857f179c44fa358444ad3#diff-ce1fd18195cc9f32262108611d0dd922R25
As of right now though only simple AND query is supported so only either the matcher or the SQL engine is being used to satisfy the results of a query.  In the next set of commits and PR I will be providing support for OR and subclauses so there will be a need at that point to add more tests so that the "standard" way of querying is validated appropriately through `QueryExecutorStandardTest`.

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
alfinkel: Copying response from above:
I believe that the resolution we came to today was to leave this PR as it is wrt the handling of unimplemented functionality and that in the next set of commits and subsequent new PR to account for unimplemented functionality more gracefully there.

sync-core/src/main/java/com/cloudant/sync/query/QueryResult.java
```diff
@@ -55,109 +55,129 @@ public QueryResult(List<String> docIds,
...
-             */
-            private Iterator<BasicDocumentRevision> nextSubIterator(List<String> ids) {
-                HashMap<String, BasicDocumentRevision> map = new HashMap<String, BasicDocumentRevision>();
-                for(BasicDocumentRevision revision : datastore.getDocumentsWithIds(ids)) {
-                    map.put(revision.getId(), revision);
+                range.location = range.location + range.length;
+
+                if (stop) {
+                    // Limit reached.  Move range location to the end.
+                    range.location = originalDocIds.size();
```
mikerhodes: I'd not use this technique of pretending the end of the list has been reached when really you've reached the limit, instead I'd make `stop` class level and make the hasNext check include it: 

```java
if (stop) {
    return false;
} else {
    return (documentBlock.hasNext() || range.location < originalDocIds.size());
}
```

Perhaps renamed `stop` to `limitReached`.

alfinkel: Done.https://github.com/cloudant/sync-android/commit/beba6a1c5a2456f5dfdf0ff8f508d931d52ba34c#diff-5f5fe154c6c97af9fc67a7b9f8ae9c87R86

PR #83, opened 2015-01-15
Fix tests and put them into correct categories:

tomblench:
- Some tests now correctly marked RequireRunningCouchDB.class
- Other tests which don't require CouchDB re-written with mocks

`gradle sync-core:test` now runs correctly with CouchDB turned off. For strange gradle multi-project reasons, `gradle test` doesn't work and should probably be addressed in another ticket.


mikerhodes:
All looks fairly innocent, +1.

PR #84, opened 2015-01-20
Execute all datastore db operations on a single thread

rhyshort:
All database operations are performed on a single operation thread,
this stops leaking of connections when a user's application uses the
datastore on two different threads, but only calls close on a single
thread.

Remove threadlocal code from SQLiteWrapper, enforcing the need to perform
each database operation on the same thread.

Added a TestUtil method to create a connection to underlying database
for the datastore.

Remove test which checks if the underlying db is closed when
datastore.close() is called, this test can't be implemented since the
database can only be called on the executor service thread.

Note: Exception handling is not up to standard, the exceptions 
thrown by each method needs to be looked at in a later PR.


sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -184,15 +236,12 @@ public BasicDocumentRevision getDocument(String id) {
         return getDocument(id, null);
     }
 
-    @Override
-    public BasicDocumentRevision getDocument(String id, String rev) {
-        Preconditions.checkState(this.isOpen(), "Database is closed");
-        Preconditions.checkArgument(!Strings.isNullOrEmpty(id), "DocumentRevisionTree id can not be empty");
+    private BasicDocumentRevision getDocumentInBlock(String id, String rev){
```
mikerhodes: I'd prefer to see methods which require you to already be in a task to take the database as an argument -- they could even be static. This makes it harder to accidentally call them from outside a submitted task.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -184,15 +236,12 @@ public BasicDocumentRevision getDocument(String id) {
...
         Cursor cursor = null;
         try {
             String[] args = (rev == null) ? new String[]{id} : new String[]{id, rev};
             String sql = (rev == null) ? GET_DOCUMENT_CURRENT_REVISION : GET_DOCUMENT_GIVEN_REVISION;
-            cursor = this.sqlDb.rawQuery(sql, args);
+            cursor = sqlDb.rawQuery(sql, args);
             if (cursor.moveToFirst()) {
                 long sequence = cursor.getLong(3);
                 List<? extends Attachment> atts = attachmentManager.attachmentsForRevision(sequence);
```
mikerhodes: `attachmentsForRevision` calls the database for this datastore, and is able to because the database is a public property. It needs to not be, allowing the database object to escape makes it essentially impossible to maintain the serial nature of database calls we're trying to introduce.

That is, all access to the underlying SQL database needs to be mediated from a single class which ensures everything happens on a queue. It might be that a helper class is required given the database is accessed from various places.

rhyshort:
Updates:

Helper class: SQLDatabaseQueue  provides an SQLDatabase object when a SQLQueueCallable is being called. This removes the database from being stored in the Datastore. 

Pull Replication now uses a copy of the datastore object its replicating to. This stops replication transactions treading on the toes of user datastore transactions.

mikerhodes:
Core issue here is still that database access knowledge is distributed across classes -- e.g., it'd be good if the queue had an inTransaction-type method like FMDB's. Right now there's still a lot of coupling between classes to keep database access safe.

sync-core/src/test/java/com/cloudant/sync/util/TestUtils.java
```diff
@@ -153,5 +148,27 @@ public static File loadFixture(String fileName){
        }
    }
 
+   public static SQLDatabase getDatabaseConnectionToExsitingDb(SQLDatabase db){
```
tomblench: typo for 'existing'

sync-core/src/test/java/com/cloudant/sync/util/TestUtils.java
```diff
@@ -153,5 +148,27 @@ public static File loadFixture(String fileName){
...
+       if(Misc.isRunningOnAndroid())
+           return db;
+
+       try {
+           String filePath = (String) db.getClass()
```
tomblench: Why do we need to do this dynamic invoke here?

rhyshort: At the time of writing, I didn't want to make a test dependancy on JavaSE, since during the android builds java-se is not compiled, only core and android are

tomblench: Actually I don't even understand the purpose of this method, why do we need to open a new connection to an existing database (in test code), and why does it just return the current connection on android?

rhyshort: Yeah, I think it was because originally, when I removed the threadlocal code, the tests didn't have access to the internal queues, looking at it again, I should change the tests to run db actions on the internal queues.

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -93,30 +93,47 @@ public IndexManager(Datastore datastore) {
...
+
+                }
+            }).get();
+        } catch (InterruptedException e) {
+            logger.log(Level.SEVERE,"Failed to close db",e);
+            throw new RuntimeException(e);
+        } catch (ExecutionException e) {
+            logger.log(Level.SEVERE, "Failed to close db", e);
+        }
         queue.shutdown();
```
mikerhodes: Doesn't `queue.shutdown()` close the database -- is 122 - 135 needed?

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Submits a database task for execution
+     * @param callable The task to be performed
+     * @param <T> The type of object that is returned from the task
+     * @return
+     */
+    public <T> Future<T> submit(SQLQueueCallable<T> callable){
+        callable.setDb(db);
+        callable.setInTransaction(false);
```
mikerhodes: I would call the ivar and method `runInTransaction`, the method's giving an instruction and the current naming doesn't make that clear.

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLQueueCallable.java
```diff
@@ -0,0 +1,72 @@
...
+    private SQLDatabase db;
+    private boolean inTransaction = false;
+
+    @Override
+    final public T call() throws Exception {
+        if(inTransaction){
+            try {
+                db.beginTransaction();
+                T returned =  call(db);
+                db.setTransactionSuccessful();
```
mikerhodes: I think I'd add a short comment here that `call() throws if transaction should be aborted` so no one makes the mistake I made when originally reading that there's no way to fail a transaction.

sync-core/src/test/java/com/cloudant/sync/datastore/AttachmentTest.java
```diff
@@ -84,10 +85,18 @@ public void setBadAttachmentsTest() throws IOException, ConflictException, SQLEx
...
-            Cursor c2 = datastore.getSQLDatabase().rawQuery("select sequence from revs;", null);
-            Assert.assertEquals("Revs table not empty", c2.getCount(), 1);
+            //TODO This needs to happen on the queue inside of the datastore
+            datastore.runOnDbQueue(new SQLQueueCallable<Object>() {
+                @Override
+                public Object call(SQLDatabase db) throws Exception {
+                    Cursor c1 = db.rawQuery("select sequence from attachments;", null);
+                    Assert.assertEquals("Attachments table not empty", c1.getCount(), 0);
+                    Cursor c2 = db.rawQuery("select sequence from revs;", null);
+                    Assert.assertEquals("Revs table not empty", c2.getCount(), 1);
```
mikerhodes: Have you checked these asserts actually cause test failure if they happen on the queue? I've had problems in the past where Asserts need to happen on the same thread as the tests, but that might have been on iOS.

rhyshort: Since `ThreadLocal` variables have been removed unless we provide the test with a database connection for the main thread, we will get an `SQLiteException` thrown from the underlying database. I could change it use the `TestUtils.getDatabaseConnectionToExistingDb` to get around this, but I saw this method as a temporary stop gap until the query tests were changed.

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
rhyshort: No, it only shutsdown the queue, the database would remain open unless the call to database.close() is made on the queue's executor thread

mikerhodes: Is it not this method being called then: https://github.com/cloudant/sync-android/pull/84/files#diff-aff2a570e63fa1ae9414a7e1e6568532R97?

rhyshort: No, the QueryCode is using the ExecutorService as its queue, rather than `SQLDatabaseQueue`. I made only minimal changes to query to counter the removal of `ThreadLocal`  I can make those changes on this PR if you think it should be done now not later

mikerhodes: This is fine.

sync-core/src/test/java/com/cloudant/sync/datastore/AttachmentTest.java
mikerhodes: We checked that the Asserts fail correctly.

mikerhodes:
This looks good to me now. I suggest squashing it up. I'd also revisit your commit message as what it says isn't strictly true. We execute all database requests on a serial queue _not_ a single thread -- the queue is free to use a new thread if it wants, all it guarantees is that the queued callables happen in serial. You should update the commit message to discuss running on a serial queue rather than a single thread.

PR #85, opened 2015-01-21
42784 correct encoding of database path

tomblench:
Some refactoring to ensure the user passes in a correctly encoded URL for remote databases. This includes removing the 'default DB' anti-pattern in the Mazha library.

This way, we don't try and second-guess the path structure which breaks reverse proxies/forwarding etc.

Documentation needs to be updated explaining that users need to do their own URL escaping and documenting some of the withdrawn APIs.


sync-core/src/main/java/com/cloudant/sync/replication/PullReplication.java
```diff
@@ -15,7 +15,7 @@
 public class PullReplication extends Replication {
 
     /**
-     * URI for this replication's remote database.
+     * URI for this replication's remote database (including username and password if necessary).
```
mikerhodes: Could this say "Include username and password in the URL, or supply an Authorization header using optionalHeaders"?

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -50,61 +47,45 @@
...
-    protected JSONHelper getJson() {
-        return json;
-    }
-
-    public List<String> allDbs() {
-        URI allDBUri = uriHelper.allDbsUri();
-        return getList(allDBUri);
+    protected JSONHelper getJsonHelper() {
+        return jsonHelper;
     }
```
mikerhodes: You use `this.jsonHelper` in `revsDiff` rather than the accessor. As jsonHelper is final, you could perhaps just make the ivar protected rather than having the accessor.

mikerhodes:
I think once we test the slashed-path versions by creating a subclass (if I'm not suggesting something that won't work) we're good here.

PR #86, opened 2015-01-27
Add support for complex querying

alfinkel:
- Update QueryExecutor to handle Or query nodes.
- Update QuerySqlTranslator to handle OR, OR and AND sub clauses, NOT,
and EXISTS.
- Update QueryValidator to handle compound clauses.
- Update UnindexedMatcher to handle OR, OR and AND sub clauses, NOT,
and EXISTS.
- Add tests to AbstractQueryTestBase to test all complex query
functionality.
- Add UnindexedMatcher, QuerySqlTranslator, and QueryValidator tests.
- Split out the set up of data into separate abstract class for tests.
- Create an extended query base test abstract class used to execute
tests for queries that are not covered by indexes.
- Fix bug in IndexUpdater that came up during dotted notation testing.
- Change return type for QueryResult.size() from long to int so that it
matches the return type of QueryResult.documentIds().size().

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
@@ -207,15 +207,23 @@ class QueryExecutor {
                 } else {
                     accumulator = Sets.intersection(accumulator, childIds);
                 }
-
-                // TODO optimisation is to bail here if accumulator is empty
             }
 
             return accumulator;
         }
         if (node instanceof OrQueryNode) {
             Set<String> accumulator = null;
-            // TODO - OR execute query tree logic...
+
+            OrQueryNode orNode = (OrQueryNode) node;
+            for (QueryNode qNode: orNode.children) {
+                Set<String> childIds = executeQueryTree(qNode, db);
+                if (accumulator == null) {
+                    accumulator = new HashSet<String>(childIds);
+                } else {
+                    accumulator = Sets.union(accumulator, childIds);
+                }
+            }
+
             return accumulator;
         } else if (node instanceof SqlQueryNode) {
             SqlQueryNode sqlNode = (SqlQueryNode) node;
--- a/sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
@@ -96,7 +96,6 @@ class QuerySqlTranslator {
 
     private static final String AND = "$and";
     private static final String OR = "$or";
-    private static final String EQ = "$eq";
     private static final String NOT = "$not";
     private static final String EXISTS = "$exists";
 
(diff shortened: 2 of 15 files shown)


sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
```diff
@@ -207,15 +207,23 @@ private boolean validateSortDocument(List<Map<String, String>> sortDocument) {
                 } else {
                     accumulator = Sets.intersection(accumulator, childIds);
                 }
-
-                // TODO optimisation is to bail here if accumulator is empty
             }
 
             return accumulator;
         }
         if (node instanceof OrQueryNode) {
```
mikerhodes: Looks like this needs to be `else if`.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -217,7 +216,35 @@ private static QueryNode translateQuery(Map<String, Object> query,
             // We could optimise for OR parts where we have an appropriate compound index,
             // but we don't for now.
 
-            // TODO - implement OR logic...
+            for (Object basicClause : basicClauses) {
+                List<Object> wrappedClause = Arrays.asList(basicClause);
```
mikerhodes: You don't create this `wrappedClause` for the `AND` case -- is it needed here?

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
alfinkel: This was a straight port from iOS version. https://github.com/cloudant/CloudantQueryObjc/blob/master/Pod/Classes/CDTQQueryExecutor.m#L240.  I can change it to an `else if` since I don't think functionally it will change anything but I wanted the Android and iOS versions to be as in synch as possible.  Let me know if you still rather it be set to an `else if`.

sync-core/src/main/java/com/cloudant/sync/query/QueryValidator.java
```diff
@@ -164,9 +175,98 @@ private static boolean validateSelector(Map<String, Object> selector) {
...
+            if (clause.size() != 1) {
+                String msg;
+                msg = String.format("Operator argument clause should have one key value pair: %s",
+                                    clauses.toString());
+                logger.log(Level.SEVERE, msg);
+                break;
+            }
+
+            String key = (String) clause.keySet().toArray()[0];
+            if (Arrays.asList("$or", "$not", "$and").contains(key)) {
```
mikerhodes: Create this list statically.

mikerhodes: Actually don't bother for now!

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
```diff
@@ -304,23 +322,73 @@ protected static boolean compareNE(Object l, Object r) {
...
+            return lStr.compareTo(rStr) < 0;
+        } else if (r instanceof String) {
+            // At this point in the logic l can only be a number
+            return true;  // INT < STRING
+        } else {
+            // At this point in the logic both l and r can only be numbers
+            Number lNum = (Number) l;
+            Number rNum = (Number) r;
+
+            return lNum.longValue() < rNum.longValue();
```
mikerhodes: This fails for, e.g., `1.1 < 1.2`.

alfinkel: Oh man, I missed that completely.  I fixed it in compareEq but left it here.  I'll fix it and add more tests to the unindexed matcher tests to specifically test all compare methods.

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
```diff
@@ -304,23 +322,73 @@ protected static boolean compareNE(Object l, Object r) {
     //  3. TEXT
     //  4. BLOB
     protected static boolean compareLT(Object l, Object r) {
-        // TODO: 275 - 309 CDTQUnindexedMatcher.m
-        return false;
+        if (l == null) {
+            return false;  // null fails all lt/gt/lte/gte tests
+        } else if (!(l instanceof String || l instanceof Number && !(l instanceof Float))) {
```
mikerhodes: Seems like you want to pull this out as a separate method.

mikerhodes:
Okay, looks like the main thing is the compare bug, and more tests.

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
alfinkel: Created an FB issue to revisit this later in both Android and iOS.
https://cloudant.fogbugz.com/f/cases/43772/Cloudant-Query-Revisit-executeQueryTree-and-executeSelectorTree-if-else-structure

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
alfinkel: This is fixed and more tests have been added.

PR #87, opened 2015-01-28
Add copyright notices to files which had them missing

tomblench:


mikerhodes:
I suggest we merge this to get the headers in as they set up the Apache licence. We can decide whether this needs to be IBM later.

PR #88, opened 2015-01-29
Add sorting support to Query

alfinkel:
- Implement sorting logic in QueryExecutor for indexed queries
- Add sorting tests
- Set copyright date to 2015 for AbstractQueryExtendedTestBase
- Remove unused imports in QuerySqlTranslatorTest

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
@@ -230,7 +230,9 @@ class QueryExecutor {
                 if (accumulator == null) {
                     accumulator = new HashSet<String>(childIds);
                 } else {
-                    accumulator = Sets.intersection(accumulator, childIds);
+                    if (childIds != null) {
+                        accumulator = Sets.intersection(accumulator, childIds);
+                    }
                 }
             }
 
@@ -248,7 +250,9 @@ class QueryExecutor {
                 if (accumulator == null) {
                     accumulator = new HashSet<String>(childIds);
                 } else {
-                    accumulator = Sets.union(accumulator, childIds);
+                    if (childIds != null) {
+                        accumulator = Sets.union(accumulator, childIds);
+                    }
                 }
             }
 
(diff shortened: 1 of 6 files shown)


sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
```diff
@@ -205,7 +227,9 @@ private boolean validateSortDocument(List<Map<String, String>> sortDocument) {
                 if (accumulator == null) {
                     accumulator = new HashSet<String>(childIds);
                 } else {
-                    accumulator = Sets.intersection(accumulator, childIds);
+                    if (childIds != null) {
+                        accumulator = Sets.intersection(accumulator, childIds);
+                    }
```
mikerhodes: Feels more logical to me to have an `if (childIds == null) { continue; }` after the assignment to `childIds`, as this avoids creating the accumulator if there are never any children. Also, perhaps the `HashSet` constructor might not like nil?

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
```diff
@@ -266,27 +292,135 @@ private boolean validateSortDocument(List<Map<String, String>> sortDocument) {
                                  List<Map<String, String>> sortDocument,
                                  Map<String, Object> indexes,
                                  SQLDatabase db) {
-        // TODO - sort logic...
-        return null;
+        boolean smallResultSet = (docIdSet.size() < SMALL_RESULT_SET_SIZE_THRESHOLD);
+        SqlParts orderBy = sqlToSortIds(docIdSet, sortDocument, indexes);
+        List<String> sortedIds = new ArrayList<String>();
```
mikerhodes: Probably move the assignment inside the `if`, so we can return `null` without the `else`, or mirror the assignment to `nil` in the objc code rather than doing an early return.

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
alfinkel: Moved check for null to after childIds assignment.  https://github.com/cloudant/sync-android/commit/6179fea1b1139126f9e958e9eaf59fdea174709a#diff-80ec4d1d12847da53428d6c14a88dbb7R227

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
alfinkel: Changed sortedIds assignment inside `else` block to mirror objc implementation.  https://github.com/cloudant/sync-android/commit/6179fea1b1139126f9e958e9eaf59fdea174709a#diff-80ec4d1d12847da53428d6c14a88dbb7R331

mikerhodes:
+1 once Travis is happy.  It looks like the four commits *aside from* 44b9739 should be squashed together, then we can merge.

PR #90, opened 2015-02-03
Add field projection support to query

alfinkel:
- Update QueryResult to handle field projection.
- Create ProjectedDocumentRevision a sub-class of
BasicDocumentRevision used to ensure that the mutableCopy contract is
upheld when dealing with documents retrieved using query field
projection.
- Add method to DocumentRevisionBuilder to build a
ProjectedDocumentRevision.
- Add query field projection tests.

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/datastore/DocumentRevisionBuilder.java
+++ b/sync-core/src/main/java/com/cloudant/sync/datastore/DocumentRevisionBuilder.java
@@ -59,6 +59,7 @@ public class DocumentRevisionBuilder {
     private long docInternalId = -1;
     private long parent = -1;
     private List<? extends Attachment> attachments = null;
+    private Datastore datastore = null;
 
     /**
      * <p>Builds and returns the {@code BasicDocumentRevision} for this builder.</p>
@@ -200,6 +201,16 @@ public class DocumentRevisionBuilder {
     }
 
     /**
+     * <p>Sets the datastore for this builder.</p>
+     * @param datastore the datastore
+     * @return the builder object for chained calls
+     */
+    public DocumentRevisionBuilder setDatastore(Datastore datastore) {
+        this.datastore = datastore;
+        return this;
+    }
+
+    /**
      * Builds and returns the {@link com.cloudant.sync.datastore.MutableDocumentRevision} for this builder.
      * @return {@link com.cloudant.sync.datastore.MutableDocumentRevision} for this builder.
      */
@@ -208,6 +219,21 @@ public class DocumentRevisionBuilder {
         revision.body = this.body;
         revision.docId = this.docId;
         return revision;
+    }
+
+    /**
+     * Builds and returns the {@link com.cloudant.sync.datastore.ProjectedDocumentRevision} for this builder.
+     * @return {@link com.cloudant.sync.datastore.ProjectedDocumentRevision} for this builder.
+     */
+    public ProjectedDocumentRevision buildProjected() {
+        BasicDocumentRevision.BasicDocumentRevisionOptions options = new BasicDocumentRevision.BasicDocumentRevisionOptions();
+        options.sequence = sequence;
+        options.docInternalId = docInternalId;
+        options.deleted = deleted;
+        options.current = current;
+        options.parent = parent;
+        options.attachments = attachments;
+        return new ProjectedDocumentRevision(docId, revId, body, options, datastore);
     }
 
     /**
--- a/sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
@@ -230,9 +230,7 @@ class QueryExecutor {
                 if (accumulator == null) {
                     accumulator = new HashSet<String>(childIds);
                 } else {
-                    if (childIds != null) {
-                        accumulator = Sets.intersection(accumulator, childIds);
-                    }
+                    accumulator = Sets.intersection(accumulator, childIds);
                 }
             }
 
@@ -250,9 +248,7 @@ class QueryExecutor {
                 if (accumulator == null) {
                     accumulator = new HashSet<String>(childIds);
                 } else {
-                    if (childIds != null) {
-                        accumulator = Sets.union(accumulator, childIds);
-                    }
+                    accumulator = Sets.union(accumulator, childIds);
                 }
             }
 
--- a/sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
@@ -196,7 +196,7 @@ class QueryExecutor {
             return true;
         }
         for (String field: fields) {
-            if (field.contains("\\.")) {
+            if (field.contains(".")) {
                 String msg = String.format("Projection field cannot use dotted notation: %s",
                         field);
                 logger.log(Level.SEVERE, msg);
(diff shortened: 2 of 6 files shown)


sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
```diff
@@ -196,7 +196,7 @@ private boolean validateFields(List<String> fields) {
             return true;
         }
         for (String field: fields) {
-            if (field.contains("\\.")) {
+            if (field.contains(".")) {
```
mikerhodes: To me this is a separate commit, "Fix bug in validating dotted projection fields", as it fixes an existing bug.

sync-core/src/main/java/com/cloudant/sync/datastore/DocumentRevisionBuilder.java
```diff
@@ -211,6 +222,21 @@ public MutableDocumentRevision buildMutable() {
...
+    public ProjectedDocumentRevision buildProjected() {
+        BasicDocumentRevision.BasicDocumentRevisionOptions options = new BasicDocumentRevision.BasicDocumentRevisionOptions();
+        options.sequence = sequence;
+        options.docInternalId = docInternalId;
+        options.deleted = deleted;
+        options.current = current;
+        options.parent = parent;
+        options.attachments = attachments;
+        return new ProjectedDocumentRevision(docId, revId, body, options, datastore);
+    }
```
mikerhodes: `DocumentRevisionBuilder` is public. Building a projected document revision shouldn't be a public operation. I don't think updating the public API of the library for this is warranted.

I see that mutable document revisions are also created by this class. I don't think that's a good thing either, but we can remove that later.

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
```diff
@@ -250,9 +248,7 @@ private boolean validateFields(List<String> fields) {
                 if (accumulator == null) {
                     accumulator = new HashSet<String>(childIds);
                 } else {
-                    if (childIds != null) {
-                        accumulator = Sets.union(accumulator, childIds);
-                    }
+                    accumulator = Sets.union(accumulator, childIds);
```
mikerhodes: Why remove this check now?

sync-core/src/test/java/com/cloudant/sync/query/QueryFilterFieldsTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+    @Test
+    public void returnsFieldSpecifiedOnly() {
+        // query - { "name" : "mike" }
+        Map<String, Object> query = new HashMap<String, Object>();
+        query.put("name", "mike");
+        QueryResult queryResult = im.find(query, 0, Long.MAX_VALUE, Arrays.asList("name"), null);
+        for (DocumentRevision rev : queryResult) {
+            Map<String, Object> revBody = rev.getBody().asMap();
+            assertThat(revBody.keySet(), contains("name"));
```
mikerhodes: Does this check that there are no other keys in the dictionary?

sync-core/src/test/java/com/cloudant/sync/query/QueryFilterFieldsTest.java
```diff
@@ -0,0 +1,203 @@
...
+        }
+    }
+
+    @Test
+    public void returnsNullWhenUsingDottedNotation() {
+        // query - { "name" : "mike" }
+        Map<String, Object> query = new HashMap<String, Object>();
+        query.put("name", "mike");
+        assertThat(im.find(query, 0, Long.MAX_VALUE, Arrays.asList("name.blah"), null),
+                is(nullValue()));
```
mikerhodes: Wrong indent, `is` call is inside the `assertThat` call, indent to beyond the bracket?

sync-core/src/test/java/com/cloudant/sync/query/QueryFilterFieldsTest.java
```diff
@@ -0,0 +1,203 @@
...
+            } catch (ConflictException e) {
+                Assert.fail("Failed to delete document revision");
+                e.printStackTrace();
+            }
+
+            assertThat(((ProjectedDocumentRevision) rev).mutableCopy(), is(nullValue()));
+        }
+    }
+
+}
```
mikerhodes: Set up IDEA to add new lines to the end of files by default (http://stackoverflow.com/questions/16761227/how-to-make-intellij-idea-insert-a-new-line-at-every-end-of-file). But it's not worth fixing this one up right now.

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
alfinkel: This check really should have been removed when we moved the childIds == null check above the accumulator creation. https://github.com/cloudant/sync-android/commit/a4626ee5d3f29d9899d43e4a01467be78d92bd19#diff-80ec4d1d12847da53428d6c14a88dbb7L227

Basically this check for childIds != null is redundant and needed to be removed.

sync-core/src/test/java/com/cloudant/sync/query/QueryFilterFieldsTest.java
alfinkel: Yes the `contains` Hamcrest matcher method works just like `containsInAnyOrder` and actually means "contains and is limited to...in content".

sync-core/src/main/java/com/cloudant/sync/datastore/DocumentRevisionBuilder.java
mikerhodes: I see there isn't an easy way to do something that isn't public given our class structure. Let's leave this for now.

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
mikerhodes: Separate commit here too then.

sync-core/src/test/java/com/cloudant/sync/query/QueryFilterFieldsTest.java
mikerhodes: Ok.

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
alfinkel: Separated into its own commit.  https://github.com/cloudant/sync-android/commit/dca31acc83c8414a141238b73bc34182058d384d

sync-core/src/test/java/com/cloudant/sync/query/QueryFilterFieldsTest.java
alfinkel: Fixed in https://github.com/cloudant/sync-android/commit/fde66732f755b51b30bbea43a00d30672e1fd403

sync-core/src/test/java/com/cloudant/sync/query/QueryFilterFieldsTest.java
alfinkel: Ok

sync-core/src/main/java/com/cloudant/sync/query/QueryExecutor.java
alfinkel: Separated to its own commit https://github.com/cloudant/sync-android/commit/cfcced806804a1b16573bce95252d7595ba4039d

mikerhodes:
Let's merge. +1.

PR #91, opened 2015-02-04
Changes to work with Cloudant local

rhyshort:
- Make sequence numbers transparent, previously they were expected
  to be strings, this causes failures when replicating with
  Cloudant Local which uses arrays for sequence numbers.

- Also added a more configurable couch config to enable
  testing with any couchDB server.


sync-core/src/test/java/com/cloudant/common/CouchTestBase.java
```diff
@@ -25,7 +29,10 @@ public CouchConfig getCouchConfig(String db) {
                         " is required to run tests with Cloudant.");
             }
             return config;
-        } else {
+        } else if (SPECIFIED_COUCH) {
+            return SpecifiedCouch.defaultConfig(db);
```
tomblench: I don't really see the value of this SpecifiedCouch class. Why not just let the user pass in a url string and have done with it?

rhyshort: it follows the pattern we've used previously with cloudant, it has its trade offs, there are a lot more config options, a url would be easier, but may be confusing is you don't know how to pass a username and password through the url

sync-core/src/main/java/com/cloudant/mazha/CouchURIHelper.java
```diff
@@ -61,6 +63,17 @@ public URI changesUri(Map<String, Object> query) {
                 this.rootUriString,
                 "_changes"
         );
+
+        //lets find the since parameter
+        if(query.containsKey("since")){
```
tomblench: would it be better to do the JSON encoding at the call site (whichever method is preparing the _changes url) rather than special-casing it here?

rhyshort: Not really, encoding it here means only the helper needs to know about encoding, where as if we place it at the call site for the method we have the potential for the implementation detail about sequence numbers to leak out more

mikerhodes:
Split 3b9557f in two:

1. `Also added a more configurable couch config to enable  testing with any couchDB server.`
2. The fixes for Local.

"Also" should be a warning light in your head.

Also, we made the sequence number opaque to the code, not transparent. I'd just use "pass-through".

sync-core/src/main/java/com/cloudant/mazha/CouchURIHelper.java
```diff
@@ -61,6 +63,17 @@ public URI changesUri(Map<String, Object> query) {
...
+            Object since = query.get("since");
+            if(!(since instanceof String)){
```
mikerhodes: Did you confirm that the wrong thing happens if we remove this check and pass a string? That is, we end up with `since="123-blah"` rather than `since=123-blah`.

sync-core/src/main/java/com/cloudant/mazha/CouchURIHelper.java
mikerhodes: Tom might have a point, it's a slightly weird place to have this logic. But given the warped nature of some of this code it might be the most sensible right now. Who calls this method with a `since` param?

sync-core/src/test/java/com/cloudant/common/CouchTestBase.java
mikerhodes: Can we replace the `TEST_WITH_CLOUDANT` stuff with `SPECIFIED_COUCH`? It seems that they're both doing the same job, that is allowing the user to configure something other than a localhost-ish couchdb (just CloudantConfig is a bit more specific)?

sync-core/src/test/java/com/cloudant/sync/replication/DBWithSlashReplicationTest.java
```diff
@@ -66,7 +66,7 @@ public void replicationPullAndPushDbWithSlash() throws Exception{
         replicator.start();
         Assert.assertEquals(Replicator.State.STARTED, replicator.getState());
 
-        while(replicator.getState() != Replicator.State.COMPLETE) {
+        while(replicator.getState() != Replicator.State.COMPLETE || replicator.getState() == Replicator.State.ERROR) {
```
mikerhodes: This and the change to `PullReplicatorTest` look like they belong in a separate commit which explains what the change is for.

sync-core/src/main/java/com/cloudant/mazha/CouchURIHelper.java
rhyshort: Yeah, I ran a test against couch 1.6 and it was failing because it was encoding some quotes in the url.

sync-core/src/main/java/com/cloudant/sync/replication/CouchDB.java
```diff
@@ -43,8 +43,8 @@
     public String getCheckpoint(String checkpointId);
     public void putCheckpoint(String checkpointId, String sequence);
 
-    public ChangesResult changes(String lastSequence, int limit);
-    public ChangesResult changes(Replication.Filter filter,String lastSequence, int limit);
+    public ChangesResult changes(Object lastSequence, int limit);
+    public ChangesResult changes(Replication.Filter filter,Object lastSequence, int limit);
```
mikerhodes: Check your formatting, spaces after `,`.

mikerhodes:
Make the commit messages a little more result-orientated rather than what one did orientated.

Also, any docs need updating with the removal of CloudantConfig?

Code is looking good though, pleased we fixed this quite quickly.

rhyshort:
Yeah I updated the docs in 41e220b

rhyshort:
Verbal +1 from @mikerhodes merging

PR #92, opened 2015-02-05
Fix OR query bug when some indexes are missing

alfinkel:
Each clause in an OR query is treated separately.  If an index does not exist to handle a clause in an OR-ed query then the post hoc matcher must handle the query results.  This PR addresses the bug where the post hoc matcher was previously using a subset of the full list of documents to provide the (incorrect) query results.  With the work included here the full list of documents is now used.


alfinkel:
@mikerhodes - I also added another test that I meant to include with the field projection work.  It is separated out into its own commit https://github.com/cloudant/sync-android/commit/d5ef75c8b60f222a15863a8b92595efe733646a6.

mikerhodes:
Looks good, fixup 39fad58 into c771fdb then merge. +1.

PR #93, opened 2015-02-09
43599 improve exceptions

rhyshort:
Improve the exceptions that are thrown from the datastore when a error occurs. Currently most of the exceptions that are thrown are unchecked exceptions, and most are inappropriate for the situation they are being thrown for. Exceptions such as IOException thrown when perhaps a better exception should be used. The aim of this PR is to add the following exception hierarchy and have the API fulfil its contract as laid out by the JavaDoc. 
```
- DatastoreException
  |
  |
  |
  |
  |- DocumentException
  |		|
  |		|- InvalidDocumentException
  |		|- DocumentNotFoundException
  |		|- ConflictException 
  |
  |
  |- AttachmentException
  |		|
  |		|- AttachmentNotSavedException
  |
  |- DatastoreNotCreatedException
  |
```

The main purpose for this PR is get comments on what the Exceptions should be and thrown where, before correcting all JavaDoc and documenting all API changes for the user.

On that note things left to do:

- JavaDoc updates
- Documentation on API changes


sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -107,21 +107,35 @@ public void addAttachment(SQLDatabase db,PreparedAttachment a, BasicDocumentRevi
...
         // move file to blob store, with file name based on sha1
         File newFile = fileFromKey(sha1);
-        try{
-            FileUtils.moveFile(a.tempFile, newFile);
-        } catch (FileExistsException fee) {
-            // File with same SHA1 hash in the store, we assume it's the same content so can discard
-            // the duplicate data we have just downloaded
-            a.tempFile.delete();
+
+        if(newFile.exists()){
```
tomblench: I'm not sure if it's useful to add this newFile.exists() check. You're still catching the FileExistsException anyway, so it's redundant. Obviously you can't remove that catch because there would be a race condition on the file system. So just try to do the move and deal with it in the catch block.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -283,7 +292,7 @@ public DocumentRevisionTree call(SQLDatabase db) throws Exception {
         return null;
     }
 
-    private DocumentRevisionTree getAllRevisionsOfDocumentInQueue(SQLDatabase db, String docId){
+    private DocumentRevisionTree getAllRevisionsOfDocumentInQueue(SQLDatabase db, String docId) throws DocumentNotFoundException, AttachmentException {
```
tomblench: Some long lines with these `throws` clauses, here and below

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -630,23 +647,21 @@ private BasicDocumentRevision createDocument(SQLDatabase db,String docId, final
...
             options.data = body.asBytes();
             options.available = true;
             options.copyAttachments = false;
-            long newSequence = insertRevision(db,options);
-            if (newSequence < 0) {
-                throw new IllegalStateException("Error inserting new revision: "+options);
-            }
-
-            BasicDocumentRevision doc = getDocumentInQueue(db, docId, options.revId);
-            documentCreated = new DocumentCreated(doc);
```
tomblench: I guess you dropped `documentCreated = new DocumentCreated(doc);` because it was an unused variable. This is supposed to be posted on the event queue and must have been deleted in a previous commit. A bit bad that this wasn't caught by a unit test.

rhyshort: The notifications are fired after the queue completes, so there was some logic move around in previous work to do with DB queues. I've taken the approach to do clean up when I can. There may be some cases where the notification is fired anymore, but its a small price to pay for avoid the possibility of a user deadlocking our internal queue.

tomblench: OK I now see that these events get fired elsewhere.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -611,7 +620,15 @@ private BasicDocumentRevision createDocument(SQLDatabase db,String docId, final
             // * normal insert logic for a new document
 
             InsertRevisionOptions options = new InsertRevisionOptions();
-            BasicDocumentRevision potentialParent = this.getDocumentInQueue(db, docId, null);
+            BasicDocumentRevision potentialParent = null;
```
tomblench: Remove the unused `documentCreated` variable above, and indent all of this left one level to be inline with the above test.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -791,53 +795,66 @@ private BasicDocumentRevision deleteDocumentInQueue(SQLDatabase db, final String
...
-                                docId);
-                        if(revisionTree == null) {
-                            throw new IllegalArgumentException("Document does not exist for id: " + docId);
-                        } else if (!revisionTree.leafRevisionIds().contains(prevRevId)) {
-                            throw new ConflictException("Revision to be deleted is not a leaf node:"
-                                    + prevRevId);
-                        }
+        DocumentRevisionTree revisionTree;
+        try {
+            revisionTree = getAllRevisionsOfDocumentInQueue(db, docId);
```
tomblench: Seems like you have removed some of the error checking in this commit, "Document does not exist for id" and "Revision to be deleted is not a leaf node:"

Also it might be worth explaining why an AttachmentException is being re-thrown as a DocumentNotFoundException. I know it's legitimate because it's a reason why we can't fetch the document but without a comment it's not obvious.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -791,53 +795,66 @@ private BasicDocumentRevision deleteDocumentInQueue(SQLDatabase db, final String
...
+            try {
+                insertRevision(db, options);
+            } catch (AttachmentException e) {
+                //this shouldn't happen since attachments are not copied
+                throw new RuntimeException("Attachments should not be copied for a delete op",e);
+            }
 
-                    return deletedRevision;
+            try {
+                return getDocumentInQueue(db, preRevision.getId(),
```
tomblench: Make it clearer that we are fetching back the deleted document to return to the caller. A comment and maybe re-instating the `deletedRevision` variable would help.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -791,53 +795,66 @@ private BasicDocumentRevision deleteDocumentInQueue(SQLDatabase db, final String
...
+                        newRevisionId);
+            } catch (AttachmentException e){
+                throw new DocumentNotFoundException(e);
```
tomblench: As above where this pattern is seen, explain why it's re-thrown as a different exception.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -921,25 +930,20 @@ private long insertRevision(SQLDatabase db,InsertRevisionOptions options) {
             logger.fine("New revision inserted: " + options.docNumericId + ", " + options.revId);
             newSequence = db.insert("revs", args);
             if (newSequence < 0) {
-                throw new IllegalStateException("Unknown error inserting new updated doc, please checking log");
+                throw new IllegalStateException("Unknown error inserting new updated doc, please checking log"); //TODO is this recoverable?
```
tomblench: newSequence < 0 means the row couldn't be inserted, which is probably something quite serious and probably not recoverable. Also take the opportunity to clean up the grammar on this error message.

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
rhyshort: Its kinda a hacky fix, for a problem I encountered, the temp file had already been copied, so we already have a file with the same sha1 hash, but the temp file doesn't exist. This was the only way I could think of to fix it. If that section of code is reverted and `AttachmentsConflictsTest` is run the problem is reproduced.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
rhyshort: The leaf node check is after the catch block, it still throws an ConflictException if it's not a leaf node.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -977,16 +994,17 @@ private BasicDocumentRevision doGetLocalDocument(SQLDatabase db, String docId, S
 
                 return builder.buildBasicDBObjectLocalDocument();
             } else {
-                return null;
+                throw new DocumentNotFoundException("No local document found with id: " + docId + ", revId: " + revId);
             }
         } catch (SQLException e) {
-            throw new SQLRuntimeException("Error getting local document with id: " + docId, e);
+            throw new DocumentNotFoundException("Error getting local document with id: " + docId, e);
```
tomblench: Is this really a DocumentNotFoundException? Obviously something went wrong with the SQL so we can't return the document, but isn't DatastoreException more appropriate?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -998,7 +1016,7 @@ private BasicDocumentRevision doGetLocalDocument(SQLDatabase db, String docId, S
                 result.add(row);
             }
         } catch (SQLException e) {
-            e.printStackTrace();  //To change bodyOne of catch statement use File | Settings | File Templates.
+           throw new DocumentNotFoundException(e);
```
tomblench: As above, is DocumentNotFoundException appropriate?

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -159,7 +170,7 @@ protected PreparedAndSavedAttachments prepareAttachments(Collection<? extends At
         return preparedAndSavedAttachments;
     }
 
-    protected void setAttachments(SQLDatabase db,BasicDocumentRevision rev, PreparedAndSavedAttachments preparedAndSavedAttachments) throws IOException {
+    protected void setAttachments(SQLDatabase db,BasicDocumentRevision rev, PreparedAndSavedAttachments preparedAndSavedAttachments) throws AttachmentNotSavedException, AttachmentException {
```
tomblench: AttachmentNotSavedException is reduandant here as it's a subclass of AttachmentException

tomblench: (there are a few more of these elsewhere)

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -857,15 +887,7 @@ public Object call(SQLDatabase db) throws Exception {
...
-            if(e.getCause() != null){
-                if(e.getCause() instanceof DocumentNotFoundException){
-                    throw (DocumentNotFoundException)e.getCause();
-                } else {
-                    throw new RuntimeException(e);
-                }
-            } else {
-                throw new RuntimeException(e);
-            }
+           throw new DocumentNotFoundException(String.format("Did not find document with id: %s",docId),e);
```
tomblench: s/Did not/Could not/

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1250,7 +1301,12 @@ private long insertDocumentHistoryIntoExistingTree(SQLDatabase db, BasicDocument
         options.data = newRevision.asBytes();
         options.available = true;
         options.copyAttachments = false;
-        long sequence = insertRevision(db,options);
+        long sequence;
+        try {
+            sequence = insertRevision(db, options);
+        } catch (AttachmentException e){
+            throw new RuntimeException("Attachments should not be copied when inserting document history",e);
```
tomblench: "should not"? do you mean you don't expect this exception to be thrown, since we're not doing the copy attachments part?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
tomblench: (or rather DocumentException)

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
rhyshort: I'm not sure of the best pattern to use, I think it'll come down to the API design, where I think we should document all the exceptions that can be thrown.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
rhyshort: Yeah, perhaps the wording could be improved, but if we hit that catch clause we should throw a RuntimeException rather than swallowing it

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -595,58 +608,65 @@ public BasicDocumentRevision call(SQLDatabase db) throws Exception {
...
+            // resurrecting the document
+        }
 
-            BasicDocumentRevision doc = getDocumentInQueue(db, docId, options.revId);
-            documentCreated = new DocumentCreated(doc);
+        if (potentialParent != null) {
+            if (!potentialParent.isDeleted()) {
+                // current winner not deleted, can't insert
+                throw new IllegalArgumentException("Can not insert new doc, likely the docId exists already: "
+                        + docId);
```
mikerhodes: Does this get thrown as a ConflictException somewhere else; it sure sounds like one.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -681,7 +702,7 @@ public BasicDocumentRevision call(SQLDatabase db) throws Exception {
 
                         long lineId = db.insert("localdocs", values);
                         if (lineId < 0) {
-                            throw new IllegalArgumentException("Can not insert new local doc, likely the docId exists already: "
+                            throw new DocumentException("Can not insert new local doc, likely the docId exists already: "
                                     + docId);
```
mikerhodes: Is this error even right? I don't think local docs are versions, at least not on Couch.

rhyshort: honestly, I don't know, from the code I thought local docs had revisions

mikerhodes: Plausibly this doesn't act the same as Couch. Something we might want to fix later.

sync-core/src/main/java/com/cloudant/sync/datastore/DatastoreExtended.java
```diff
@@ -263,6 +263,7 @@ public void forceInsert(BasicDocumentRevision rev,
      *
      * @return List of <code>Attachment</code>
      */
-    public List<? extends Attachment> attachmentsForRevision(BasicDocumentRevision rev);
+    public List<? extends Attachment> attachmentsForRevision(BasicDocumentRevision rev) throws
+            AttachmentException;
```
mikerhodes: `@throws` for the new exceptions?

sync-core/src/main/java/com/cloudant/sync/datastore/Datastore.java
```diff
@@ -328,7 +329,8 @@ public void resolveConflictsForDocument(String docId, ConflictResolver resolver)
      * @see Datastore#getEventBus()
      * @see com.cloudant.sync.datastore.Datastore#deleteDocumentFromRevision(BasicDocumentRevision)
      */
-    public List<BasicDocumentRevision> deleteDocument(String id) throws ConflictException;
+    public List<BasicDocumentRevision> deleteDocument(String id) throws ConflictException,
+            DocumentException;
```
mikerhodes: `@throws` docs.

sync-core/src/main/java/com/cloudant/sync/datastore/Datastore.java
```diff
@@ -267,7 +268,7 @@ public void resolveConflictsForDocument(String docId, ConflictResolver resolver)
      * @throws IOException if there is a problem saving any new attachments
      * @see Datastore#getEventBus()
      */
-    public BasicDocumentRevision createDocumentFromRevision(MutableDocumentRevision rev) throws IOException;
+    public BasicDocumentRevision createDocumentFromRevision(MutableDocumentRevision rev) throws IOException, DocumentException;
```
mikerhodes: Seems odd that this doesn't throw an `AttachmentException` if update can?

mikerhodes: Or to state another way, should updateDocumentFromRevision just throw DocumentException?

sync-core/src/main/java/com/cloudant/sync/datastore/DocumentNotFoundException.java
```diff
@@ -19,13 +19,21 @@
 /**
  * Thrown when a document cannot be found.
  */
-public class DocumentNotFoundException extends RuntimeException {
+public class DocumentNotFoundException extends DocumentException {
+
+    public DocumentNotFoundException() {
+
+    }
```
mikerhodes: Seems like we could have a `docId, revId` and `docId` constructors which automatically create appropriate messages and call super appropriately rather than creating the message every time.

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -107,22 +108,32 @@ public void addAttachment(SQLDatabase db,PreparedAttachment a, BasicDocumentRevi
...
         File newFile = fileFromKey(sha1);
-        try{
+        try {
             FileUtils.moveFile(a.tempFile, newFile);
         } catch (FileExistsException fee) {
             // File with same SHA1 hash in the store, we assume it's the same content so can discard
             // the duplicate data we have just downloaded
             a.tempFile.delete();
+        } catch (FileNotFoundException e) {
+            //if the new exists we already have a copy of the attachment
```
mikerhodes: `if the new` _what_ `exists`?

mikerhodes: Seems like you might want to explain this a bit better. This is something like "If we had an error moving the file, but we already have a copy of the attachment in the datastore, we can ignore the problem as a copy of the data is already in the datastore. Otherwise, we must throw an exception."

This also seems to apply in the `IOException` case, so I'm not sure why that one just throws whatever happens. Is it because if `FileNotFoundException` is thrown we're guaranteed `newFile` hasn't been partially written whereas for `IOException` that's not the case?

Do we need to clean up after ourselves (e.g., for partly written new files or the temp file) for any of these exceptions?

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -272,11 +283,17 @@ protected void copyAttachment(SQLDatabase db, long parentSequence, long newSeque
...
-        DatabaseUtils.closeCursorQuietly(c);
+    protected void copyAttachments(SQLDatabase db, long parentSequence, long newSequence) throws AttachmentException {
+        Cursor c = null;
+        try {
+             c = db.rawQuery(SQL_ATTACHMENTS_SELECT_ALL,
+                    new String[]{String.valueOf(parentSequence)});
+            copyCursorValuesToNewSequence(db, c, newSequence);
+        } catch (SQLException e){
+            throw new AttachmentException("Failed to copy attachments",e);
+        }finally {
```
mikerhodes: spacing; probably more of these.

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentNotSavedException.java
```diff
@@ -0,0 +1,35 @@
...
+        super(message);
+    }
+
+    public AttachmentNotSavedException(Exception causedBy){
+        super(causedBy);
+    }
+
+    public AttachmentNotSavedException(String message, Exception casuedBy){
+        super(message,casuedBy);
+    }
```
mikerhodes: Spacing, again could there be helper constructors.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -595,58 +608,65 @@ public BasicDocumentRevision call(SQLDatabase db) throws Exception {
...
+            // resurrecting the document
+        }
 
-            BasicDocumentRevision doc = getDocumentInQueue(db, docId, options.revId);
-            documentCreated = new DocumentCreated(doc);
+        if (potentialParent != null) {
+            if (!potentialParent.isDeleted()) {
+                // current winner not deleted, can't insert
+                throw new ConflictException(String.format("Can not create new doc, document with id %s already exists "
+                        , docId));
```
mikerhodes: "Cannot", "new" is redundant, "doc" -> "document" (be consistent)

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -595,58 +608,65 @@ public BasicDocumentRevision call(SQLDatabase db) throws Exception {
...
+            }
+            // if we got here, parent rev was deleted
+            this.setCurrent(db,potentialParent, false);
```
mikerhodes: spacing. comma looks like a dot.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -595,58 +608,65 @@ public BasicDocumentRevision call(SQLDatabase db) throws Exception {
...
+            options.revId = CouchUtils.generateNextRevisionId(potentialParent.getRevision());
+            options.docNumericId = potentialParent.getInternalNumericId();
+            options.parentSequence = potentialParent.getSequence();
+        } else {
+            // otherwise we are doing a 'normal' create document
```
mikerhodes: Scare quotes around "normal", why?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -595,58 +608,65 @@ public BasicDocumentRevision call(SQLDatabase db) throws Exception {
...
+        options.copyAttachments = false;
+        insertRevision(db, options);
 
+        try {
+            BasicDocumentRevision doc =  getDocumentInQueue(db, docId, options.revId);
             logger.finer("New document created: " + doc.toString());
-
             return doc;
+        } catch (DocumentNotFoundException e){
+            throw new RuntimeException("Programming error we just inserted it so we should have it",e);
```
mikerhodes: Doesn't sound like a programming error to me? And this message is pretty useless for anyone using the library.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -718,28 +739,25 @@ private BasicDocumentRevision updateDocument(SQLDatabase db,String docId, String
...
-            BasicDocumentRevision newRevision = this.getDocumentInQueue(db, preRevision.getId(),
-                    newRevisionId);
+            this.setCurrent(db, preRevision, false);
+            String newRevisionId = this.insertNewWinnerRevision(db, body, preRevision, copyAttachments);
+            return this.getDocumentInQueue(db, preRevision.getId(), newRevisionId);
 
-            documentUpdated = new DocumentUpdated(preRevision, newRevision);
-            return newRevision;
+        } catch (DocumentNotFoundException e){
+            throw new IllegalArgumentException("The document trying to update does not exist.");
```
mikerhodes: Bad message, non-sensical as the document has no agency of its own.

mikerhodes: Also, is IllegalArgumentException correct? There's nothing wrong with the argument of itself, DocNotFound seems more logical here.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -764,25 +782,21 @@ public BasicDocumentRevision call(SQLDatabase db) throws Exception {
                         if (rowsUpdated == 1) {
                             return doGetLocalDocument(db,docId, newRevId);
                         } else {
-                            throw new IllegalStateException("Error updating local docs: " + preRevision);
+                            throw new DocumentException("Error updating local docs: " + preRevision);
```
mikerhodes: "doc" singular.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -791,53 +805,69 @@ private BasicDocumentRevision deleteDocumentInQueue(SQLDatabase db, final String
 
         CouchUtils.validateRevisionId(prevRevId);
 
-                    BasicDocumentRevision deletedRevision = null;
+        BasicDocumentRevision deletedRevision = null;
+            BasicDocumentRevision preRevision;
```
mikerhodes: spacing

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -791,53 +805,69 @@ private BasicDocumentRevision deleteDocumentInQueue(SQLDatabase db, final String
...
+        try {
+            preRevision = getDocumentInQueue(db, docId, prevRevId);
```
mikerhodes: Let's call it `prevRevision` to be consistent with `prevRevId`.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -791,53 +805,69 @@ private BasicDocumentRevision deleteDocumentInQueue(SQLDatabase db, final String
...
+        }  catch (AttachmentException e){
+            throw new DocumentNotFoundException(e);
+        }
```
mikerhodes: I've seen this flipping of exception type a few times now. 

Are there any cases where, when `getDocumentInQueue` throws an `AttachmentException` that it's not converted into `DocumentNotFoundException` exception by the caller? If not, do the conversion inside `getDocumentInQueue`.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -791,53 +805,69 @@ private BasicDocumentRevision deleteDocumentInQueue(SQLDatabase db, final String
...
-                                    + prevRevId);
-                        }
+        DocumentRevisionTree revisionTree;
+        try {
+            revisionTree = getAllRevisionsOfDocumentInQueue(db, docId);
+        } catch(AttachmentException e){
+            // we can't get load the document, due to an attachment error,
+            // throw an exception saying we couldn't find the document.
+            throw new DocumentNotFoundException(e);
+        }
```
mikerhodes: Another AttachmentException -> DocumentNotFoundException.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -791,53 +805,69 @@ private BasicDocumentRevision deleteDocumentInQueue(SQLDatabase db, final String
...
-                            options.available = false;
-                            options.copyAttachments = false;
-                            insertRevision(db,options);
-                            deletedRevision = getDocumentInQueue(db, preRevision.getId(),
-                                    newRevisionId);
-                            documentDeleted = new DocumentDeleted(preRevision, deletedRevision);
-                        }
+         if (!revisionTree.leafRevisionIds().contains(prevRevId)) {
+            throw new ConflictException("Revision to be deleted is not a leaf node:"
+                    + prevRevId);
```
mikerhodes: Say something that's more meaningful to a developer, like "Document has newer revisions than the revision passed to delete".

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -791,53 +805,69 @@ private BasicDocumentRevision deleteDocumentInQueue(SQLDatabase db, final String
...
+            options.deleted = true;
+            options.current = preRevision.isCurrent();
+            options.data = JSONUtils.EMPTY_JSON;
+            options.available = false;
+            options.copyAttachments = false;
+            try {
+                insertRevision(db, options);
+            } catch (AttachmentException e) {
+                //this shouldn't happen since attachments are not copied
+                throw new RuntimeException("Attachments should not be copied for a delete op",e);
```
mikerhodes: I don't even understand what this means.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -791,53 +805,69 @@ private BasicDocumentRevision deleteDocumentInQueue(SQLDatabase db, final String
...
+            }
+
+            try {
+                //get the deleted document revision to return to the user
+                return getDocumentInQueue(db, preRevision.getId(), newRevisionId);
+            } catch (AttachmentException e){
+                //throw document not found since we failed to load the document
+                throw new DocumentNotFoundException(e);
+            }
```
mikerhodes: Another AttachmentException -> DocumentNotFoundException.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -791,53 +805,69 @@ private BasicDocumentRevision deleteDocumentInQueue(SQLDatabase db, final String
...
+        } else {
+            throw new DocumentNotFoundException("Previous Revision is already deleted");
+        }
```
mikerhodes: Perhaps just have:

```java
if (!preRevision.isDeleted()) {
    throw new DocumentNotFoundException("Previous Revision is already deleted");
}
```

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -857,15 +887,7 @@ public Object call(SQLDatabase db) throws Exception {
...
-            if(e.getCause() != null){
-                if(e.getCause() instanceof DocumentNotFoundException){
-                    throw (DocumentNotFoundException)e.getCause();
-                } else {
-                    throw new RuntimeException(e);
-                }
-            } else {
-                throw new RuntimeException(e);
-            }
+           throw new DocumentNotFoundException(String.format("Could not find document with id: %s",docId),e);
```
mikerhodes: If this is the only exception it could be, why not just rethrow the one from inside the callable?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -921,25 +943,20 @@ private long insertRevision(SQLDatabase db,InsertRevisionOptions options) {
             logger.fine("New revision inserted: " + options.docNumericId + ", " + options.revId);
             newSequence = db.insert("revs", args);
             if (newSequence < 0) {
-                throw new IllegalStateException("Unknown error inserting new updated doc, please checking log");
+                throw new IllegalStateException("Unknown error inserting new updated doc, please check log");
```
mikerhodes: *what* log?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -954,7 +971,8 @@ private long insertStubRevision(SQLDatabase db, long docNumericId, String revId,
     }
 
     // Keep in mind we do not keep local document revision history
-    private BasicDocumentRevision doGetLocalDocument(SQLDatabase db, String docId, String revId) {
+    private BasicDocumentRevision doGetLocalDocument(SQLDatabase db, String docId, String revId)
+            throws DatastoreException {
```
mikerhodes: What about the document exceptions?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1067,90 +1083,115 @@ public Object call(SQLDatabase db) throws Exception{
...
-                                        throw e;
-                                    }
+
+                    // sequence here is -1, but we need it to insert the attachment - also might
+                    // be wanted by subscribers
+                    BasicDocumentRevision revisionFromDB = null;
+                    try {
+                        revisionFromDB = getDocumentInQueue(db,rev.getId(),null);
+                    } catch (DocumentNotFoundException e){
+                        //this is expected
```
mikerhodes: Explain why.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1067,90 +1083,115 @@ public Object call(SQLDatabase db) throws Exception{
...
-                                        if(doc != null) {
-                                            for (PreparedAttachment att : preparedAttachments.get(key)) {
+                                        if (doc != null) {
+                                            for (PreparedAttachment att : preparedAttachments.get
+                                                    (key)) {
                                                 attachmentManager.addAttachment(db, att, doc);
                                             }
                                         }
+                                    } catch (DocumentNotFoundException e){
+                                        continue;
```
mikerhodes: Not even worth printing a log message? Should we ever get a doc not found given we're inside a transaction?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1067,90 +1083,115 @@ public Object call(SQLDatabase db) throws Exception{
...
-                                         "attachment to " +
-                                        "the datastore, terminating replication", e);
-                                throw e;
                             }
-
+                        } catch (Exception e) {
+                            logger.log(Level.SEVERE, "There was a problem adding an " +
+                                    "attachment to " +
+                                    "the datastore, terminating replication", e);
+                            throw e;
```
mikerhodes: Will this really terminate the replication? Or just abort this particular transaction?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1067,90 +1083,115 @@ public Object call(SQLDatabase db) throws Exception{
...
-                                eventBus.post(documentUpdated);
-                            }
+
+                    }
+                    if (ok) {
+                        logger.log(Level.FINER, "Inserted revision: %s", rev);
+                        if (documentCreated != null) {
+                            eventBus.post(documentCreated);
+                        } else if (documentUpdated != null) { //TODO this event *should* not be
+                        // fired on the db queue.
```
mikerhodes: Is there a bug somewhere to remind us to fix this? Why shouldn't it be called on the queue? Why's it fine for created notifications?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1067,90 +1083,115 @@ public Object call(SQLDatabase db) throws Exception{
...
                         }
+                    }
 
 
                     return null;
                 }
             }).get();
         } catch (InterruptedException e) {
-            e.printStackTrace();
+            throw new RuntimeException(e);
```
mikerhodes: It seems pretty radical to throw a runtime exception for an interrupted exception -- do we do this generally? We probably need to consider what to do in this case if so. In general, we can probably retry?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1187,7 +1231,13 @@ private long doForceInsertExistingDocumentWithHistory(SQLDatabase db,BasicDocume
         Preconditions.checkArgument(revisions.size() > 0, "Revision history should have at least one revision." );
 
         // First look up all locally-known revisions of this document:
-        DocumentRevisionTree localRevs = getAllRevisionsOfDocumentInQueue(db, newRevision.getId());
+
+        DocumentRevisionTree localRevs;
+        try {
+            localRevs = getAllRevisionsOfDocumentInQueue(db, newRevision.getId());
+        } catch (DocumentNotFoundException e){
+            throw new RuntimeException("This should not be thrown, library programming error", e);
```
mikerhodes: Explain in a comment why it shouldn't be thrown, and possibly make a better log message so people can file a bug report if they encounter it.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
mikerhodes: Improve the message. I can't fathom what you're trying to say, let alone the poor programmer who receives this message.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1677,7 +1744,16 @@ private String insertNewWinnerRevision(SQLDatabase db,DocumentBody newWinner, Ba
...
         options.available = true;
         options.copyAttachments = copyAttachments;
-        this.insertRevision(db,options);
+        try {
+            this.insertRevision(db,options);
+        } catch (AttachmentException e){
+            if(copyAttachments){
+                throw e;
+            }else {
+                throw new RuntimeException("Attachments should not have been copied",e);
```
mikerhodes: This mystery error again.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1703,7 +1780,7 @@ private static BasicDocumentRevision getFullRevisionFromCurrentCursor(Cursor cur
             parent = cursor.getLong(7);
         } else if(cursor.columnType(7) == Cursor.FIELD_TYPE_NULL) {
         } else {
-            throw new IllegalArgumentException("Unexpected type: " + cursor.columnType(7));
+            throw new IllegalArgumentException("Unexpected type: " + cursor.columnType(7)); //TODO Should be straight Runtime Exception? Definitely not an argument
```
mikerhodes: Pretty sure Runtime is fine here. It's something that shouldn't happen I guess.

Any idea why the check for the column check? Does the schema change in a way that this might be true?

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1968,7 +2034,7 @@ public BasicDocumentRevision call(SQLDatabase db) throws Exception {
         } catch (InterruptedException e) {
             logger.log(Level.SEVERE,"Failed to delete document",e);
         } catch (ExecutionException e) {
-            logger.log(Level.SEVERE,"Failed to delete document",e);
+            throw new DocumentException("Failed to delete document",e);
```
mikerhodes: Worth appending `e`'s message to these generic messages that appear around the place? I guess the stack trace might contain `e`'s message?

sync-core/src/main/java/com/cloudant/sync/datastore/DocumentException.java
```diff
@@ -0,0 +1,41 @@
...
+ * and limitations under the License.
+ */
+
+package com.cloudant.sync.datastore;
+
+
+/**
+ * Created by Rhys Short on 05/02/15.
+ */
+public class DocumentException extends DatastoreException {
```
mikerhodes: What's the reasoning for DocumentException to extend DatastoreException? It feels like this might not be necessary.

sync-core/src/main/java/com/cloudant/sync/datastore/ProjectedDocumentRevision.java
```diff
@@ -34,16 +39,21 @@
...
     public MutableDocumentRevision mutableCopy() {
-        BasicDocumentRevision rev = datastore.getDocument(this.getId());
-        if (rev == null) {
-            return null;
-        }
+        try {
+            BasicDocumentRevision rev = datastore.getDocument(this.getId());
+            if (rev == null) {
+                return null;
+            }
```
mikerhodes: Can we remove this now?

sync-core/src/main/java/com/cloudant/sync/datastore/ProjectedDocumentRevision.java
```diff
@@ -34,16 +39,21 @@
...
+            // Don't want to return an updated version, breaks contract of mutableCopy
+            if (!rev.getRevision().equals(this.getRevision())) {
+                return null;
+            }
 
-        // Don't want to return an updated version, breaks contract of mutableCopy
-        if (!rev.getRevision().equals(this.getRevision())) {
+            return rev.mutableCopy();
+        } catch (DocumentNotFoundException e){
+            logger.log(Level.SEVERE,String.format("Failed to load document %s rom datastore",this.toString()),e);
```
mikerhodes: "from"

sync-core/src/main/java/com/cloudant/sync/datastore/SavedAttachment.java
```diff
@@ -58,10 +60,10 @@ public boolean isLarge() {
     public boolean shouldInline(PushAttachmentsInline inlinePreference) {
         // push_attachments_inline: false = always push multipart; small = push small attachments inline; true = always push inline
         if (inlinePreference == PushAttachmentsInline.False || inlinePreference == PushAttachmentsInline.Small && this.isLarge()) {
-            System.out.println("inline false");
+            logger.info("inline false");
```
mikerhodes: Make "finer" or perhaps  make the log message rather more useful.

sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
```diff
@@ -298,22 +299,24 @@ private int processOneChangesBatch(ChangesResultWrapper changeFeeds)
...
+                                        try {
+                                            BasicDocumentRevision dr = this.targetDb.getDbCore().getDocument(documentRevs.getId(), revId);
+                                                Attachment a = this.targetDb.getDbCore()
+                                                        .getAttachment(dr, attachmentName);
+                                                if (a != null) {
+                                                    // skip attachment, already got it
+                                                    continue;
+                                                }
+                                        } catch (DocumentNotFoundException e){
+                                            //do nothing
```
mikerhodes: Why?

sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
```diff
@@ -352,7 +359,12 @@ private int processOneChangesBatch(ChangesResultWrapper changeFeeds)
     public String getReplicationId() {
         HashMap<String, String> dict = new HashMap<String, String>();
         dict.put("source", this.sourceDb.getIdentifier());
-        dict.put("target", this.targetDb.getIdentifier());
+        try {
+            dict.put("target", this.targetDb.getIdentifier());
+        } catch (DatastoreException e){
+            logger.log(Level.WARNING,"Failed to get public id using default",e);
+            dict.put("target","touchdb_defaultUUID");
```
mikerhodes: This method should fail if you can't get the name of the target. Otherwise you're just going to end up with incorrect checkpoint docs.

sync-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java
```diff
@@ -315,7 +318,12 @@ private ItemsToPush missingRevisionsToJsonDocs(
 
     public String getReplicationId() {
         HashMap<String, String> dict = new HashMap<String, String>();
-        dict.put("source", this.sourceDb.getIdentifier());
+        try {
+            dict.put("source", this.sourceDb.getIdentifier());
+        } catch (DatastoreException e){
+            logger.log(Level.WARNING,"Failed to get ID using default",e);
+            dict.put("source","touchdb_defaultUUID");
+        }
```
mikerhodes: Again, silently swallowing this seems like a Very Bad Idea.

sync-core/src/main/java/com/cloudant/sync/replication/DatastoreWrapper.java
```diff
@@ -85,8 +87,8 @@ private String getCheckpointDocumentId(String replicatorIdentifier) {
         return "_local/" + replicatorIdentifier;
     }
 
-    public void bulkInsert(DocumentRevsList documentRevsList, Map<String[],List<PreparedAttachment>> preparedAttachments, boolean pullAttachmentsInline) {
-        for(DocumentRevs documentRevs: documentRevsList) {
+    public void bulkInsert(DocumentRevsList documentRevsList, Map<String[],List<PreparedAttachment>> preparedAttachments, boolean pullAttachmentsInline) throws DocumentException  {
+        for(DocumentRevs documentRevs: documentRevsList){
```
mikerhodes: erroneously removed a space here.

sync-core/src/test/java/com/cloudant/sync/datastore/AttachmentTest.java
```diff
@@ -83,7 +83,7 @@ public void setBadAttachmentsTest() throws Exception {
             rev1_mut.attachments.put(att3.name, att3);
             newRevision = datastore.updateDocumentFromRevision(rev1_mut);
             Assert.fail("FileNotFoundException not thrown");
-        } catch (FileNotFoundException fnfe) {
+        } catch (AttachmentException fnfe) {
```
mikerhodes: Rename the variable

sync-core/src/test/java/com/cloudant/sync/datastore/BasicDatastoreCRUDTest.java
```diff
@@ -140,21 +140,21 @@ private Map createMapWithSpecialField() {
...
+    public void createLocalDocument_docIdAndBody_success() throws Exception {
         String docId = CouchUtils.generateDocumentId();
         BasicDocumentRevision rev = datastore.createLocalDocument(docId, bodyOne);
         validateNewlyCreateLocalDocument(rev);
         Assert.assertEquals(docId, rev.getId());
     }
 
-    @Test(expected = IllegalArgumentException.class)
-    public void createLocalDocument_existDocId_exception() {
+    @Test(expected = DocumentException.class)
```
mikerhodes: Can any of these be more specific. I think they should check for the sub-class they expect to get if they can be more specific.

sync-core/src/test/java/com/cloudant/sync/datastore/BasicDatastoreConflictsTest.java
```diff
@@ -428,8 +427,8 @@ public void resurrectWinningDocument() throws IOException, ConflictException {
         Assert.assertEquals(deleted.getSequence(), resurrected.getParent());
     }
 
-    @Test(expected = IllegalArgumentException.class)
-    public void resurrectWinningDocumentFails() throws IOException, ConflictException {
+    @Test(expected = DocumentException.class)
```
mikerhodes: Just a note to make all of these expected exception clauses as exact as possible.

sync-core/src/test/java/com/cloudant/sync/indexing/IndexManagerIndexTest.java
```diff
@@ -718,8 +717,8 @@ public void run() {
                     rev.body = body;
                     try {
                         datastore.createDocumentFromRevision(rev);
-                    } catch (IOException ioe) {
-                        System.out.println("IOException "+ioe);
+                    } catch (Exception ioe) {
```
mikerhodes: Rename variable.

sync-core/src/test/java/com/cloudant/sync/datastore/MutableDocumentNoInitialAttachmentsTest.java
```diff
@@ -118,8 +118,8 @@ public void updateBodySetInvalidAttachments() throws ConflictException {
         BasicDocumentRevision updated = null;
         try {
             updated = datastore.updateDocumentFromRevision(update);
-            Assert.fail("Expected IOException; not thrown");
-        } catch (IOException ioe) {
+            Assert.fail("Expected AttachmentException; not thrown");
+        } catch (AttachmentException ioe) {
```
mikerhodes: Rename variable.

mikerhodes:
I think I'd have two root exceptions, one for Datastore and one for Document. The Datastore-style exceptions are generally of a very different magnitude to Document ones. I'd also consider throwing Datastore exceptions for things like SQL problems in most cases -- they mostly are problems with the whole datastore rather than documents, which are more conflicts and so on -- and if you get one SQLException chances are future operations on that datastore are going to fail.

What do you think @tomblench on the using Datastore failures for things like SQL/IO/etc exceptions?

```
- DatastoreException
  |
  |- DatastoreNotCreatedException

- DocumentException
  |
  |- InvalidDocumentException
  |- DocumentNotFoundException
  |- ConflictException 
  |
  |
  |- AttachmentException
  |     |
  |     |- AttachmentNotSavedException
 ```

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 13 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: @tomblench @mikerhodes How does `AttachmentException should not been thrown, this is a library programming error` sound?

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 15 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: Okay, I'll raise a fb ticket

File: sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java (hunk 20 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java]: From inspecting the code in the lib `FIleNotFoundException` is thrown when a check if performed to see if the `sourceFile` exists before attempting to move the file. Thinking about it, we should clean up when `IOException`

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 25 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: Honestly I don't know, the comment was there previously, I left it as it was, I will change though

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 26 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: Ideally, a user should never see this exception thrown, since the insert passed we fully expect the document to be found, so the exception shouldn't be throw. However we can't swallow the exception, an error will appear somewhere else. Perhaps the message should read `Could not get newly created document with id: %s ` ?

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 34 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: When deleting a revision the attachments for the previous revision shouldn't be copied, so the `AttachmentException` should never be thrown, unless someone changes `options.copyAttachments = false; ` to ` options.copyAttachments = true`. Perhaps  a better message is `Document was being deleted, attachments should not be copied to new revision` ?

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 37 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: I've always had the impression that getting the cause of the exception and throwing that is bad practise, perhaps if we have :
```
        try {
            int rowsDeleted = queue.submit(new SQLQueueCallable<Integer>() {
                @Override
                public Object call(SQLDatabase db) throws Exception {
                    String[] whereArgs = {docId};
                    return db.delete("localdocs", "docid=? ", whereArgs);
                }
            }).get();
           if (rowsDelete == 0 ) {
                throw new DocumentNotFoundException(String.format("Could not find document with id: %s",docId),e);
            throw new RuntimeException(e);
        } catch (ExecutionException e) {
           throw new DocumentException(e)
        }
``` 
it would avoid `DocumentNotFoundException` being the cause of a `DocumentNotFoundException`

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 38 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: I wish I knew what this referred to. I guess the original author assumed there would always be a log.  I'm not convinced this exception is right either, Do you have a view on what exception should be thrown?

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 41 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: From my understanding it was expected that `getDocumentInQueue` could return null for the document at the specified revision and it was safe to continue. Since now  `getDocumentInQueue` throws `DocumentNotFoundException` instead of returning null it's fine just to continue to the loop. I'll add in a comment to explain this.

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 42 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: If I recall correctly, if you follow the stack up, the replication will get terminated when the exception reaches the `ReplicationStrategy` so both the transaction and replication should get aborted

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 43 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: No there's not, I'll fix that now. The reason why it shouldn't be called on the queue is that the event bus notifies objects on the same thread that calls it, so a user could deadlock the library by attempting to get a new document while on the queue thread.

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 47 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: I don't know why, as far as I can tell from the schema we only add columns, so I assume it was to guard against programming errors

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 48 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: Java's standard stack traces normally look like 

```
some.exception.here: a message here
some.class.name(name.java:80)
....
caused by caused by message here
some.other.class.name(name.java 99)
```

so I think we can get away with not appending the error message

File: sync-core/src/main/java/com/cloudant/sync/datastore/DocumentException.java (hunk 49 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/DocumentException.java]: Its mostly for convenience of the developer, so if they want to treat all exceptions the same, they are free to do so without requiring them to use Java 7 language features

File: sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentNotSavedException.java (hunk 22 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentNotSavedException.java]: Looking at the code usages, its not really worth adding helper constructors, it looks like they will only be called once

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 31 above)
Comment (rhyshort) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: `createDocument` throws the attachment exception, to stop the datastore creating a document when it shouldn't error because it failed to load the parent https://github.com/cloudant/sync-android/blob/43599_improve_exceptions/sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java#L631

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 25 above)
Comment (mikerhodes) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: Looks to me like "normal" means "there's no document with that ID in the database". But I think we can leave it. It's clear enough from the code now I read it.

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 26 above)
Comment (mikerhodes) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: `RuntimeException` sounds like the right one, you're correct.

I'd be a little more panicy in the message, `Couldn't get document we just inserted (id: %s); this shouldn't happen, please file an issue with as much detail as possible.`.

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 31 above)
Comment (mikerhodes) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: I can't parse that sentence, " to stop the datastore creating a document when it shouldn't error" -- I think altering the use of commas might help?

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 33 above)
Comment (mikerhodes) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: I think suggest a fix too: "Document has newer revisions than the revision passed to delete; get the newest revision of the document and try again."

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 34 above)
Comment (mikerhodes) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: I'd suggest "This shouldn't happen because options.copyAttachments is set to false (as we don't do it for deletions), and AttachmentException should only occur during the copying of attachments."

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 38 above)
Comment (mikerhodes) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: Hopefully the SQL library will have logged an error. Right now we probably can't say something much better.

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 42 above)
Comment (mikerhodes) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: To me, the ReplicationStrategy should be the one logging that replication is terminated -- this code can't know that. I'd remove the "terminating replication".

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 44 above)
Comment (mikerhodes) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: Looks like we should leave this for now.

File: sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java (hunk 47 above)
Comment (mikerhodes) [inline sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java]: I assume at some point `cursor` may have been provided from a query that didn't include `parent` in the list of columns to get.

As far as I can see, this is never the case any more (though it's really quite broken that this method has to know the order of columns in the cursor and special case one).

I suggest assuming `parent` is present and deleting the `if` and seeing if the tests fail.

File: sync-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java [hunk 62]
```diff
@@ -313,7 +316,7 @@ private ItemsToPush missingRevisionsToJsonDocs(
         return allOpenRevisions;
     }
 
-    public String getReplicationId() {
+    public String getReplicationId() throws DatastoreException {
```
mikerhodes: Throwing this exception appears to terminate the replication, whereas in the pull case it just causes the putting of the checkpoint doc to fail and an error is logged, but the replication continues. Should the push also just log a warning if the checkpointing fails?

sync-core/src/test/java/com/cloudant/sync/datastore/BasicDatastoreCRUDTest.java
rhyshort: probably for this one no, unless we want to introduce conflict exception. But that brings the problem of localdocs having revisions when they shouldn't

sync-core/src/test/java/com/cloudant/sync/datastore/BasicDatastoreConflictsTest.java
rhyshort: Ideally it would be possible to do this, but since we have the queue, the cause exception gets wrapped in an execution exception

sync-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java
rhyshort: Yeah, I think it should do that, I've updated the code

mikerhodes:
Cool, looks like we're now good. +1 to merge.

PR #95, opened 2015-02-17
Fix NOT processing logic in query

alfinkel:
This PR addresses a bug that was found related to NOT processing in Query.  Using MongoDB query results as the "gold" standard and our baseline, changes were made to both the SQL engine and the post hoc matcher logic to fix NOT processing to essentially return results sets that contain documents that are NOT in the positive version of the query.  This is keeping in line with how MongoDB query handles NOT processing.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+
+# 0.11.0 (Unreleased)
+
+- [FIX] Using MongoDB query as the "gold" standard, query support for `NOT` has been fixed to return result sets correctly (as in MongoDB query).  Previously, the result set from a query like `{ "pet": { "$not" { "$eq": "dog" } } }` would include a document like `{ "pet" : [ "cat", "dog" ], ... }` because the array contains an object that isn't `dog`.  The new behavior is that `$not` now inverts the result set, so this document will no longer be included because it has an array element that matches `dog`.
+
+
+
 # 0.10.0 (2015-02-16)
 
 - [NEW] Added `listAllDatastores` method to `DatastoreManager`.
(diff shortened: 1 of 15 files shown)


CHANGES.md
```diff
@@ -1,3 +1,67 @@
...
+```
+
+This is different than the result set that comes from `{ "pet" : { "$lte": "dog" } }` (results below) which is the logical opposite to `{ "pet" : { "$gt": "dog" } }`.  Whereas `{ "pet" : { "$not" : { "$gt": "dog" } } }` is its "NOT-ted" opposite.
+
+```
+{ "_id" : "mike12", "name" : "mike", "age" : 12, "pet" : [ "cat", "dog" ] }
+{ "_id" : "mike34", "name" : "mike", "age" : 34, "pet" : [ "cat", "dog", "fish" ] }
+{ "_id" : "john22", "name" : "john", "age" : 22, "pet" : "cat" }
+```
+
```
mikerhodes: You need to succinctly describe the change here. At the moment this is more documentation. I think you should be able to do this with a single document example which before was matched and now isn't (or vice versa). Does:

> Previously, the query `{ "pet": { "$not" { "$eq": "dog" } } }` would match a document like `{ "pet" : [ "cat", "dog" ], ... }` because the array contains an object that isn't `dog`. The new behaviour is that `$not` inverts the result set, so this document would no longer be included because it has an element matching `dog`.

alfinkel: Fixed in 38b4605b75e2dead232963c2e68defe67afc8410

mikerhodes:
Cool, +1 after squashing.

PR #97, opened 2015-02-23
Move $not/$ne processing to normalization stage

alfinkel:
Move $not and $ne processing to be part of normalization instead of
having individual sets of logic to handle $not and $ne in both the
post hoc matcher and the SQL engine.  This allows us to keep things
less complex in the matcher and the SQL engine logic and also allows us
to maintain one set of logic for $not and $ne processing instead of two.

Changes made:

- Add logic to handle multiple $not operators in a row during
normalization.  This condition was previously unhandled.
- Add logic to transform shorthand notion such as $ne into the long
form.  In the case of $ne it would be $not..$eq.  All future shorthand
processing will be handled as part of normalization.
- Remove $not and $ne query manipulation from SQL translator logic.
- Remove $not and $ne query manipulation from post hoc matcher logic.
- Add additional tests to test the handling of sequential $not and $ne
operators.
- Remove WHERE clause tests that specifically use $ne in the query
since a query will never have an $ne operator in it by the time it
reaches the WHERE clause generation logic.

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/QueryValidator.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QueryValidator.java
@@ -26,10 +26,22 @@ import java.util.logging.Logger;
  */
 class QueryValidator {
 
-    public static final String AND = "$and";
-    public static final String OR = "$or";
-    public static final String EQ = "$eq";
+    private static final String AND = "$and";
+    private static final String OR = "$or";
+    private static final String EQ = "$eq";
+    private static final String NOT = "$not";
+    private static final String NE = "$ne";
 
+    // negatedShortHand is used for operator shorthand processing.
+    // A shorthand operator like $ne has a longhand representation
+    // that is { "$not" : { "$eq" : ... } }.  Therefore the negation
+    // of the $ne operator is $eq.
+    // Presently only $ne is supported.  More to come soon...
+    private static final Map<String, String> negatedShortHand = new HashMap<String, String>() {
+        {
+            put(NE,EQ);
+        }
+    };
     private static final Logger logger = Logger.getLogger(QueryValidator.class.getName());
 
     /**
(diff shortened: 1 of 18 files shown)


sync-core/src/main/java/com/cloudant/sync/query/QueryValidator.java
```diff
@@ -26,10 +26,19 @@
  */
 class QueryValidator {
 
-    public static final String AND = "$and";
-    public static final String OR = "$or";
-    public static final String EQ = "$eq";
-
+    private static final String AND = "$and";
```
tomblench: These strings also appear in QuerySqlTranslator.java and UnindexedMatcher.java - can we have them defined in one place?

sync-core/src/main/java/com/cloudant/sync/query/QueryValidator.java
```diff
@@ -146,6 +167,117 @@
...
+
+        for (Object fieldClause: clause) {
+            Object predicate;
+            String fieldName;
+            // if this isn't a map, we don't know what to do so add the clause
+            // to the accumulator to be dealt with later as part of the final selector
+            // validation.
+            if (fieldClause instanceof Map && !((Map) fieldClause).isEmpty()) {
+                Map<String, Object> fieldClauseMap = (Map<String, Object>) fieldClause;
+                fieldName = (String) fieldClauseMap.keySet().toArray()[0];
```
tomblench: this `keySet().toArray()[0];` pattern arises in lots of places. I guess because of the basic structure of mongo queries you only ever expect there to be one key.

I know it's a bit late in the game to raise this, but it seems a bit pointless to use Map<K,V> all over the place if there is only going to be one key. Maybe on iOS it makes a bit more sense because of the Map literal support. But really these instances are Pair<String,Object> (2-tuple or whatever you want to call it)

alfinkel: I think in iOS we have the same symptom as in the Java implementation.  In iOS we have the `allKeys[0]` pattern in as many places as `keySet().toArray()[0]` in Java.  Your point is well taken but I think whatever we choose to do going forward should be applied to both implementations for consistency and like you said, it is late in the game to make a change.

sync-core/src/main/java/com/cloudant/sync/query/QueryValidator.java
alfinkel: This is something thats been on my radar as well.  Its done in 24a70d3e1a88cbd2183cca6f5dea90cc598a0d5f.

PR #98, opened 2015-02-26
Set library version in sample to latest.release

rhyshort:
Change the sample build file to dynamic resolve the latest released version of sync-android. `latest.release` is a place holder for the latest released version of a dependancy. Making this change reduces the number of steps involved in a release.


tomblench:
+1

PR #100, opened 2015-03-02
Retire legacy indexing code

alfinkel:
This PR retires the previous indexing code base out of this repository and provides a migration document for existing users of the deprecated code.  Going forward any users of the legacy indexing code will need to use the new query code that replaces the legacy implementation along with the migration document to update their implementations.


mikerhodes:
This looks good. I think we can merge as is, +1.

PR #103, opened 2015-03-04
Move $not/$ne processing to normalization stage

alfinkel:
Move $not and $ne processing to be part of normalization instead of
having individual sets of logic to handle $not and $ne in both the
post hoc matcher and the SQL engine. This allows us to keep things
less complex in the matcher and the SQL engine logic and also allows us
to maintain one set of logic for $not and $ne processing instead of two.

Changes made:

Add logic to handle multiple $not operators in a row during normalization. This condition was previously unhandled.
Add logic to transform shorthand notion such as $ne into the long form. In the case of $ne it would be $not..$eq. All future shorthand processing will be handled as part of normalization.
Remove $not and $ne query manipulation from SQL translator logic.
Remove $not and $ne query manipulation from post hoc matcher logic.
Add additional tests to test the handling of sequential $not and $ne operators.
Remove WHERE clause tests that specifically use $ne in the query since a query will never have an $ne operator in it by the time it reaches the WHERE clause generation logic.

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/QueryConstants.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QueryConstants.java
@@ -1,0 +1,33 @@
+//  Copyright (c) 2015 Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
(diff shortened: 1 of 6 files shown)


mikerhodes:
+1 when travis is happy.

PR #104, opened 2015-03-04
Add support for $in and $nin operators in query

alfinkel:
This PR adds the $in and $nin operator support to Query.  Additionally, there was some code clean up and  refactoring done to the SQL Translator logic to aid in future maintenance, readability and some SQL optimization.

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
@@ -423,86 +423,66 @@ class QuerySqlTranslator {
                     boolean exists = !((Boolean) predicateValue);
                     // since this clause is negated we need to negate the bool value
                     whereClauses.add(convertExistsToSqlClauseForFieldName(fieldName, exists));
-                } else if (operator.equals(IN)) {
-                    String sqlOperator = operatorMap.get(operator);
-                    String tableName = IndexManager.tableNameForIndex(indexName);
-                    String inOperands = "(";
-
-                    // The predicate map value must be a List here.
-                    // This was validated during normalization.
-                    List<Object> inList = (List<Object>) negatedPredicate.get(operator);
-                    for (Object negatedPredicateValue : inList) {
-                        if (validatePredicateValue(negatedPredicateValue)) {
-                            if (inOperands.endsWith("?")) {
-                                inOperands = String.format("%s, ?", inOperands);
-                            } else {
-                                inOperands = String.format("%s ?", inOperands);
-                            }
-                            sqlParameters.add(String.valueOf(negatedPredicateValue));
-                        } else {
-                            logger.log(Level.SEVERE, "Predicate value is invalid.");
-                            return null;
-                        }
-                    }
-                    inOperands = String.format("%s )", inOperands);
-                    String whereClause = whereClauseForNot(fieldName,
-                                                           sqlOperator,
-                                                           tableName,
-                                                           inOperands);
-                    whereClauses.add(whereClause);
                 } else {
+                    String whereClause;
                     String sqlOperator = operatorMap.get(operator);
                     String tableName = IndexManager.tableNameForIndex(indexName);
-                    String whereClause = whereClauseForNot(fieldName, sqlOperator, tableName, "?");
-
-                    whereClauses.add(whereClause);
-                    predicateValue = negatedPredicate.get(operator);
-                    if (validatePredicateValue(predicateValue)) {
-                        sqlParameters.add(String.valueOf(predicateValue));
+                    String argument;
+                    if (operator.equals(IN)) {
+                        // The predicate map value must be a List here.
+                        // This was validated during normalization.
+                        List<Object> inList = (List<Object>) negatedPredicate.get(operator);
+                        argument = buildWhereClauseArgument(inList, sqlParameters);
+                        // If the argument is a single "?" placeholder then we are
+                        // dealing with a single equality comparison.  Therefore,
+                        // we set the operator to $eq.
+                        if (argument != null && argument.equals("?")) {
+                            sqlOperator = operatorMap.get(EQ);
+                        }
                     } else {
-                        logger.log(Level.SEVERE, "Predicate value is invalid.");
+                        predicateValue = negatedPredicate.get(operator);
+                        argument = buildWhereClauseArgument(Arrays.asList(predicateValue),
+                                                            sqlParameters);
+                    }
+                    if (argument == null) {
+                        // An error occurred and logged in buildWhereClauseArgument
                         return null;
                     }
+                    whereClause = whereClauseForNot(fieldName, sqlOperator, tableName, argument);
+                    whereClauses.add(whereClause);
                 }
-
             } else {
                 if (operator.equals(EXISTS)) {
                     boolean exists = (Boolean) predicate.get(operator);
                     whereClauses.add(convertExistsToSqlClauseForFieldName(fieldName, exists));
-                } else if (operator.equals(IN)) {
-                    String sqlOperator = operatorMap.get(operator);
-                    String whereClause = String.format("\"%s\" %s (", fieldName, sqlOperator);
-
-                    // The predicate map value must be a List here.
-                    // This was validated during normalization
-                    List<Object> inList = (List<Object>) predicate.get(operator);
-                    for (Object predicateValue : inList) {
-                        if (validatePredicateValue(predicateValue)) {
-                            if (whereClause.endsWith("?")) {
-                                whereClause = String.format("%s, ?", whereClause);
-                            } else {
-                                whereClause = String.format("%s ?", whereClause);
-                            }
-                            sqlParameters.add(String.valueOf(predicateValue));
-                        } else {
-                            logger.log(Level.SEVERE, "Predicate value is invalid.");
-                            return null;
-                        }
-                    }
-                    whereClause = String.format("%s )", whereClause);
-                    whereClauses.add(whereClause);
                 } else {
+                    String whereClause;
                     String sqlOperator = operatorMap.get(operator);
-                    String whereClause = String.format("\"%s\" %s ?", fieldName, sqlOperator);
-
-                    whereClauses.add(whereClause);
-                    Object predicateValue = predicate.get(operator);
-                    if (validatePredicateValue(predicateValue)) {
-                        sqlParameters.add(String.valueOf(predicateValue));
+                    String argument;
+                    if (operator.equals(IN)) {
+                        // The predicate map value must be a List here.
+                        // This was validated during normalization.
+                        List<Object> inList = (List<Object>) predicate.get(operator);
+                        argument = buildWhereClauseArgument(inList, sqlParameters);
+                        // If the argument is a single "?" placeholder then we are
+                        // dealing with a single equality comparison.  Therefore,
+                        // we set the operator to $eq.
+                        if (argument != null && argument.equals("?")) {
+                            sqlOperator = operatorMap.get(EQ);
+                        }
                     } else {
-                        logger.log(Level.SEVERE, "Predicate value is invalid.");
+                        Object predicateValue = predicate.get(operator);
+                        argument = buildWhereClauseArgument(Arrays.asList(predicateValue),
+                                                            sqlParameters);
+                    }
+                    if (argument == null) {
+                        // An error occurred and logged in buildWhereClauseArgument
                         return null;
                     }
+                    whereClause = String.format("\"%s\" %s %s", fieldName,
+                                                                sqlOperator,
+                                                                argument);
+                    whereClauses.add(whereClause);
                 }
             }
         }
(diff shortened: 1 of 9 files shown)


sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -423,36 +424,65 @@ protected static SqlParts whereSqlForAndClause(List<Object> clause, String index
...
+                        // The predicate map value must be a List here.
+                        // This was validated during normalization.
+                        List<Object> inList = (List<Object>) negatedPredicate.get(operator);
+                        argument = buildWhereClauseArgument(inList, sqlParameters);
+                        // If the argument is a single "?" placeholder then we are
+                        // dealing with a single equality comparison.  Therefore,
+                        // we set the operator to $eq.
+                        if (argument != null && argument.equals("?")) {
+                            sqlOperator = operatorMap.get(EQ);
+                        }
```
mikerhodes: I'm not sure this special casing for the single case is worthwhile, and it seems a bit fragile to me.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -468,6 +498,27 @@ protected static SqlParts whereSqlForAndClause(List<Object> clause, String index
         return SqlParts.partsForSql(where, parameterArray);
     }
 
+    private static String buildWhereClauseArgument(List<Object> values,
```
mikerhodes: Seems like a better name would be `buildWhereClausePlaceholders`.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -468,6 +498,27 @@ protected static SqlParts whereSqlForAndClause(List<Object> clause, String index
...
+        List<String> inOperands = new ArrayList<String>();
+        for (Object value : values) {
+            if (validatePredicateValue(value)) {
+                inOperands.add("?");
+                sqlParameters.add(String.valueOf(value));
+            } else {
+                logger.log(Level.SEVERE, "Predicate value is invalid.");
+                return null;
+            }
+        }
```
mikerhodes: Should we note that this validation needs moving to the validator?

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -468,6 +498,27 @@ protected static SqlParts whereSqlForAndClause(List<Object> clause, String index
...
+
+        if (inOperands.size() == 1) {
+            return inOperands.get(0);
+        } else {
+            Joiner opJoiner = Joiner.on(", ").skipNulls();
+            return String.format("( %s )", opJoiner.join(inOperands));
+        }
```
mikerhodes: I think this is a bit of premature optimisation too.

sync-core/src/main/java/com/cloudant/sync/query/QueryValidator.java
```diff
@@ -457,18 +471,12 @@ private static boolean validateExistsArgument(Object exists) {
...
-            }
-        } else if (value instanceof List) {
-            for (Object element : (List) value) {
-                valid = valid && validateQueryValue(element);
-            }
-        } else if (value instanceof Float) {
+
+        if (value instanceof Float) {
             valid = false;
+            logger.log(Level.SEVERE, "Float value encountered in query");
```
mikerhodes: Give some help: "Float value found in query; use Double instead"

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
```diff
@@ -258,7 +265,7 @@ private boolean executeSelectorTree(QueryNode node, DocumentRevision rev) {
     private boolean valueCompare(Object actual, String operator, Object expected) {
         boolean passed;
 
-        if (operator.equals("$eq")) {
+        if (operator.equals("$eq") || operator.equals("$in")) {
```
mikerhodes: Might as well move this to the constants if you're changing the line.

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
```diff
@@ -233,17 +234,23 @@ private boolean executeSelectorTree(QueryNode node, DocumentRevision rev) {
                 operator = (String) operatorExpression.keySet().toArray()[0];
             }
 
-            Object expected = operatorExpression.get(operator);
+            Object expectedValues = operatorExpression.get(operator);
+            if (!(expectedValues instanceof List)) {
+                expectedValues = Arrays.asList(expectedValues);
+            }
```
mikerhodes: Might be worth a note here that you are doing this so $in can be treated like $eq and we don't have to special case $in (which I like).

mikerhodes: If you're considering this, might be worth doing it with `actual` too, so you're using a consistent pattern (and the `for` loop becomes simpler).

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
mikerhodes: If we're supporting this, we might want to consider what happens if we use too many placeholders (SQLite has a limit per query).

See (9) at https://www.sqlite.org/limits.html. Previously we've assumed 500 max to be on the safe side.

sync-core/src/test/java/com/cloudant/sync/query/QuerySqlTranslatorTest.java
```diff
@@ -889,12 +919,31 @@ public void usesCorrectSQLOperatorWhenUsingNOTLTE() {
...
+        notOp.put("$not", op);
+        Map<String, Object> name = new HashMap<String, Object>();
+        name.put("name", notOp);
+
+        SqlParts where = QuerySqlTranslator.whereSqlForAndClause(Arrays.<Object>asList(name),
+                                                                 indexName);
+        String expectedSubWhere = "WHERE \"name\" IN ( ?, ? ))";
+        String expected = String.format("_id NOT IN (SELECT _id FROM %s %s",
+                                        indexTable,
+                                        expectedSubWhere);
```
mikerhodes: I'd prefer to see the whole sub-select statement as a separate variable. At the very least, have the balancing bracket for the `SELECT` in the second string.

sync-core/src/test/java/com/cloudant/sync/query/QueryValidatorTest.java
```diff
@@ -301,4 +390,24 @@ public void returnsNullForInvalidOperatorWithNOT() {
...
 
+    @Test
+    public void returnsNullForNOTWithoutOperator() {
+        Map<String, Object> query = new HashMap<String, Object>();
+        // query - { "name" : { "$not" : "mike" } }
+        Map<String, Object> notMike = new HashMap<String, Object>();
+        notMike.put("$not", "mike");
+        query.put("name", notMike);
+        assertThat(QueryValidator.normaliseAndValidateQuery(query), is(nullValue()));
+    }
```
mikerhodes: Curious: does Mongo treat this as `$not: $eq`, or is it invalid there too?

sync-core/src/test/java/com/cloudant/sync/query/UnindexedMatcherTest.java
```diff
@@ -1078,17 +1078,119 @@ public void matchBadItemWithNot() {
...
+        Map<String, Object> op = new HashMap<String, Object>();
+        op.put("$in", Arrays.<Object>asList("mike", "fred"));
+        Map<String, Object> notOp = new HashMap<String, Object>();
+        notOp.put("$not", op);
+        Map<String, Object> selector = new HashMap<String, Object>();
+        selector.put("name", notOp);
+        selector = QueryValidator.normaliseAndValidateQuery(selector);
+        UnindexedMatcher matcher = UnindexedMatcher.matcherWithSelector(selector);
+        assertThat(matcher.matches(rev), is(false));
+    }
```
mikerhodes: Do we need any `$nin` specific tests here?

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
mikerhodes: A comment here saying something like "$in implies $eq for each value in its array (caller breaks array into separate values)" could be useful, as it's a bit confusing given you'd expect `expected` to be an array if you were reading in isolation.

mikerhodes: Or, perhaps do the flip of operator to $in in the caller itself where you're doing other pieces of transformation, meaning $in logic is more consolidated in one place.

sync-core/src/test/java/com/cloudant/sync/query/QueryValidatorTest.java
alfinkel: It is invalid in Mongo as well.

sync-core/src/test/java/com/cloudant/sync/query/UnindexedMatcherTest.java
alfinkel: Any occurrence of `$nin` is transformed to `{ "$not" : { "$in" : ... } }` during normalization.  Tests exist to validate that that transformation always occurs.  Having a `$nin` test here would really only be re-testing the normalization/validation process and would pretty much be the same as having a `{ "$not" : { "$in" : ... } }` test of the Matcher logic which we already have.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: Removed in 5b5040e6a64ad6720ef17ace362f2020397d0880.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: Changed `buildWhereClauseArgument` to `placeholdersForInList` as the method has become more clear/concise due to the removal of redundant predicate validation.  Done in 5b5040e6a64ad6720ef17ace362f2020397d0880.

alfinkel: Created an FB defect for placeholder limit logic to be handled at a later time:  https://cloudant.fogbugz.com/f/cases/45063/Add-placeholder-limit-logic-to-where-clause-generation

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: This is actually redundant validation as it already exists in the validator.  `validateClause` https://github.com/cloudant/sync-android/blob/44270-query-in-nin-operator-support/sync-core/src/main/java/com/cloudant/sync/query/QueryValidator.java#L409 in the validator validates each clause which includes predicate validation using the `validatePredicateValue` method.  This validation in the translator has been removed in 5b5040e6a64ad6720ef17ace362f2020397d0880.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: Removed and simplified in 5b5040e6a64ad6720ef17ace362f2020397d0880.

sync-core/src/main/java/com/cloudant/sync/query/QueryValidator.java
alfinkel: Added additional commentary to message.  Done in 5b5040e6a64ad6720ef17ace362f2020397d0880.

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
alfinkel: Replaced all literals with constants in the `if-else if...` block.  Flipped `$in` to `$eq` in the calling routine.  Done in 5b5040e6a64ad6720ef17ace362f2020397d0880.

sync-core/src/main/java/com/cloudant/sync/query/UnindexedMatcher.java
alfinkel: Done in 5b5040e6a64ad6720ef17ace362f2020397d0880.

sync-core/src/test/java/com/cloudant/sync/query/QuerySqlTranslatorTest.java
alfinkel: Done in 5b5040e6a64ad6720ef17ace362f2020397d0880.

mikerhodes:
Okay, I think we're good here. The changes in 5b5040e need to be split up and either exist as separate commits, if useful, or squashed into the appropriate commits from earlier. E.g., simplifying code due to being able to make the assumption that things are certain types would probably be separate commits, as would QueryConstants stuff.

mikerhodes:
+1 once Travis is happy.

PR #105, opened 2015-03-05
Support all test command line options on android

rhyshort:
Removed android specific couch configuration from tests, the test runner
now handles the conversion. Android will always use the specified
couch config.

build.gradle adds all commandline options with their defaults, exception
to this is username and password which has defaults of empty strings.

### DIFF
--- a/AndroidTest/app/build.gradle
+++ b/AndroidTest/app/build.gradle
@@ -3,6 +3,7 @@ apply plugin: 'com.android.application'
 def testResultsFile = "testResults_"+UUID.randomUUID().toString()+".xml"
 def scriptLocation = project.projectDir.getAbsolutePath(); //this is the location of THIS file
 def testSrcDir = "../../sync-core/src/test/java/" // this is the location of your test code, relative to this file
+def testConfig = convertTestSysPropsToHash()
 
 android {
     def classes = findAllTestClasses(new File("$scriptLocation/$testSrcDir")).toString().replace("[","{").replace("]","}");
@@ -19,6 +20,11 @@ android {
         buildConfigField "Class[]", "classToTests", "$classes"
         buildConfigField "String", "testOutputFile", "\"$testResultsFile\""
         buildConfigField "Class[]", "testExcludes", "{com.cloudant.common.SystemTest.class, com.cloudant.common.RequireRunningCouchDB.class, com.cloudant.common.PerformanceTest.class}"
+
+        //pass test config as key value pairs
+
+        buildConfigField "String[][]", "TEST_CONFIG", testConfig
+
     }
 
     productFlavors {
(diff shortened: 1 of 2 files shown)


sync-core/src/test/java/com/cloudant/common/CouchTestBase.java
```diff
@@ -38,19 +38,12 @@
 
     public CouchConfig getCouchConfig(String db) {
 
+        //Android is always a specified couch via the AndroidTest runner
```
tomblench: isn't this comment misleading - doesn't SPECIFIED_COUCH have uses outside of android testing?

rhyshort: Yes it does, I'll clarify the comment

AndroidTest/app/build.gradle
```diff
@@ -103,6 +108,24 @@ task(uploadFixtures,type:Exec) {
     commandLine "$android/platform-tools/adb","push", "-p", "$scriptLocation/../../fixture", "/sdcard/fixture"
 }
 
+def convertTestSysPropsToHash (){
```
tomblench: Here is a slightly more 'groovy' (!) way to do it (I'm not too sure on my indenting):
```
def convertTestSysPropsToHash2 (){
    return "{" +
        System.properties.grep { prop ->
        prop.key.startsWith("test") }
        .collect(new ArrayList()) { prop -> "{\""+prop.key+"\",\""+ prop.value+"\"}"}
        .join(",") +
        "}"
}
```

AndroidTest/app/build.gradle
```diff
@@ -19,6 +20,10 @@ android {
         buildConfigField "Class[]", "classToTests", "$classes"
         buildConfigField "String", "testOutputFile", "\"$testResultsFile\""
         buildConfigField "Class[]", "testExcludes", "{com.cloudant.common.SystemTest.class, com.cloudant.common.RequireRunningCouchDB.class, com.cloudant.common.PerformanceTest.class}"
+
+
+        buildConfigField "String[][]", "TEST_CONFIG", testConfig
```
tomblench: Maybe a comment like "pass test config as an array of key/value pairs"

AndroidTest/app/build.gradle
```diff
@@ -103,6 +109,17 @@ task(uploadFixtures,type:Exec) {
     commandLine "$android/platform-tools/adb","push", "-p", "$scriptLocation/../../fixture", "/sdcard/fixture"
 }
 
+def convertTestSysPropsToHash (){
+
```
tomblench: How about indenting it this way (with the dots lining up)?
```
    return "{" +
        System.properties
        .grep { prop -> prop.key.startsWith("test") }
        .collect(new ArrayList()) { prop -> "{\""+prop.key+"\",\""+ prop.value+"\"}"}
        .join(",") +
        "}"
}
```

tomblench:
+1

PR #107, opened 2015-03-16
Stablize BasicReplicatorMockTests

rhyshort:
Stablize basic replicator mock tests, by increasing the window to
check if the thread is alive, by mocking the run method to
wait 1 second before returning

### DIFF
--- a/sync-core/src/test/java/com/cloudant/sync/replication/BasicReplicatorMockTest.java
+++ b/sync-core/src/test/java/com/cloudant/sync/replication/BasicReplicatorMockTest.java
@@ -22,6 +22,8 @@ import com.google.common.eventbus.Subscribe;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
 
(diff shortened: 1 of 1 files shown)


sync-core/src/test/java/com/cloudant/sync/replication/BasicReplicatorMockTest.java
```diff
@@ -256,8 +258,17 @@ public void start_stoppedThenRestarted() throws Exception {
         verify(mockListener).complete(rc);
         verify(mockListener, never()).error(any(ReplicationErrored.class));
 
+        doAnswer(new Answer() {
```
tomblench: Can this be done in the setUp() block?

rhyshort: I'm not really sure, I didn't want to mess with the other tests since I'm not sure how the set up their replicator to produce errors or if the a mocked class methods just return.

tomblench: OK I guess the other tests don't need it and it could slow them down.
+1

PR #109, opened 2015-03-27
Add Index class to manage index attribute validation

alfinkel:
This PR introduces an Index class which will be used by the IndexManager, IndexCreator, and IndexUpdater to set up Index objects for eventual index creation and modification.  The hooks into the IndexManager, IndexCreator, and IndexUpdater for the Index class will be implemented in a separate PR.  This PR focuses on the following:
 
- Add Index.java and add field names, index name, index type, and
index settings as attributes.
- Add "set up" validation for all attributes.
- Add `compareIndexTypeTo` method that will be used by the IndexCreator
to compare an Index's type and settings with already existing indexes
in our SQLite tables.
- Add IndexTest.java to test Index class functionality.

BugID: 45745

reviewer: @mikerhodes 
reviewer: @rhyshort

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/Index.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/Index.java
@@ -1,0 +1,187 @@
+//  Copyright (c) 2015 Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+
+package com.cloudant.sync.query;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * This class provides functionality to manage an index
+ */
+class Index {
+
+    private static final Logger logger = Logger.getLogger(Index.class.getCanonicalName());
+
+    public static final String JSON_TYPE = "json";
+
+    public static final String TEXT_TYPE = "text";
+
+    private static final List<String> validTypes = Arrays.asList(JSON_TYPE, TEXT_TYPE);
+
+    private static final String TEXT_TOKENIZE = "tokenize";
+
+    private static final String TEXT_DEFAULT_TOKENIZER = "simple";
+
+    private static final List<String> validSettings = Arrays.asList(TEXT_TOKENIZE);
+
+    protected final List<Object> fieldNames;
+
+    protected final String indexName;
+
+    protected final String indexType;
+
+    protected final Map<String, String> indexSettings;
+
+    private ObjectMapper objectMapper;
+
+    private Index(List<Object> fieldNames,
+                  String indexName,
+                  String indexType,
+                  Map<String, String> indexSettings) {
+        this.fieldNames = fieldNames;
+        this.indexName = indexName;
+        this.indexType = indexType;
+        this.indexSettings = indexSettings;
+    }
+
+    /**
+     * This method sets the index type to the default setting of "json"
+     *
+     * @param fieldNames the field names in the index
+     * @param indexName the index name
+     * @return the Index object or null if arguments passed in were invalid.
+     */
+    public static Index setUp(List<Object> fieldNames, String indexName) {
+        return setUp(fieldNames, indexName, JSON_TYPE);
+    }
+
+    public static Index setUp(List<Object> fieldNames, String indexName, String indexType) {
+        return setUp(fieldNames, indexName, indexType, null);
+    }
+
+    /**
+     * The setUp method handles index specific validation and ensures that the constructed
+     * Index object is valid.
+     *
+     * @param fieldNames the field names in the index
+     * @param indexName the index name
+     * @param indexType the index type (json or text)
+     * @param indexSettings the optional settings used to configure the index.
+     *                      Only supported parameter is 'tokenize' for text indexes only.
+     * @return the Index object or null if arguments passed in were invalid.
+     */
+    public static Index setUp(List<Object> fieldNames,
+                              String indexName,
+                              String indexType,
+                              Map<String, String> indexSettings) {
+        if (fieldNames == null || fieldNames.isEmpty()) {
+            logger.log(Level.SEVERE, "No field names were provided.");
+            return null;
+        }
+
+        if (indexName == null || indexName.isEmpty()) {
+            logger.log(Level.SEVERE, "No index name was provided.");
+            return null;
+        }
+
+        if (indexType == null || indexType.isEmpty()) {
+            logger.log(Level.SEVERE, "No index type was provided.");
+            return null;
+        }
+
+        if (!validTypes.contains(indexType.toLowerCase())) {
+            logger.log(Level.SEVERE, String.format("Invalid index type %s.", indexType));
+            return null;
+        }
+
+        if (indexType.equalsIgnoreCase(JSON_TYPE) && indexSettings != null) {
+            logger.log(Level.WARNING, String.format("Index type is %s, index settings %s ignored.",
+                                                    indexType,
+                                                    indexSettings.toString()));
+            indexSettings = null;
+        } else if (indexType.equalsIgnoreCase(TEXT_TYPE)) {
+            if (indexSettings == null) {
+                indexSettings = new HashMap<String, String>();
+                indexSettings.put(TEXT_TOKENIZE, TEXT_DEFAULT_TOKENIZER);
+                logger.log(Level.FINE, String.format("Index type is %s, defaulting settings to %s.",
+                        indexType,
+                        indexSettings.toString()));
+            } else {
+                for (String parameter : indexSettings.keySet()) {
+                    if (!validSettings.contains(parameter.toLowerCase())) {
+                        String msg = String.format("Invalid parameter %s in index settings %s.",
+                                                   parameter,
+                                                   indexSettings);
+                        logger.log(Level.SEVERE, msg);
+                        return null;
+                    }
+                }
+            }
+        }
+
+        return new Index(fieldNames, indexName, indexType, indexSettings);
+    }
+
+    /**
+     * Compares the index type and accompanying settings with the passed in arguments.
+     *
+     * @param indexType the index type to compare to
+     * @param indexSettings the indexSettings to compare to
+     * @return true/false - whether there is a match
+     */
+    protected boolean compareIndexTypeTo(String indexType, String indexSettings) {
+        if (!this.indexType.equalsIgnoreCase(indexType)) {
+            return false;
+        }
+
+        if (this.indexSettings == null && indexSettings == null) {
+            return true;
+        } else if (this.indexSettings == null || indexSettings == null) {
+            return false;
+        }
+
+        Map<String, Object> settings;
+        try {
+            settings = getObjectMapper().readValue(indexSettings,
+                    new TypeReference<Map<String, Object>>() {
+                    });
+        } catch (IOException e) {
+            String msg = String.format("Error processing index settings %s",
+                                       this.indexSettings.toString());
+            logger.log(Level.SEVERE, msg, e);
+            return false;
+        }
+
+        // We perform a deep comparison of hash maps to ensure that both objects
+        // and any sub-objects are equal regardless of order within the maps.
+        return this.indexSettings.equals(settings);
+    }
+
+    private ObjectMapper getObjectMapper() {
+        if (objectMapper == null) {
+            objectMapper = new ObjectMapper();
+        }
+        return objectMapper;
+    }
+
+}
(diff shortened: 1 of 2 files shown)


sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * This class provides functionality to manage an index
+ */
+public class Index {
+
+    private static final Logger logger = Logger.getLogger(Index.class.getName());
```
rhyshort: `Index.class.getCanonicalName()` should be used instead of `Index.class.getName()`

sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -0,0 +1,165 @@
...
+     * @param fieldNames the field names in the index
+     * @param indexName the index name
+     * @param indexType the index type (json or text)
+     * @param indexSettings the optional settings used to configure the index.
+     *                      Only supported parameter is 'tokenize' for text indexes only.
+     * @return the Index object or null if arguments passed in were invalid.
+     */
+    public static Index setUp(List<Object> fieldNames,
+                              String indexName,
+                              String indexType,
```
rhyshort: I think the index type should be an enumeration, it will save us some validation code since we won't need to check if the index type given to us is a valid index type

sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -0,0 +1,165 @@
...
+                              Map<String, String> indexSettings) {
+        if (fieldNames == null || fieldNames.isEmpty()) {
+            logger.log(Level.SEVERE, "No field names were provided.");
+            return null;
```
rhyshort: I don't think we should be logging an error when the user passes us information which invalid. I would also consider this being a programming error, I would throw an InvalidArgumentException here and for the similar expressions below

sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /**
+     * Compares the index type and accompanying settings with the passed in arguments.
+     *
+     * @param indexType the index type to compare to
+     * @param indexSettings the indexSettings to compare to
+     * @return true/false - whether there is a match
+     */
+    protected boolean compareIndexTypeTo(String indexType, String indexSettings) {
```
rhyshort: How is this method going to be used? I'm wondering if it would be better for this class to implement comparable instead.

sync-core/src/test/java/com/cloudant/sync/query/IndexTest.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.Map;
+
+public class IndexTest {
+
+    private List<Object> fieldNames;
+    private String indexName;
+
+    @Before
+    public void setUp() {
+        fieldNames = Arrays.<Object>asList("name", "age");
```
rhyshort: What is the reason for having the fieldNames as an Object list rather than a String list?

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: Good point.  I will need to do this throughout the query code. For now I did it in this Index class and added an FB ticket for the remainder of the query code base (https://cloudant.fogbugz.com/f/cases/46006/Change-all-getName-logger-references-to-genCanonicalName).  Done in 88ea555325a939d2f755d7fb6fc666ff12e35289

sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -0,0 +1,165 @@
...
+
+    public static final String JSON_TYPE = "json";
+
+    public static final String TEXT_TYPE = "text";
+
+    private static final List<String> validTypes = Arrays.asList(JSON_TYPE, TEXT_TYPE);
+
+    private static final String TOKENIZE = "tokenize";
+
+    private static final String DEFAULT_TOKENIZER = "simple";
```
mikerhodes: Can you namespace these, that is prefix with `TEXT_`?

sync-core/src/main/java/com/cloudant/sync/query/Index.java
mikerhodes: Currently we are returning `null` for something went wrong in the Query code. Latterly, we will look to address how we notify of errors to improve this.

Note Al is logging at `SEVERE` level, which I think is the same as logging an error?

sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -0,0 +1,165 @@
...
+                                                    indexType,
+                                                    indexSettings.toString()));
+            indexSettings = null;
+        } else if (indexType.equalsIgnoreCase(TEXT_TYPE)) {
+            if (indexSettings == null) {
+                indexSettings = new HashMap<String, String>();
+                indexSettings.put(TOKENIZE, DEFAULT_TOKENIZER);
+                logger.log(Level.INFO, String.format("Index type is %s, defaulting settings to %s.",
+                        indexType,
+                        indexSettings.toString()));
```
mikerhodes: Debug level I'd say

sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -0,0 +1,165 @@
...
+
+        if (this.indexSettings == null && indexSettings == null) {
+            return true;
+        } else if (this.indexSettings == null || indexSettings == null) {
+            return false;
+        }
+
+        String settingsAsAString;
+        try {
+            settingsAsAString = new ObjectMapper().writeValueAsString(this.indexSettings);
```
mikerhodes: Should the objectmapper be cached? Not sure whether this initialisation is slow?

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: I agree with your point on enumeration however, this `Index` class is meant to, in part, be a way to declutter the `IndexManager`, `IndexCreator`, and `IndexUpdater` code.  At present the IndexManager public methods are our interface to the end user.  The IndexManager `ensureIndexed` methods allow for the end user to provide an indexType of String.  In the future it might make sense to lock this down to an enum.  For now I'd like to leave it as it is and revisit this later.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -0,0 +1,165 @@
...
+        } catch (JsonProcessingException e) {
+            String msg = String.format("Error processing index settings %s",
+                                       this.indexSettings.toString());
+            logger.log(Level.SEVERE, msg, e);
+            return false;
+        }
```
mikerhodes: Your problem here might be in ordering semantics, the ordering of keys in the JSON from your indexSettings map might be different from the string passed in, even though the structures are semantically the same. I.e.,:

`{ "name: "mike", "age": 12}` vs. `{"age": 12, "name": "mike"}`.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -0,0 +1,165 @@
...
+                  String indexType,
+                  Map<String, String> indexSettings) {
+        this.fieldNames = fieldNames;
+        this.indexName = indexName;
+        this.indexType = indexType;
+        this.indexSettings = indexSettings;
+    }
+
+    public static Index setUp(List<Object> fieldNames, String indexName) {
+        // JSON index type is the default
```
mikerhodes: Could just write JavaDoc that indicates this.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * This class provides functionality to manage an index
+ */
+public class Index {
```
mikerhodes: Could this be default, package-private?

sync-core/src/test/java/com/cloudant/sync/query/IndexTest.java
```diff
@@ -0,0 +1,135 @@
...
+    public void comparesIndexSettingsAndReturnsInEquality() {
+        Index index = Index.setUp(fieldNames, indexName, "text");
+        assertThat(index.compareIndexTypeTo("text", "{\"tokenize\":\"porter\"}"), is(false));
+    }
+
+    @Test
+    public void comparesIndexSettingsAndReturnsEquality() {
+        Index index = Index.setUp(fieldNames, indexName, "text");
+        assertThat(index.compareIndexTypeTo("text", "{\"tokenize\":\"simple\"}"), is(true));
+    }
```
mikerhodes: Related to a comment above, as `compareIndexTypeTo` takes a string, I would have comparison of semantically equivalent but different orderings:

```java
public void comparesIndexSettingsAndReturnsEquality() {
    Map indexSettings = /* Set up indexSettings Map, { "tokenise": "simple", "foo": "bar" } */;
    Index index = Index.setUp(fieldNames, indexName, "text", indexSettings);
    assertThat(index.compareIndexTypeTo("text", "{\"tokenize\":\"simple\", \"foo\":\"bar\"}"), is(true));
    assertThat(index.compareIndexTypeTo("text", "{\"foo\":\"bar\", \"tokenize\":\"simple\"}"), is(true));
}
```

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: That is correct.  The logging of errors here is meant to replace the logic that was formerly in the IndexManager that performed the same logging of errors.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: Right now this method provides a way (again) to declutter the IndexCreator logic that checks whether an index already exists or needs to be created.  It could have also been written as a method within the IndexCreator itself.  I think that as the Index class evolves it may wind up that we change it to implement Comparable.  But a valid point.

sync-core/src/test/java/com/cloudant/sync/query/IndexTest.java
alfinkel: The list of fieldNames can be a String list or it can be a list that contains Map objects that provide the filename as well as an index sort direction.  For example: a Map object could be `{ "fieldname" : "DESC" }` or `{ "filename" : "ASC" }`.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: There does not appear to be a Debug level.

rhyshort: There isn't, but when we converted to java.util.logging, I adopted using `Level.FINE` as the debug level

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: Done in 2ffd47f.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: Changed to FINE in 2ffd47f.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: Done in 2ffd47f.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: Changed the method to convert the JSON string to a Map and compare Map objects instead of comparing JSON strings.  Done in 2ffd47f.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: Done in 2ffd47f.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: Done in 2ffd47f.

sync-core/src/test/java/com/cloudant/sync/query/IndexTest.java
alfinkel: Adding a test for this is not needed presently since an Index instance will only initialize if valid parms are provided.  Currently this is limited to "tokenize".  So we can only have one entry in indexSettings.  Once we add support for more valid parms I will add more tests to ensure this scenario is tested.  I have however, added in logic to ensure that this problem is not encountered in the future.  See comment above.

alfinkel:
Adding an FB ticket to revisit the Index class and how it interacts with the Indexing code.  BugID: 46029.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -0,0 +1,185 @@
...
+                    new TypeReference<Map<String, Object>>() {
+                    });
+        } catch (IOException e) {
+            String msg = String.format("Error processing index settings %s",
+                                       this.indexSettings.toString());
+            logger.log(Level.SEVERE, msg, e);
+            return false;
+        }
+
+        return this.indexSettings.equals(settings);
```
mikerhodes: I had to double-check this is a deep-comparison. Is this worth noting, or is it well-known to experienced java programmers?

sync-core/src/test/java/com/cloudant/sync/query/IndexTest.java
mikerhodes: Accepted.

mikerhodes:
+1 on this. Build failure looks unrelated (simulator failed to boot); any further info @rhyshort?

sync-core/src/main/java/com/cloudant/sync/query/Index.java
rhyshort: Okay makes sense, if you haven't yet raise a ticket to revisit so we don't forget

rhyshort:
+1.

The build failure is nothing to worry about, emulator failed to launch for the java configuration, android passed though, so between travis and Jenkins the build passes.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
alfinkel: I'll add a comment here and then merge.

PR #110, opened 2015-03-30
44537: Replace Apache Commons HttpClient

tomblench:
The aims of this PR are as follows:

* Remove existing Apache Commons HttpClient and remove code bloat associated with a large 3rd party library
* Ameliorate issues surrounding thread/filehandle exhaustion with existing HttpRequests class on Android

Current status:
* Tests pass on Android/JSE
* Performance (test execution time) comparable with existing solution
* Some functionality lacking or has TODOs
* Need to audit that no existing functionality is lost, eg ability to set timeouts (although it was impossible for library consumers to ever access this configuration)
* CouchClient is a mess - it's been modified on a precautionary basis but it needs refactoring and unused methods should be removed
* HTTP error/exception handling has been modelled in a way to be compatible with existing code but needs to be tidied up

reviewer @mikerhodes 
reviewer @rhyshort


sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -0,0 +1,212 @@
...
+        this.requestMethod = requestMethod;
+        this.contentType = contentType;
+        this.url = url;
+        this.requestProperties = new HashMap<String, String>();
+        this.requestFilters = new LinkedList<HttpConnectionFilter>();
+        this.responseFilters = new LinkedList<HttpConnectionFilter>();
+    }
+
+    public void setInput(final String input) {
+        this.input = new InputStreamFactory() {
```
rhyshort: Couldn't this method just call `setInput(input.getBytes());`

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -0,0 +1,212 @@
...
+                connection.setRequestProperty("Expect", "100-continue");
+            }
+
+            HttpConnectionFilterContext currentContext = new HttpConnectionFilterContext(this);
+            for (HttpConnectionFilter requestFilter : requestFilters) {
+                currentContext = requestFilter.filter(currentContext);
+            }
+
+            if (input != null) {
+                // this will throw an exception if it doesn\t get a 100-continue
```
rhyshort: Typo, `doesn\t` -> `doesn't`

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -0,0 +1,212 @@
...
+                int bufSize = 1024;
+                int nRead = 0;
+                byte[] buf = new byte[bufSize];
+                InputStream is = input.getInputStream();
+                OutputStream os = connection.getOutputStream();
+
+                while((nRead = is.read(buf)) >= 0) {
+                    os.write(buf, 0, nRead);
+                }
+            }
```
rhyshort: We are already using  the `IOUtils` class in other places, I think we should use 'IOUtils.copy' here.

sync-core/src/main/java/com/cloudant/http/HttpConnectionManager.java
```diff
@@ -0,0 +1,47 @@
...
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+/**
+ * Created by tomblench on 23/03/15.
+ */
+public class HttpConnectionManager {
+
```
rhyshort: I'm not convinced we need this class. It doesn't add anything useful since we can just use `new HttpConnection` also the method names are also misleading since they don't connect to the remote unless I've missed something

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -61,29 +59,57 @@ public URI getRootUri() {
...
+
+    public void createDb() {
         InputStream is = null;
         try {
-            is = httpClient.put(this.uriHelper.getRootUri());
+            HttpConnection connection = HttpConnectionManager.connect("PUT", "application/json", this.uriHelper.getRootUri());
+            is = this.executeToInputStream(connection);
             DBOperationResponse res = jsonHelper.fromJson(new InputStreamReader(is), DBOperationResponse.class);
             if (!res.getOk()) {
                 throw new ServerException("Response from couch db server: " + res.toString());
```
rhyshort: This looks like an unchecked exception, it should be changed to be checked

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
mikerhodes: Pretty sure this is here because `IOUtils` caused every request to take several seconds to complete for reasons we couldn't quite discern.

sync-core/src/main/java/com/cloudant/http/HttpConnectionManager.java
mikerhodes: Inclined to agree this class could just be noise -- did it have a greater role to play before or have a planned greater role?

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -61,29 +59,57 @@ public URI getRootUri() {
...
+            } catch (IOException ioe2) {
+                // response can't be retrieved
+            }
         }
+        try {
+            // TODO restore this message in the last throw block
+            //String msg = String.format("Request: %s << Status: %s (%s) ", request.getRequestLine(), code, response.getStatusLine().getReasonPhrase()
+            code = connection.getConnection().getResponseCode();
+            if (code == 200 || code == 201 || code == 202) { // success (ok | created | accepted)
+                // ok
```
mikerhodes: A nice approach here is `code / 100 == 2` which covers us for any 200 code. Given we don't know whether any of the three we're checking for are strictly correct, we might as well go for any 200 code.

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -61,29 +59,57 @@ public URI getRootUri() {
...
+            } else if (code == 404) {
+                throw new NoResourceException("todo"); // TODO fix message
+            } else {
+                is = connection.getConnection().getErrorStream();
+                CouchException ex = this.jsonHelper.fromJson(new InputStreamReader(is), CouchException.class);
+                ex.setStatusCode(code);
+                throw ex;
```
mikerhodes: Close the error stream.

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
mikerhodes: In a separate PR.

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -178,19 +211,25 @@ public Response create(Object document) {
...
-            throw e;
         } finally {
             closeQuietly(is);
         }
     }
 
+    // TODO TODO TODO
+    // slim down the number of methods here, we don't need all of these variants
+    // make revid null if we don't want to specify it; bool flag for compressed/uncompressed atts
+
```
mikerhodes: Some other time; it's nice to be able to eyeball the changes here don't change things.

sync-core/src/main/java/com/cloudant/http/HttpConnectionManager.java
mikerhodes: I can see a use for helper methods here like `HttpConnectionManager.GET(URL url, String contentType)` or `HttpConnectionManager.PUT(URL url, String contentType)`.

I'd also consider shortening the class name to just `Http` because `Http.PUT(URL url, String contentType)` looks quite clear.

However, these names might be a bit misleading in the current implementation because the methods return a connection object whereas the methods look like they carry out the request.

sync-core/src/main/java/com/cloudant/http/HttpConnectionFilter.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ A <code>HttpConnectionFilter</code> can either be configured as a Request Filter or a Response Filter:
+
+ A Request Filter is run before the request is made to the server.
+
+ A Response Filter is run after the response is obtained from the
+ server but before the output stream is returned to the original client. The Response
+ Filter enables two main behaviours:
+
+ - Modifying the response for every request
```
mikerhodes: What modifications are "safe" here? Can the response filter read the response content stream for example, can/how would it modify that?

sync-core/src/main/java/com/cloudant/http/HttpConnectionFilterContext.java
```diff
@@ -0,0 +1,30 @@
+package com.cloudant.http;
+
+/**
+ * Created by tomblench on 30/03/15.
+ */
+public class HttpConnectionFilterContext {
+
+    public boolean replayRequest;
```
mikerhodes: Suggest `retryRequest` as replaying implies redoing the request verbatim (that is with whatever values the request filter gave last time).

mikerhodes:
I think this PR does two things, which are separate. Can you rewrite history to allow for PRing separately:

1. Migrate code to HttpConnection _sans_ filtering.
2. Add filtering.

Then we can discuss the API for filtering separately from the migration to not using HttpClient.

mikerhodes:
@tomblench can you update the description with what this PR is now proposing? Looks like you've just removed the filter work, but I want to make sure I'm understanding that correctly.

tomblench:
@mikerhodes updated description at the top to remove references to the filter pipeline stuff

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -0,0 +1,234 @@
...
+            connection.setDoOutput(true);
+            if (inputLength != -1) {
+                // TODO on 1.7 upwards this method takes a long, otherwise int
+                connection.setFixedLengthStreamingMode((int)this.inputLength);
+            } else {
+                // TODO some situations where we can't do chunking, like multipart/related
+                /// https://issues.apache.org/jira/browse/COUCHDB-1403
+                connection.setChunkedStreamingMode(1024);
+            }
+            connection.setRequestProperty("Expect", "100-continue");
```
mikerhodes: Be good to see a brief explanation/link to RFC for what the expected behaviour is here, and at least a "this header helps prevent us sending data when server would reject it (e.g., for bad credentials)"

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -61,29 +58,72 @@ public URI getRootUri() {
...
+    // - if there's a couch error returned as json, unmarshall and throw
+    // - anything else, just throw the IOException back, use the cause part of the exception?
 
-    public List<String> getList(URI uri) {
-        InputStream is = null;
+    // it needs to catch eg FileNotFoundException and rethrow to emulate the previous exception handling behaviour
+    private InputStream executeToInputStream(HttpConnection connection) throws CouchException {
+
+        // all couchclient requests want to receive application/json responses
+        connection.requestProperties.put("Accept", "application/json");
```
mikerhodes: I think all the `GET` etc. requests already set this where needed, or they probably should do?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -0,0 +1,234 @@
...
+            // this will throw an exception if it doesn't get a 100-continue
+            int bufSize = 1024;
+            int nRead = 0;
+            byte[] buf = new byte[bufSize];
+            InputStream is = input;
+            OutputStream os = connection.getOutputStream();
+
+            while((nRead = is.read(buf)) >= 0) {
+                os.write(buf, 0, nRead);
+            }
```
mikerhodes: I'd like to see a note here about why we're not closing the connection.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -0,0 +1,234 @@
...
+    private final String requestMethod;
+    private final String contentType;
+    public final URL url;
+
+    // created in executeInternal
+    private HttpURLConnection connection;
+
+    // optional & settable
+    private String username;
+    private String password;
```
mikerhodes: These are `private` are they ever settable? I can't see a way to set them.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -0,0 +1,234 @@
...
+        this.inputLength = input.length;
+    }
+
+    /**
+     * Set the InputStream of data to be sent to the server
+     * @param input InputStream of data to be sent to the server
+     */
+    public void setInput(InputStream input) {
+        this.input = input;
+    }
```
mikerhodes: I'd prefer these to be more HTTP-related in name, `setBody`?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -0,0 +1,234 @@
...
+        }
+    }
+
+    /**
+     * Execute request without returning data from server.
+     * @throws IOException
+     */
+    public void execute() throws IOException {
+        executeInternal();
+    }
```
mikerhodes: Is there a reason we need the `executeInternal()` method, given we could just use `execute()` in the `executeToX` methods?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * Execute request and return data from server as a String
+     * @return String of data from server, if any
+     * @throws IOException
+     */
+    public String executeToString() throws IOException {
```
mikerhodes: How about calling these `responseAsString()` etc?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
mikerhodes: Ah missed the getter/setter. Deleting these comments.

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -542,34 +614,35 @@ private String createBulkSerializedDocsPayload(List<String> serializedDocs) {
...
         URI uri = this.uriHelper.documentUri(mpw.getId(), options);
         HashMap<String, String> headers = new HashMap<String, String>();
         String contentType = "multipart/related;boundary=" + mpw.getBoundary();
-        InputStream is = this.httpClient.putStream(uri, contentType, mpw, mpw.getContentLength());
-        return jsonHelper.fromJson(new InputStreamReader(is), Response.class);
+        HttpConnection connection = Http.PUT(contentType, uri);
+        connection.setInput(mpw, mpw.getContentLength());
+        try {
+            InputStream is = connection.executeToInputStream();
+            return jsonHelper.fromJson(new InputStreamReader(is), Response.class);
```
mikerhodes: Would it be useful to have a method on HttpConnection, `responseAsClass(Response.class)`?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
tomblench: `setInput` is named as a side effect of `HttpURLConnection` having methods `setDoInput` and `getInputStream` but you could argue that's a leaky abstraction. Request and Response Body are probably easier to think about than input and output.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
tomblench: `responseAsString()` would be a lie because no I/O gets done until this call is made. It's really "Execute the request and get the response as a string". If we're happy to lie a little bit for the sake of brevity, then `responseAsString()` is acceptable.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
tomblench: Good point, I'll scrub that.

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
tomblench: Are you confusing this with the `content-type` header which is set by the first arg to `Http.POST` etc?

All the `CouchClient` methods which do something with the response body need set this header, and it doesn't do any harm for the ones which don't process the response body. So it's briefer to have this in one helper method than at all the call sites.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
mikerhodes: I'd prefer the body naming if you're fine with that.

tomblench: @mikerhodes - No strong opinions either way so I'll change it.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
mikerhodes: I find the executeToX naming quite difficult to understand. I'd be happy with `responseAsString()`, but I'd also be happy with an explicit `execute()` step, which returned something you could get the response from (even if that's `self`), i.e., `connection.execute().responseAsString()`.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
mikerhodes: Addressed, but GH hasn't  marked this outdated.

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
mikerhodes: Oh, I see, yes.

sync-core/src/main/java/com/cloudant/http/Http.java
```diff
@@ -0,0 +1,94 @@
...
+    public static HttpConnection GET(URI uri)
+    {
+        return connect("GET", null, uri);
+    }
+
+    public static HttpConnection PUT(String contentType,
+                                     URL url)
+    {
+        return connect("PUT", contentType, url);
+    }
```
mikerhodes: I think all these methods should have the URL first. I.e.:

```java
public static HttpConnection PUT(URL url, String contentType)
```

`connect` should take it second:

```java
return connect("PUT", url, contentType);
```

As should HttpConnection constructors:

```java
return new HttpConnection(requestMethod, uri.toURL(), contentType);
```

tomblench: Fixed in ccf0ef1

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
tomblench: I used your two-step suggestion and implemented it in 6c9ea32

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
tomblench: Done in 6c9ea32

mikerhodes:
What happens when one calls an execute method after one of the other ones have been called? Do we need to have some guards in to protect against this if it doesn't work?

mikerhodes:
I guess the above comment is more do we need to implement a mini-state machine to make sure operations happen in the right order (e.g., you can't set the request body after calling execute())?

tomblench:
@mikerhodes the underlying `HttpURLConnection` is very fussy about things happening in the right order and will probably throw an IOException or perhaps an unchecked exception. Ultimately this is only supposed to be a thin layer around `HttpURLConnection` which is so awkward to use, so I'm loathe to add extra checks etc.

mikerhodes:
Can you at least add a couple of tests to explore what happens? We should understand whether we're protected or not.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -0,0 +1,251 @@
...
+ * hc.setUsername("user");
+ * hc.setPassword("pass");
+ * hc.requestProperties.put("x-some-header", "some-value");
+ * hc.setRequestBody("{\"hello\": \"world\"});
+ * String result = hc.execute().responseAsString();
+ * // get the underlying HttpURLConnection if you need to do something a bit more advanced:
+ * int response = hc.getConnection().getResponseCode();
+ * </pre>
+ *
+ * @see java.net.HttpURLConnection
```
mikerhodes: Make a note here that instances of the class shouldn't be shared across threads.

mikerhodes:
+1 once threading comment is make :+1:

Probably want to end up with at least two commits, one for add the HttpConnection and Http classes, then a second for the changes to the code base. Perhaps even three, one for each of the classes too. Then the comments to the commits can describe any design decisions or quirks.

rhyshort:
+1

PR #111, opened 2015-03-31
Change AndroidTest target SDK to version 19

rhyshort:
The AndroidTest application should be targeting an version of android which runs on a phone, rather than only on wearables. This PR lowers the target SDK version to API level 19 (Android 4.4).

Reviewer @mikerhodes 
Reviewer @indisoluble


mikerhodes:
Sounds good, +1.

indisoluble:
+1

API level 20 is equal to 19 plus the wearable extension. No point in having that constraint.

PR #112, opened 2015-04-10
Add support for text indexing

alfinkel:
This PR adds the ability to use text indexing with Query.  Text indexing is necessary in order to
perform full text search queries.  This PR concentrates on index management while a separate/later PR will add full text search querying capabilities.

_Some things to look for are:_

1.  The way the metadata structure is constructed, both for a new implementation as well as a
migration from a previous release.
2.  The creation of the full text search index in SQLite (VIRTUAL TABLE).
3.  The new public IndexManager.ensureIndexed method that provides the developer with a way 
to set custom index settings.  Right now this is only used for changing the default tokenizer of the
VIRTUAL TABLE.
4.  I've temporarily added logic to ignore a text index when choosing indexes for a query.
This ensures that if a text index is created it will not compromise any current query processing.  This 
will be removed once support for `$text` and `$search` are added in a separate PR.

_New tests (of note):_

- Test the migration of metadata for current users from a previous release to this one.
- Update IndexUpdaterTest to make separate passes for JSON index testing and TEXT index testing.

reviewer @mikerhodes 
reviewer @rhyshort 

BugId: 46057

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/Index.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/Index.java
@@ -21,7 +21,6 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -71,16 +70,16 @@ class Index {
      * @param indexName the index name
      * @return the Index object or null if arguments passed in were invalid.
      */
-    public static Index setUp(List<Object> fieldNames, String indexName) {
-        return setUp(fieldNames, indexName, JSON_TYPE);
+    public static Index getInstance(List<Object> fieldNames, String indexName) {
+        return getInstance(fieldNames, indexName, JSON_TYPE);
     }
 
-    public static Index setUp(List<Object> fieldNames, String indexName, String indexType) {
-        return setUp(fieldNames, indexName, indexType, null);
+    public static Index getInstance(List<Object> fieldNames, String indexName, String indexType) {
+        return getInstance(fieldNames, indexName, indexType, null);
     }
 
     /**
-     * The setUp method handles index specific validation and ensures that the constructed
+     * This method handles index specific validation and ensures that the constructed
      * Index object is valid.
      *
      * @param fieldNames the field names in the index
@@ -90,7 +89,7 @@ class Index {
      *                      Only supported parameter is 'tokenize' for text indexes only.
      * @return the Index object or null if arguments passed in were invalid.
      */
-    public static Index setUp(List<Object> fieldNames,
+    public static Index getInstance(List<Object> fieldNames,
                               String indexName,
                               String indexType,
                               Map<String, String> indexSettings) {
@@ -175,6 +174,25 @@ class Index {
         // We perform a deep comparison of hash maps to ensure that both objects
         // and any sub-objects are equal regardless of order within the maps.
         return this.indexSettings.equals(settings);
+    }
+
+    /**
+     * Converts the index settings to a JSON string
+     *
+     * @return the JSON representation of the index settings
+     */
+    protected String settingsAsJSON() {
+        String json = null;
+        if (indexSettings != null) {
+            try {
+                json = getObjectMapper().writeValueAsString(indexSettings);
+            } catch (JsonProcessingException e) {
+                String msg = String.format("Error processing index settings %s",
+                                           this.indexSettings.toString());
+                logger.log(Level.SEVERE, msg, e);
+            }
+        }
+        return json;
     }
 
     private ObjectMapper getObjectMapper() {
--- a/sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
@@ -48,41 +48,30 @@ class IndexCreator {
         this.queue = queue;
     }
 
-    protected static String ensureIndexed(List<Object> fieldNames,
-                                          String indexName,
-                                          String indexType,
+    protected static String ensureIndexed(Index index,
                                           SQLDatabase database,
                                           Datastore datastore,
                                           ExecutorService queue) {
         IndexCreator executor = new IndexCreator(database, datastore, queue);
 
-        return executor.ensureIndexed(fieldNames, indexName, indexType);
+        return executor.ensureIndexed(index);
     }
 
     /**
-     *  Add a single, possibly compound, index for the given field names.
+     *  Add a single, possibly compound, possibly text index for the given field names.
      *
      *  This function generates a name for the new index.
      *
-     *  @param fieldNames List of field names in the sort format
-     *  @param indexName Name of index to create
-     *  @param indexType "json" is the only supported type for now
+     *  @param index The object that defines an index.  Includes field list, name, type and options.
      *  @return name of created index
      */
-    private String ensureIndexed(List<Object> fieldNames,
-                                 final String indexName,
-                                 final String indexType) {
-        if (fieldNames == null || fieldNames.isEmpty()) {
-            logger.log(Level.SEVERE, "No field names were passed to ensureIndexed");
+    @SuppressWarnings("unchecked")
+    private String ensureIndexed(final Index index) {
+        if (index == null) {
             return null;
         }
 
-        if (indexName == null || indexName.isEmpty()) {
-            logger.log(Level.SEVERE, "No index name was passed to ensureIndexed");
-            return null;
-        }
-
-        final List<String> fieldNamesList = removeDirectionsFromFields(fieldNames);
+        final List<String> fieldNamesList = removeDirectionsFromFields(index.fieldNames);
 
         for (String fieldName: fieldNamesList) {
             if (!validFieldName(fieldName)) {
@@ -95,7 +84,7 @@ class IndexCreator {
         Set<String> uniqueNames = new HashSet<String>(fieldNamesList);
         if (uniqueNames.size() != fieldNamesList.size()) {
             String msg = String.format("Cannot create index with duplicated field names %s"
-                                       , fieldNames);
+                                       , index.fieldNames);
             logger.log(Level.SEVERE, msg);
         }
 
@@ -108,25 +97,33 @@ class IndexCreator {
             fieldNamesList.add(0, "_id");
         }
 
-        // Does the index already exist; return success if it does and is same, else fail
+        // Check the index limit.  Limit is 1 for "text" indexes and unlimited for "json" indexes.
+        // Then check whether the index already exists; return success if it does and is same,
+        // else fail.
         try {
             Map<String, Object> existingIndexes = listIndexesInDatabaseQueue();
-            if (existingIndexes != null && existingIndexes.get(indexName) != null) {
-                @SuppressWarnings("unchecked")
-                Map<String, Object> index = (Map<String, Object>) existingIndexes.get(indexName);
-                String existingType = (String) index.get("type");
-                @SuppressWarnings("unchecked")
-                List<String> existingFieldsList = (List<String>) index.get("fields");
+            if (indexLimitReached(index, existingIndexes)) {
+                String msg = String.format("Index limit reached.  Cannot create index %s.",
+                                           index.indexName);
+                logger.log(Level.SEVERE, msg);
+                return null;
+            }
+            if (existingIndexes != null && existingIndexes.get(index.indexName) != null) {
+                Map<String, Object> existingIndex =
+                        (Map<String, Object>) existingIndexes.get(index.indexName);
+                String existingType = (String) existingIndex.get("type");
+                String existingSettings = (String) existingIndex.get("settings");
+                List<String> existingFieldsList = (List<String>) existingIndex.get("fields");
                 Set<String> existingFields = new HashSet<String>(existingFieldsList);
                 Set<String> newFields = new HashSet<String>(fieldNamesList);
-
-                if (existingType.equalsIgnoreCase(indexType) && existingFields.equals(newFields)) {
-                    boolean success = IndexUpdater.updateIndex(indexName,
+                if (existingFields.equals(newFields) &&
+                    index.compareIndexTypeTo(existingType, existingSettings)) {
+                    boolean success = IndexUpdater.updateIndex(index.indexName,
                                                                fieldNamesList,
                                                                database,
                                                                datastore,
                                                                queue);
-                    return success ? indexName : null;
+                    return success ? index.indexName : null;
                 }
             }
         } catch (ExecutionException e) {
@@ -146,8 +143,9 @@ class IndexCreator {
                 // Insert metadata table entries
                 for (String fieldName: fieldNamesList) {
                     ContentValues parameters = new ContentValues();
-                    parameters.put("index_name", indexName);
-                    parameters.put("index_type", indexType);
+                    parameters.put("index_name", index.indexName);
+                    parameters.put("index_type", index.indexType);
+                    parameters.put("index_settings", index.settingsAsJSON());
                     parameters.put("field_name", fieldName);
                     parameters.put("last_sequence", 0);
                     long rowId = database.insert(IndexManager.INDEX_METADATA_TABLE_NAME,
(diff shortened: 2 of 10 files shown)


sync-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -177,6 +176,25 @@ protected boolean compareIndexTypeTo(String indexType, String indexSettings) {
         return this.indexSettings.equals(settings);
     }
 
+    /**
+     * Converts the index settings to a JSON string
+     *
+     * @return the JSON representation of the index settings
+     */
+    protected String getIndexSettingsAsAString() {
```
rhyshort: I think this should be named something like `toJSONString` it makes it clearer from the method name the kind of string that is being returned

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -94,18 +93,32 @@ public IndexManager(Datastore datastore) {
         queue = Executors.newSingleThreadExecutor();
 
         final String filename = datastore.extensionDataFolder(EXTENSION_NAME) + File.separator
-                                                                        + "indexes.sqlite";
+                                                                              + "indexes.sqlite";
         SQLDatabase sqlDatabase = null;
         try {
             sqlDatabase = queue.submit(new Callable<SQLDatabase>() {
```
rhyshort: Not for this PR, but should be done while refactoring this class. It shouldn't hold on to sqlDatabase, instead it should be using an SQLDatabaseQueue to interact with the database, @alfinkel do you have plans to refactor parts of this class or do we need another ticket to handle this?

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -94,18 +93,32 @@ public IndexManager(Datastore datastore) {
...
+                        schemaIndex = new String[] {
+                                "CREATE TABLE " + INDEX_METADATA_TABLE_NAME + " ( "
+                                                + "index_name TEXT NOT NULL, "
+                                                + "index_type TEXT NOT NULL, "
+                                                + "index_settings TEXT NULL, "
+                                                + "field_name TEXT NOT NULL, "
+                                                + "last_sequence INTEGER NOT NULL);"
+                        };
+                    }
+
```
rhyshort: Can we do the migrations here in the style of the datastore, running each migration after another?

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
alfinkel: @rhyshort I had no plans of refactoring the IndexManager.  When did using SQLDatabaseQueue to interact with the database become a requirement?

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
alfinkel: @rhyshort I'm not sure what you mean, can you point me to a code sample?

alfinkel: @rhyshort nevermind, I think I found what you were referring to.  

https://github.com/cloudant/sync-android/blob/master/sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java#L112

https://github.com/cloudant/sync-android/blob/master/sync-core/src/main/java/com/cloudant/sync/datastore/DatastoreConstants.java

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
rhyshort: Discussed with @alfinkel  on Slack

alfinkel: After our Slack-chat I now remember why we did not refactor the IndexManager back when we refactored the rest of the project to use SQLDatabaseQueue.  I opened a separate FB ticket https://cloudant.fogbugz.com/f/cases/46120 to handle the refactoring for the IndexManager.

sync-core/src/main/java/com/cloudant/sync/query/Index.java
mikerhodes: `settingsAsJSON` probably works: `myIndex.settingsAsJSON()` seems to read well.

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -108,25 +97,33 @@ private String ensureIndexed(List<Object> fieldNames,
...
             Map<String, Object> existingIndexes = listIndexesInDatabaseQueue();
-            if (existingIndexes != null && existingIndexes.get(indexName) != null) {
-                @SuppressWarnings("unchecked")
-                Map<String, Object> index = (Map<String, Object>) existingIndexes.get(indexName);
-                String existingType = (String) index.get("type");
-                @SuppressWarnings("unchecked")
-                List<String> existingFieldsList = (List<String>) index.get("fields");
+            if (indexLimitReached(index, existingIndexes)) {
+                String msg = String.format("Index limit reached.  Cannot create index %s.",
+                                           index.indexName);
```
mikerhodes: Either here or in `indexLimitReached` the log message needs to indicate what "index limit reached" means. Nowhere do you say something like "You already have a text index, %s. You can only define a single text index per datastore. Delete %s and recreate to include the new fields.".

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -252,6 +268,32 @@ protected static boolean validFieldName(String fieldName) {
...
+     * @param existingIndexes the list of already existing indexes
+     * @return whether the index limit has been reached
+     */
+    @SuppressWarnings("unchecked")
+    protected static boolean indexLimitReached(Index index, Map<String, Object> existingIndexes) {
+        if (index.indexType.equalsIgnoreCase(Index.TEXT_TYPE)) {
+            for (String name : existingIndexes.keySet()) {
+                Map<String, Object> existingIndex = (Map<String, Object>) existingIndexes.get(name);
+                String type = (String) existingIndex.get("type");
+                if (type.equalsIgnoreCase("text") && !name.equalsIgnoreCase(index.indexName)) {
```
mikerhodes: Do you need to use the "text" constant?

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -264,16 +306,16 @@ protected static boolean validFieldName(String fieldName) {
...
         Joiner joiner = Joiner.on(" NONE,").skipNulls();
         String cols = joiner.join(columns);
 
         return String.format("CREATE TABLE %s ( %s NONE )", tableName, cols);
     }
 
-    private String createIndexIndexStatementForIndexName(String indexName, List<String> columns) {
-        String tableName = IndexManager.tableNameForIndex(indexName);
+    private String createIndexIndexStatementForIndex(Index index, List<String> columns) {
+        String tableName = IndexManager.tableNameForIndex(index.indexName);
```
mikerhodes: You could pass in the indexName here, avoid the change entirely.

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -94,18 +93,32 @@ public IndexManager(Datastore datastore) {
...
+                    } else {
+                        schemaIndex = new String[] {
+                                "CREATE TABLE " + INDEX_METADATA_TABLE_NAME + " ( "
+                                                + "index_name TEXT NOT NULL, "
+                                                + "index_type TEXT NOT NULL, "
+                                                + "index_settings TEXT NULL, "
+                                                + "field_name TEXT NOT NULL, "
+                                                + "last_sequence INTEGER NOT NULL);"
+                        };
+                    }
```
mikerhodes: I'd much prefer here to see a stepped progression of schema as we have in other places so the differences between each are clear. Then going to version 2 is a simple addition, whereas when we upgrade to schema version 2 now a bigger refactor will be needed which will again be harder to eyeball as correct.

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -226,35 +244,46 @@ public String ensureIndexed(List<Object> fieldNames) {
      *  @return name of created index
      */
     public String ensureIndexed(List<Object> fieldNames, String indexName) {
-        return ensureIndexed(fieldNames, indexName, "json");
+        return IndexCreator.ensureIndexed(Index.setUp(fieldNames, indexName),
```
mikerhodes: `setUp` creates an Index object rather than setting up the index itself. Is this class method -> instance of object pattern typically given the name `setUp`? E.g., in objc there's the convention of `[Index indexNamed:indexName forFields:fieldNames]` -- is there a convention like `Index.indexFor(fieldNames, indexName)`?

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -226,35 +244,46 @@ public String ensureIndexed(List<Object> fieldNames) {
...
+                                          database,
+                                          datastore,
+                                          queue);
     }
 
     /**
-     *  Add a single, possibly compound, index for the given field names.
+     *  Add a single, possibly compound, possibly text index for the given field names.
```
mikerhodes: Not sure "possibly text" is useful here -- why not "possibly JSON" too?

sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
```diff
@@ -56,7 +56,7 @@ public IndexUpdater(SQLDatabase database, Datastore datastore, ExecutorService q
     /**
      *  Update all indexes in a set.
      *
-     *  This indexes are assumed to already exist.
+     *  These indexes are assumed to already exist.
```
mikerhodes: Typo fix unrelated to this commit.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -327,6 +327,13 @@ protected static String chooseIndexForFields(Set<String> neededFields,
         String chosenIndex = null;
         for (String indexName: indexes.keySet()) {
             Map<String, Object> indexDefinition = (Map<String, Object>) indexes.get(indexName);
+
+            // TODO - Remove once the query component of full text search is added.
+            String indexType = (String) indexDefinition.get("type");
+            if (indexType.equalsIgnoreCase("text")) {
+                continue;
+            }
```
mikerhodes: I'd add an always-printed log message here too to make sure it's remembered.

sync-core/src/test/java/com/cloudant/sync/query/IndexManagerTest.java
```diff
@@ -30,6 +36,67 @@
 public class IndexManagerTest extends AbstractIndexTestBase {
 
     @Test
+    public void correctlyMigratesToNewMetadataSchema() {
```
mikerhodes: This method would be much simpler if you created a "migrate" class method on IndexManager which took a database as an argument and performed the migration. Then you could just create a test SQLite database, set it up as in v1, then run the migration on it. This would allow you to run this test without needing to set up all the scaffolding like a datastore.

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -281,4 +323,17 @@ private String createIndexIndexStatementForIndexName(String indexName, List<Stri
...
+                                                       List<String> columns,
+                                                       List<String> indexSettings) {
+        String tableName = IndexManager.tableNameForIndex(index.indexName);
+        Joiner joiner = Joiner.on(",").skipNulls();
+        String cols = joiner.join(columns);
+        String settings = joiner.join(indexSettings);
+
+        return String.format("CREATE VIRTUAL TABLE %s USING FTS4 ( %s, %s )", tableName,
+                                                                              cols,
+                                                                              settings);
```
mikerhodes: What if we have a fieldName containing an equals, in the pathological case, say "tokenizer=simple"?

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
mikerhodes: Also, what if the name is the same but the list of fields is different?

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -158,23 +156,41 @@ public Boolean call() {
...
+                    columnList.add("\"" + field + "\"");
+                }
+
+                List<String> statements = new ArrayList<String>();
+                if (index.indexType.equalsIgnoreCase(Index.TEXT_TYPE)) {
+                    List<String> settingsList = new ArrayList<String>();
+                    // Ensure that the _id and _rev columns are not included
+                    // in the text index.
+                    settingsList.add("notindexed=\"_id\"");
+                    settingsList.add("notindexed=\"_rev\"");
```
mikerhodes: So this means _id and _rev are stored, but not indexed?

alfinkel: This means that when doing a full text search on a term, that term will not be applied to the `_id` or the `_rev` columns.

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
alfinkel: All we're really interested here is to check whether another *text* index already exists.  A later check for name and field equality is performed and in the scenario where the name is the same but the fields are different, this check fails, which drops us into the table creation logic which will in turn also fail since a table does already exist albeit with different data structures.  This logic is untouched since the port of your original implementation.

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
alfinkel: Short answer:  The statement will fail with a `Error: unrecognized parameter: "tokenize=simple"`.

Longer answer:  Apparently, the `=` sign cannot be included in the column name.  It seems to trigger SQLite to think that you are passing in a parameter and value.  But since we are surrounding all of our columns with double quotes we never run into a situation where the parameter will be valid.  Like in the case of `"tokenize=simple"`, SQLite reads the parameter as `"tokenize` and the  value as `simple"`.  Hence the error message.  So really the only negative here is that we cannot honor `=` in our text index field names.

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
alfinkel: Done in c716a18a71a0afe6dd681cfd3843746b60847d80

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
alfinkel: Done in c716a18a71a0afe6dd681cfd3843746b60847d80

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
alfinkel: Now using `TEXT_TYPE` constant in c716a18a71a0afe6dd681cfd3843746b60847d80

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
alfinkel: Related to comment https://github.com/cloudant/sync-android/pull/112#discussion_r28145006.  Done in  c716a18a71a0afe6dd681cfd3843746b60847d80

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
alfinkel: Changed `setUp` to `getInstance` in c716a18a71a0afe6dd681cfd3843746b60847d80

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
alfinkel: Removed in c716a18a71a0afe6dd681cfd3843746b60847d80

sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
alfinkel: I will move this out into a separate commit before merging.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: Done in c716a18a71a0afe6dd681cfd3843746b60847d80

sync-core/src/test/java/com/cloudant/sync/query/IndexManagerTest.java
alfinkel: With the updated incremental manner in which the metadata table is being created (and altered) there is no longer a need for this test.  Removed in c716a18a71a0afe6dd681cfd3843746b60847d80.

rhyshort:
I'm happy for this to merge. Don't worry about the jenkins failure, its being worked on.

+1

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
mikerhodes: Do we check the index's settings during the exsting code? (i.e., to stop us creating an index with the same name, same fields but different settings)?

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
mikerhodes: We should note this as a comment in this code, and also write something in the documentation for the feature.

mikerhodes:
I'd change the subject of "Add text index management support" to something which is a bit clear of itself on query not yet being supported. "Add full text indexing (no querying yet)" maybe?

Otherwise +1.

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
alfinkel: When you say _existing code_ I am assuming that you mean the code in this PR since prior to this PR index settings did not exist.  In the event that an index already exists with the same name, same fields but different settings, the logic would drop us into the index creation process which would fail via SQLite (desired result) since a table already exists with the same name.  This is pretty much the same process as described in https://github.com/cloudant/sync-android/pull/112#discussion_r28170373

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
alfinkel: Added a javadoc comment to createVirtualTableStatementForIndex method including a note about using '=' in a field name.  Added FB46163 and referenced these comments to ensure we make a note of this in the documentation

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
mikerhodes: It'd be nice if we validated prior to hitting SQL (in case we work out an amazing backend someday) but I think as long as there's a test for this situation, we can catch it when it happens.

PR #113, opened 2015-04-14
Correct creation of replications

mikerhodes:
This updates docs to put username and password in the URL.

It also changes the creation of replicator objects to use the `oneway(source, target)` shorthand.

reviewer @gadamc 
reviewer @tomblench

### DIFF
--- a/doc/replication.md
+++ b/doc/replication.md
@@ -174,7 +174,7 @@ if (replicator.getState() != Replicator.State.COMPLETE) {
 }
 ```
 
-And running a full sync, that is, two one way replicaitons:
+And running a full sync, that is, two one way replication:
 
 ```java
 
--- a/doc/replication.md
+++ b/doc/replication.md
@@ -128,7 +128,10 @@ Datastore ds = manager.openDatastore("my_datastore");
 
 // Create a replicator that replicates changes from the local
 // datastore to the remote database.
-Replicator replicator = ReplicatorFactory.oneway(ds, uri);
+PushReplication push = new PushReplication();
+push.source = ds;
+push.target = uri;
+Replicator replicator = ReplicatorFactory.oneway(push);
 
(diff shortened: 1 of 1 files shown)


gadamc:
So, `oneway(source, target)` is the deprecated method, right? You want to document using that instead of the `Push/PullReplication` objects ?

tomblench:
@gadamc is right about the deprecation. It needs to look like this:

```java
PushReplication push = new PushReplication();
push.source = ds;
push.target = uri;
Replicator replicator = ReplicatorFactory.oneway(push);
```

mikerhodes:
Bum, will change back, this is the exact reason why I PRed this change :)

mikerhodes:
@gadamc @tomblench please check the update.

tomblench:
+1 :+1:

gadamc:
I think we should fully remove the ReplicatorFactory.oneway(ds, uri) methods. They are essentially unusable given that the credentials are stripped from the URI and there's no way, that I can see, to easily provide authentication (though, somebody may be able to modify the HTTP headers???) 

https://github.com/cloudant/sync-android/blob/master/sync-core/src/main/java/com/cloudant/sync/replication/ReplicatorFactory.java#L75

https://github.com/cloudant/sync-android/blob/master/sync-core/src/main/java/com/cloudant/sync/replication/ReplicatorFactory.java#L108

https://github.com/cloudant/sync-android/blob/master/sync-core/src/main/java/com/cloudant/sync/replication/ReplicatorFactory.java#L117-L121

I guess this should be done in a separate PR...

mikerhodes:
@gadamc yes, it does look pretty useless and yes, it needs a separate bug filed for it. If you're happy with this PR please leave a +1. We can file a new case to either not remove creds or remove the method.

gadamc:
+1

PR #115, opened 2015-04-20
Add full text query support

alfinkel:
_What_

This PR adds support for the `$text` and `$search` operators in Query.

_Why_

The change is to provide full text search capabilities in Query.

_How_

The normalization/validation process is updated to handle the new operators and the SQL translation process is updated to ensure correct generation of SQL and node placement in the query node tree. 

_Notes_

This PR does not include the unit tests.  Unit tests will be included in a following PR.  Logic has been temporarily added to deactivate the full text search logic until the unit tests are delivered.

reviewer @mikerhodes 
reviewer @rhyshort 

BugId: 45395

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
@@ -104,10 +104,23 @@ class QuerySqlTranslator {
         TranslatorState state = new TranslatorState();
         QueryNode node = translateQuery(query, indexes, state);
 
-        // If we haven't used a single index or an OR clause is missing an index,
-        // we need to return a query which returns every document, so the posthoc
-        // matcher can run over every document to manually carry out the query.
-        if (!state.atLeastOneIndexUsed || state.atLeastOneORIndexMissing) {
+        if (state.textIndexMissing) {
+            String msg = "No text index defined, cannot execute query containing a text search.";
+            logger.log(Level.SEVERE, msg);
+            return null;
+        } else if (state.textIndexRequired && state.atLeastOneIndexMissing) {
+            String msg = String.format("Query %s contains a text search but is missing \"json\"" +
+                                       " index(es).  All indexes must exist in order to execute a" +
+                                       " query containing a text search.  Create all necessary" +
+                                       " indexes for the query and re-execute.",
+                                       query.toString());
+            logger.log(Level.SEVERE, msg);
+            return null;
+        } else if (!state.textIndexRequired &&
+                      (!state.atLeastOneIndexUsed || state.atLeastOneORIndexMissing)) {
+            // If we haven't used a single index or an OR clause is missing an index,
+            // we need to return a query which returns every document, so the posthoc
+            // matcher can run over every document to manually carry out the query.
             Set<String> neededFields = new HashSet<String>(Arrays.asList("_id"));
             String allDocsIndex = chooseIndexForFields(neededFields, indexes);
 
@@ -156,13 +169,11 @@ class QuerySqlTranslator {
             root = new OrQueryNode();
         }
 
-        //
-        // First handle the simple "field": { "$operator": "value" } clauses. These are
-        // handled differently for AND and OR parents, so we need to have the conditional
-        // logic below.
-        //
-
+        // Compile a list of simple clauses to be handled below.  If a text clause is
+        // encountered, store it separately from the simple clauses since it will be
+        // handled later on its own.
         List<Object> basicClauses = null;
+        Object textClause = null;
 
         for (Object rawClause: clauses) {
             Map<String, Object> clause = (Map<String, Object>) rawClause;
@@ -172,9 +183,14 @@ class QuerySqlTranslator {
                     basicClauses = new ArrayList<Object>();
                 }
                 basicClauses.add(rawClause);
+            } else if (field.equalsIgnoreCase(TEXT)) {
+                textClause = rawClause;
             }
         }
 
+        // Handle the simple "field": { "$operator": "value" } clauses. These are
+        // handled differently for AND and OR parents, so we need to have the conditional
+        // logic below.
         if (basicClauses != null) {
             if (query.get(AND) != null) {
                 // For an AND query, we require a single compound index and we generate a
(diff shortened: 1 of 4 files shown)


sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -104,10 +104,23 @@ public static QueryNode translateQuery(Map<String, Object> query,
...
+            String msg = "No text index defined, cannot execute query containing a text search.";
+            logger.log(Level.SEVERE, msg);
+            return null;
+        } else if (state.textIndexRequired && state.atLeastOneIndexMissing) {
+            String msg = String.format("Query %s contains a text search but is missing \"json\"" +
+                                       " index(es).  All indexes must exist in order to execute a" +
+                                       " query containing a text search.  Create all necessary" +
+                                       " indexes for the query and re-execute.",
+                                       query.toString());
+            logger.log(Level.SEVERE, msg);
```
mikerhodes: Why is it necessary for all indexes to exist if the query contains a text predicate?

Is it an artifact of this implementation, or is it something fundamental? I'm not saying we should change this now, but more trying to understand whether this is a limitation imposed by our current implementation or whether we've just chosen to go this way for the time being?

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -250,6 +266,31 @@ private static QueryNode translateQuery(Map<String, Object> query,
             }
         }
 
+        // A text clause by nature uses its own text index it is handled
+        // separately from other simple clauses.
```
mikerhodes: Can you add an example of what `textClause` looks like to make it easier to follow the below.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -328,11 +369,9 @@ protected static String chooseIndexForFields(Set<String> neededFields,
         for (String indexName: indexes.keySet()) {
             Map<String, Object> indexDefinition = (Map<String, Object>) indexes.get(indexName);
 
-            // TODO - Remove once the query component of full text search is added.
+            // Don't choose a text index for a non-text query clause
             String indexType = (String) indexDefinition.get("type");
             if (indexType.equalsIgnoreCase("text")) {
```
mikerhodes: constant

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -372,6 +425,31 @@ protected static SqlParts selectStatementForAndClause(List<Object> clause,
...
+        }
+
+        Map<String, Object> textClause = (Map<String, Object>) clause;
+        Map<String, String> searchClause = (Map<String, String>) textClause.get(TEXT);
+
+        String tableName = IndexManager.tableNameForIndex(indexName);
+        String search = searchClause.get(SEARCH);
+
+        String sql = String.format("SELECT _id FROM %s WHERE %s MATCH ?", tableName, tableName);
+        return SqlParts.partsForSql(sql, new String[]{ String.format("'%s'", search) });
```
mikerhodes: Is this placeholder correct, looks like if my search includes a `'` I can SQL inject here?

sync-core/src/main/java/com/cloudant/sync/query/QueryValidator.java
```diff
@@ -437,6 +459,48 @@ private static boolean validateClause(Map<String, Object> clause) {
...
+                    }
+                } else {
+                    String msg = String.format("Invalid operator %s in text search", operator);
+                    logger.log(Level.SEVERE, msg);
+                }
+            }
+        } else {
+            String msg = String.format("Text search expects a Map, found %s instead.", clause);
+            logger.log(Level.SEVERE, msg);
+        }
```
mikerhodes: Can you structure this as a series of tests rather than a nested if/else clause? It's easy to miss that `return validatePredicateValue(...)` in the middle.

```java
if (!(clause instanceof Map)) { 
    String msg = String.format("Text search expects a Map, found %s instead.", clause);
    logger.log(Level.SEVERE, msg);
    return false;
}

/* etc. */
```

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: Before FTS we would pass any query that had at least one missing index to the post hoc matcher where we would then take the full list of doc ids in a datastore and inspect each document and find query matches programmatically.  I think early on we agreed to limit this FTS initial implementation to the SQL engine only since implementing a post hoc matcher component would be non-trivial.  That being the case we must for now "turn off" matcher functionality for any query that contains a FTS predicate and mandate that it be handled by the SQL engine only.  So I guess to answer your question, is that it does not appear to be anything fundamental that dictates this, but instead was a design choice made for this initial implementation.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: Added example in cdbf17d0a54e34272cbf77e02cdc4dc10c8b7921

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: Putting the single quotes in the sql as in `...MATCH '?'` causes an issue.  Escaping them in the sql String does not help.  The single quotes are necessary to tell SQLite to perform the FTS via the MATCH statement.  Instead I am now escaping all single quotes in the passed in `search` as two single quotes.  This resolves the SQL injection issue and allows for searching on text containing single quotes and apostrophes.  I added a test for this which will be included in the follow up PR.  Fixed in cdbf17d0a54e34272cbf77e02cdc4dc10c8b7921.

sync-core/src/main/java/com/cloudant/sync/query/QueryValidator.java
alfinkel: FIxed in cdbf17d0a54e34272cbf77e02cdc4dc10c8b7921

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
mikerhodes: Thanks, yup this sounds correct and safest.

mikerhodes:
I think this looks correct now -- though obviously I'd like to review your tests before we enable this code. +1.

sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
alfinkel: Added FB case 46482 - https://cloudant.fogbugz.com/f/cases/46482 to revisit this restriction in the future.

rhyshort:
+1

PR #116, opened 2015-04-22
Upgrade android gradle plug-in to 1.1.3

rhyshort:
BugzID 45999

reviewer: @tomblench 
reviewer: @mikerhodes

### DIFF
--- a/AndroidTest/build.gradle
+++ b/AndroidTest/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.0.0'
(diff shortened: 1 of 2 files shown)


tomblench:
+1 who can argue with upgrading things?

mikerhodes:
+1, though I'd like to understand whether this upgrade fixes any of the problems we have (e.g., does it make unit testing with junit 4 specifically easier, rather than just improve things in general for testing?)

rhyshort:
From what i've read of the change log, it doesn't help us yet, its mostly general improvements, but it does add unit testing support, but it doesn't run them on the emulator, its uses Java SE for unit tests, see http://tools.android.com/tech-docs/new-build-system for all the changes

PR #122, opened 2015-04-29
Fix NPE during Compaction

rhyshort:
When Compacting attachments, an NullPointerException is thrown if there are not
any attachments to compact.

BugzID 46614

reviewer @tomblench 
reviewer @alfinkel


tomblench:
+1 looks good to me

alfinkel:
+1

PR #125, opened 2015-05-08
Add fts availability check to API

alfinkel:
_What_

Add a public method to the IndexManager that allows users of the library to check whether full text search is enabled in SQLite.

_Why_

Since Cloudant Query - mobile text search leverages SQLite FTS functionality and FTS is a compile option for SQLite, it is conceivable that FTS may not be enabled in SQLite.  This method provides a check regarding SQLite FTS availability.

_How_

- Add a static method to the IndexManager to check the SQLite PRAGMA compile options for ENABLE_FTS3 and ENABLE_FTS3_PARENTHESIS.
- Add a public method to the IndexManager to return text search status.
- Add logic to the index creator to halt any text index creation if FTS is not available in SQLite.

reviewer @mikerhodes 
reviewer @rhyshort

BugId: 46093

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
@@ -58,7 +58,8 @@ class IndexCreator {
     }
 
     /**
-     *  Add a single, possibly compound, possibly text index for the given field names.
+     *  Add a single, possibly compound index for the given field names and ensure all indexing
+     *  constraints are met.
      *
      *  This function generates a name for the new index.
      *
@@ -69,6 +70,14 @@ class IndexCreator {
     private String ensureIndexed(final Index index) {
         if (index == null) {
             return null;
+        }
+
+        if (index.indexType.equalsIgnoreCase("text")) {
+            if (!IndexManager.ftsAvailable(queue, database)) {
+                logger.log(Level.SEVERE, "Text search not supported.  To add support for text " +
+                                         "search, enable FTS compile options in SQLite.");
+                return null;
+            }
         }
 
         final List<String> fieldNamesList = removeDirectionsFromFields(index.fieldNames);
--- a/sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
@@ -45,6 +45,7 @@ import com.cloudant.sync.util.DatabaseUtils;
 import java.io.File;
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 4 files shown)


sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -376,4 +380,60 @@ protected SQLDatabase getDatabase() {
...
+                                    ftsOptionCount = ftsOptionCount + 1;
+                                }
+                            }
+                        } catch (SQLException e) {
+                            String msg = "Failed to get a list of compile options from SQLite.";
+                            logger.log(Level.SEVERE, msg, e);
+                        } finally {
+                            DatabaseUtils.closeCursorQuietly(compileOptions);
+                        }
+                        return ftsOptionCount >= 2;
```
mikerhodes: Perhaps use ftsComplileOptions's length here rather than hard-coding 2?

mikerhodes: You could also use something like:

```java
String compileOption = compileOptions.getString(0);
ftsCompileOptions.remove(complieOption);
// ...
return ftsCompileOptions.isEmpty();
```
Which more explicitly checks that we saw all the options we wanted, avoiding some odd query result where `ENABLE_FTS3` is included twice in the results for some reason.

sync-core/src/test/java/com/cloudant/sync/query/IndexManagerTest.java
```diff
@@ -142,4 +142,9 @@ public void deleteATextIndex() throws Exception {
         assertThat(im.listIndexes().isEmpty(), is(true));
     }
 
+    @Test
+    public void validateTextSearchIsAvailable() {
+        assertThat(im.isTextSearchEnabled(), is(true));
+    }
```
mikerhodes: Do you want to check the compile options before asserting this?

sync-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -71,6 +72,14 @@ private String ensureIndexed(final Index index) {
             return null;
         }
 
+        if (index.indexType.equalsIgnoreCase("text")) {
+            if (!IndexManager.ftsCompileOptionsAvailable(queue, database)) {
```
mikerhodes: Just call it `ftsAvailable` rather than exposing the implementation in the method name.

alfinkel: Fixed in 17f9845

sync-core/src/test/java/com/cloudant/sync/query/IndexManagerTest.java
alfinkel: Fixed in 6d85a3d

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
alfinkel: Fixed in cc5f890

rhyshort:
+1

mikerhodes:
:+1:

PR #127, opened 2015-05-13
Encode + characters in query strings

tomblench:
FB ticket 47061:

* *What:* Correctly percent-encode plus characters (`+`) in the query part of the URL
* *Why:* Lucene queries etc need to use this character. If it isn't encoded then the `+` character is interpreted as a space by CouchDB
* *How:* Replace `+` with `%2B` in `encodeQueryParameter`
* *Testing:* Added unit test `buildDocumentUri_options_hasPlus` and tested end-to-end with `curl` (see related ticket 46986)

reviewer @mikerhodes 
reviewer @brynh

### DIFF
--- a/sync-core/src/main/java/com/cloudant/mazha/CouchURIHelper.java
+++ b/sync-core/src/main/java/com/cloudant/mazha/CouchURIHelper.java
@@ -236,7 +236,9 @@ public class CouchURIHelper {
                     null // fragment
             );
             return uri.toASCIIString()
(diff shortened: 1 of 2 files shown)


mikerhodes:
I assume you tested this against the server?

tomblench:
@mikerhodes tested end to end as discussed in case 46986 - it's pretty tricky to set up an end to end test for this

brynh:
+1

mikerhodes:
Do any docs need updating along with this change?

mikerhodes:
I'm guessing not, +1.

PR #128, opened 2015-05-15
Remove Thread Local code from AndroidSQLite

rhyshort:
Removed thread local code from AndroidSQLite, it is no longer
needed to match the behaviour of the JavaSE implementation.

BugzID: 47161

reviewer @tomblench 
reviewer @brynh

PR #129, opened 2015-05-15
Do not close SQLDatabaseQueue's database directly

rhyshort:
Use SQLDatabaseQueue's shutdown method to close the database,
without explicitly closing the underlying database from the datastore.

BugzId: 47163

reviewer @tomblench 
reviewer @brynh


tomblench:
@rhyshort I wonder if you can take the opportunity to do another refactor here and remove the dbOpen flag. You can use queue.isShutdown() instead because If the queue is open and ready for business then the database is implicitly open.

tomblench:
↑ or you could do this in #128, they both are related to this.

tomblench:
+1 :+1:

brynh:
+1

PR #130, opened 2015-05-15
Make SQLDatabaseQueue less susceptible to Race Conditions

rhyshort:
What:

Reduce the possible window for a task to be added to the SQLiteDatabaseQueue after shutdown has been called.

How:

Introduce a flag which enables SQLDatabase Queue to reject tasks, as soon as shutdown is called.
When the the queue can no longer accept jobs, RejectedExecutionException is thrown. During the shutdown method, will wait 2 seconds to allow other jobs to finish.

Things to look at:

I'm not sure of waiting two seconds is appropriate, but having the ExecutorService thread killed before the job is complete is a bit of a risk, any thoughts on this? 

BugzId: 47164

reviewer @tomblench 
reviewer @mikerhodes

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
+++ b/sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
@@ -21,6 +21,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -34,6 +36,7 @@ public class SQLDatabaseQueue {
     private final SQLDatabase db;
     private final ExecutorService queue = Executors.newSingleThreadExecutor();
     private final Logger logger = Logger.getLogger(SQLDatabase.class.getCanonicalName());
+    private volatile boolean acceptTasks = true;
 
     /**
      * Creates an SQLQueue for the database specified.
@@ -42,6 +45,8 @@ public class SQLDatabaseQueue {
      */
     public SQLDatabaseQueue(String filename) throws IOException {
         this.db = SQLDatabaseFactory.createSQLDatabase(filename);
+        //can directly add to queue no need to check
+        //if it is still running
         queue.submit(new Runnable() {
             @Override
             public void run() {
@@ -56,14 +61,13 @@ public class SQLDatabaseQueue {
      * @param version The version of the schema
      */
     public void updateSchema(final String[] schema, final int version){
-        queue.submit(new Callable<Object>() {
+        this.submit(new SQLQueueCallable<Object>() {
             @Override
-            public Object call() throws Exception {
-                SQLDatabaseFactory.updateSchema(db,schema,version);
+            public Object call(SQLDatabase db) throws Exception {
+                SQLDatabaseFactory.updateSchema(db, schema, version);
                 return null;
             }
         }); //fire and forget
-
     }
 
     /**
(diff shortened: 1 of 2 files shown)


sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
```diff
@@ -128,6 +135,11 @@ public void run() {
             }
         });
         queue.shutdown();
+        try {
+            queue.awaitTermination(200, TimeUnit.SECONDS);
```
tomblench: This wait is for 200 seconds but your PR comments say 2 seconds.

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
```diff
@@ -42,6 +45,8 @@
      */
     public SQLDatabaseQueue(String filename) throws IOException {
         this.db = SQLDatabaseFactory.createSQLDatabase(filename);
+        //can directly add to queue no need to check
```
tomblench: No need to check what and why can we directly add?

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
mikerhodes: Why bother waiting here? I'm not sure what it add to it.

Really I think you want something which calls `shutdownNow()` once `db.close()` has been called. Can you call it from the task itself?

rhyshort: I guess we could call `shutdownNow()` in the task running on the queue,the tests pass with that behavior, at least on JavaSE.

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
rhyshort: the next line finishes the comment

mikerhodes: But still doesn't explain why there's no need to check, and it is ambiguous what the `it` refers to in the following line, the `queue` or the directly added thing (presumably the runnable?).

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
```diff
@@ -121,6 +126,8 @@ public Integer call() throws Exception {
      * tasks
      */
     public void shutdown() {
+        acceptTasks = false;
+        //pass straight to queue, tasks passed via submitTaskToQueue will now be blocked.
```
mikerhodes: Not blocked but rejected

mikerhodes:
I think the approach of using the flag to reject new tasks is sound.

What I don't understand is what behaviour you are trying to achieve in `shutdown`? Is it meant to be blocking? Are you aiming that the task shutdown puts on the queue is the final executed task?

rhyshort:
The main aim is shutdown puts the final task on the queue, which is the one which closes the database. Meaning no other task should hit the queue when it is put on. My idea behind waiting for shutdown to finish, is to make the DatabaseClosed event to be fired when the DB is actually closed, not that we've shutdown the DB queue. Does that make sense?

rhyshort:
Actually I realised the way I was doing it was silly. Instead of waiting on the queue to terminate, calling `get` on the future returned from submit, would have a similar effect. But it will be more explicit what is going on

mikerhodes:
Unless you make `acceptTasks` volatile, you don't get a guarantee that your task is last on the queue. However, I'd suggest avoiding doing that because the likelihood of problems arising from this is very small and the performance cost would be high. Therefore you don't offer a solid guarantee.

Await termination was probably the better way to do it, but with a timeout of something like five minutes rather than a very short one.

I _think_ this is right?

mikerhodes:
:+1:

tomblench:
+1

PR #131, opened 2015-05-15
47171 datastore manager test fixes

ricellis:
*What*
Fix DatastoreManagerTest

*Why*
Some DatastoreManagerTest  test cases do not pass on Windows.

*How*
Ensure that Datastores opened by the test cases "list5Datastores" and "listDatastoresWithSlashes" are closed at the end of the tests before the tear down deletion.
Set a path location as read only for the DatastoreManager non-writeable location test instead of assuming that the root location is non writeable.

*Testing*
This is an existing test class.

reviewer @rhyshort 
reviewer @tomblench

### DIFF
--- a/sync-core/src/test/java/com/cloudant/sync/datastore/DatastoreManagerTest.java
+++ b/sync-core/src/test/java/com/cloudant/sync/datastore/DatastoreManagerTest.java
@@ -1,11 +1,11 @@
 /**
  * Copyright (c) 2013 Cloudant, Inc. All rights reserved.
- *
+ * <p/>
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
- *
+ * <p/>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p/>
  * Unless required by applicable law or agreed to in writing, software distributed under the
  * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
  * either express or implied. See the License for the specific language governing permissions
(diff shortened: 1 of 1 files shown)


sync-core/src/test/java/com/cloudant/sync/datastore/DatastoreManagerTest.java
```diff
@@ -142,28 +149,38 @@ public void createDatastoreWithForwardSlashChar() throws Exception {
...
-        List<String> datastores = manager.listAllDatastores();
+        Datastore ds = manager.openDatastore("datastore/mynewdatastore");
+        try {
+            List<String> datastores = manager.listAllDatastores();
 
-        Assert.assertEquals(1, datastores.size());
-        Assert.assertEquals("datastore/mynewdatastore", datastores.get(0));
+            Assert.assertEquals(1, datastores.size());
+            Assert.assertEquals("datastore/mynewdatastore", datastores.get(0));
+        }finally{
```
mikerhodes: Spaces.

ricellis: Applied the code formatter in 14d5a1d

sync-core/src/test/java/com/cloudant/sync/datastore/DatastoreManagerTest.java
```diff
@@ -142,28 +151,38 @@ public void createDatastoreWithForwardSlashChar() throws Exception {
 
     @Test
     public void list5Datastores() throws Exception {
+        List<Datastore> opened = new ArrayList<Datastore>();
 
-        for (int i = 0; i < 5; i++) {
-            manager.openDatastore("datastore" + i);
-        }
+        try {
```
tomblench: Could this be done more concisely (without the `opened` list) by putting the try inside the loop?

ricellis: I don't see how we can nest the try inside the loop because the test needs all 5 datastores to be open before it does its assertions. We also need to ensure we close any already created datastores if we subsequently have a creation failure, so we need to surround the openDatastore with the try anyway. Did you have something specific in mind that I'm missing?

rhyshort:
looks good to me +1

tomblench:
OK +1

PR #132, opened 2015-05-15
47172 close attachment streams

ricellis:
*What*
Close attachment streams after calculating sha1 hash

*Why*
If the file streams are not closed it causes AttachmentNotSavedExceptions on Windows when deletion of the temporary copy is attempted.

*How*
Add finally blocks to close the streams.

*Testing*
Exercised by existing attachment tests.

reviewer @rhyshort 
reviewer @tomblench

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
+++ b/sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
@@ -17,6 +17,7 @@ package com.cloudant.sync.datastore;
 import com.cloudant.sync.util.Misc;
 
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
 
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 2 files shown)


sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -1,11 +1,11 @@
 /**
  * Copyright (c) 2014 Cloudant, Inc. All rights reserved.
- *
+ * <p/>
```
tomblench: Do we need these para breaks? If we are concerned about how these comments like in javadoc they should probably all be wrapped in a <pre> block or something. The copyright headers probably all need revisiting soon to fix IBM/Cloudant attributions and other small variations.

sync-core/src/test/java/com/cloudant/sync/datastore/AttachmentTest.java
```diff
@@ -42,27 +43,34 @@ public void setAndGetAttachmentsTest() throws Exception {
...
 
-            SavedAttachment savedAtt2 = (SavedAttachment) datastore.attachmentsForRevision(newRevision).get(0);
+            SavedAttachment savedAtt2 = (SavedAttachment) datastore.attachmentsForRevision
+                    (newRevision).get(0);
             Assert.assertArrayEquals(expectedSha1, savedAtt2.key);
         } catch (FileNotFoundException fnfe) {
-            Assert.fail("FileNotFoundException thrown "+fnfe);
+            Assert.fail("FileNotFoundException thrown " + fnfe);
+        } finally {
+            if (fis != null) {
```
tomblench: Can we use `closeQuietly` here like you did before, and then you don't need to do the null check?

tomblench:
Apart from my comments I'm +1 on this, it needs to be done even though java makes the code so much uglier. There's a lot of whitespace changes in this commit which makes it difficult to see what is going on - if you really wanted to apply those changes perhaps they should be in a separate commit?

rhyshort:
+1 once tom's comments are resolved

ricellis:
In 4b6080d I've fixed up the formatting so that it only applied to the changes instead of also formatting the javadoc.
In 4d56f78 I've switched the test class to also use IOUtils.closeQuietly

PR #133, opened 2015-05-15
Fix failing validator test due to clause order

alfinkel:
_What_

When a test runs under certain versions of Java, the test fails due to the order of the actual normalized query object and the expected query object not matching due to query clause ordering.

_Why_

In reality the query objects should be considered as a match regardless of clause order and the test should succeed.

_How_

Change the QueryValidatorTest.normalizesMultiFieldQueryWithTextSearch test to instantiate the query object as a LinkedHashMap<String, Object> rather than a HashMap<String, Object>.  This will impose an order and will allow the comparison to the expected query which also is a LinkedHashMap to be successful.

reviewer @rhyshort 
reviewer @emlaver 

BugId: 47177


rhyshort:
+1

emlaver:
+1

PR #134, opened 2015-05-19
Add encryption of JSON datastore to sync-android

mikerhodes:
_What:_

This is the first PR along the `feature-encryption` path to allow developers to encrypt data stored on Android devices. It implements encrypting the JSON data on the device (i.e., the JSON data's SQLite database).

_Why:_

This feature is several PRs because it's a large feature, impossible to review all in one go. This PR shows including SQLCipher in sync-android along with encryption of one part of the data storage chain.

_How:_

SQLCipher .so, .jar files are now included in the sync-android build. At some point we might want to break these out into a separate sync-encryption class, but for now the jar is always there on Android. The final location of the .so files needs to be decided, as they are currently within the demo app file system tree.

The encryption follows the pattern of iOS. We have a KeyProvider interface which a developer uses to provide a 256-bit encryption key for use by SQLCipher to encrypt the database. We provide a SimpleKeyProvider which just stores the byte array the developer passes in, assuming that it's a securely generated key.

The developer passes the KeyProvider into the `openDatastore` method in `DatastoreManager`, which then passes it through to the datastore constructor, which passes it to the database queue, which passes it to the platform independent database wrapper class, which finally figures which concrete database implementation to use -- Android with or without encryption, or Java SE without encryption as we don't support encryption there yet.

A NullKeyProvider class is used to signal use of non-encrypted databases, this avoids duplicated code and `null` checks.

_Tests:_

Right now, there's a simple switch which runs all the tests using a SQLCipher-based database on an Android device or simulator.

Future work in improving these tests:

- Try to make sure test suite always tests both encrypted and non-encrypted code paths.
- Have explicit checks that when a key is passed, database is encrypted.

BugId: 43715

reviewer @tomblench 
reviewer @indisoluble 
reviewer @rhyshort


AndroidTest/app/src/main/java/cloudant/com/androidtest/MyActivity.java
```diff
@@ -48,78 +49,115 @@
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
+        //Stop the tests being re-run when we are returned to this activity
+        if(savedInstanceState == null) {
 
-        setTestProperties();
+            super.onCreate(savedInstanceState);
```
rhyshort: `super.onCreate(savedInstanceState);` should be called before we do our stuff, we always have to call the super to prevent exceptions being thrown see: http://developer.android.com/reference/android/app/Activity.html#onCreate%28android.os.Bundle%29

sync-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
```diff
@@ -153,8 +156,8 @@ public String getPath() {
     public Datastore openDatastore(String dbName) throws DatastoreNotCreatedException {
         Preconditions.checkArgument(dbName.matches(LEGAL_CHARACTERS),
                 "A database must be named with all lowercase letters (a-z), digits (0-9),"
-                  + " or any of the _$()+-/ characters. The name has to start with a"
-                  + " lowercase letter (a-z).");
+                        + " or any of the _$()+-/ characters. The name has to start with a"
```
tomblench: The one-arg version of `openDatastore` should call the two-arg `openDatastore` with NullKeyProvider as discussed.

sync-android/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLCipherSQLite.java
```diff
@@ -0,0 +1,233 @@
...
+
+import net.sqlcipher.database.SQLiteConstraintException;
+import net.sqlcipher.database.SQLiteDatabase;
+import net.sqlcipher.database.SQLiteException;
+
+public class AndroidSQLCipherSQLite extends SQLDatabase {
+
+    SQLiteDatabase database = null;
+    private Thread threadWhichOpened = null;
+    private ThreadLocal<Boolean> appearsOpen = null;
```
rhyshort: Thread Local code isn't needed here, it was removed from `AndroidSQLite` in PR #128 same should be applied here

sync-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
```diff
@@ -234,6 +272,36 @@ private Datastore createDatastore(String dbName) throws DatastoreNotCreatedExcep
         }
     }
 
+    /*
+    * Creates a datastore that requires SQLCipher encryption.
+    */
+    private Datastore createDatastore(String dbName, KeyProvider provider) throws DatastoreNotCreatedException {
```
tomblench: One-arg `createDatastore` can be removed once my above comment is implemented

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptionKey.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        if (key.length != REQUIRED_KEY_LENGTH) {
+            throw new IllegalArgumentException("Key array must be 32 bytes");
+        }
+
+        this.key = Arrays.copyOf(key, REQUIRED_KEY_LENGTH);
+    }
+
+    public byte[] getKey() {
```
tomblench: I would expect `getKey` to return a pointer to the key. If we want a copy it should be called `cloneKey` or `copyKey` etc.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -58,7 +60,7 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-class BasicDatastore implements Datastore, DatastoreExtended {
+public class BasicDatastore implements Datastore, DatastoreExtended {
```
rhyshort: What is the thinking behind making this public but not removing the interface?

sync-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
```diff
@@ -153,8 +156,8 @@ public String getPath() {
     public Datastore openDatastore(String dbName) throws DatastoreNotCreatedException {
         Preconditions.checkArgument(dbName.matches(LEGAL_CHARACTERS),
```
rhyshort: This should drop down into `openDatastore(Stirng dbName, KeyProvider provider)` providing a suitable default for the key provider

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
```diff
@@ -36,30 +36,47 @@
...
+                            .getMethod("createAndroidSQLite", String.class, KeyProvider.class)
+                            .invoke(null, new Object[]{dbFilename, provider});
+                } else {
+                    return (SQLDatabase) Class.forName("com.cloudant.sync.sqlite.android.AndroidSQLite")
+                            .getMethod("createAndroidSQLite", String.class)
+                            .invoke(null, dbFilename);
+                }
+            } else {
+                if (useSqlCipher) {
+                    throw new IOException("No SQLCipher-based database implementation for Java SE");
```
tomblench: is this really an `IOException`? i know that is the type in the `throws` clause but maybe it should be extended or we should throw an unchecked exception.

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
```diff
@@ -36,30 +36,47 @@
...
+                } else {
+                    return (SQLDatabase) Class.forName("com.cloudant.sync.sqlite.sqlite4java.SQLiteWrapper")
+                            .getMethod("openSQLiteWrapper", String.class)
+                            .invoke(null, dbFilename);
+                }
             }
+
+        } catch (Exception e) {
+            logger.log(Level.SEVERE, "Failed to load database module", e);
+            return null;
```
tomblench: i wonder if we should throw an exception here as well

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
```diff
@@ -73,25 +90,50 @@ public static SQLDatabase createSQLDatabase(String dbFilename) throws IOExceptio
...
+    /**
+     * SQLCipher-based implementation for opening database.
+     * @param dbFilename full file path of the db file
+     * @param provider Key provider object storing the SQLCipher key
+     *                 Supply a NullKeyProvider to use a non-encrypted database.
+     * @return {@code SQLDatabase} for the given filename
+     * @throws IOException if the file does not exists, and also
+     *         can not be created
+     */
+    public static SQLDatabase openSqlDatabase(String dbFilename, KeyProvider provider) throws IOException {
```
tomblench: there's a lot of shared code between `openSqlDatabase` and `createSQLDatabase`. since only the method names vary between the two can we collapse this down?

tomblench: in fact i'm not even convinced the underlying libraries make a distinction between open and create. for the non-sqlcipher it looks like the same methods `createAndroidSQLite`/`openSQLiteWrapper` are used for each case.

AndroidTest/gradle.properties
```diff
@@ -11,6 +11,12 @@
 # The setting is particularly useful for tweaking memory settings.
 # Default value: -Xmx10248m -XX:MaxPermSize=256m
 # org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+# If SQLCipher-based datastore is required, use the sqlcipher parameter:
+# org.gradle.jvmargs=-Dtest.sqlcipher.passphrase=true
+# For running tests under only one package use 'test.run-packages'
+# TODO: Add comma option for running under multiple packages
+# Example for running test cases under 'datastore' package:
+# org.gradle.jvmargs=-Dtest.run-packages=datastore
```
rhyshort: I think doc should mention `systemProp.test.run-packages=datastore` instead of `org.gradle.jvmargs`

sync-core/src/test/java/com/cloudant/sync/datastore/encryption/KeyUtilsTest.java
```diff
@@ -0,0 +1,89 @@
...
+                    @Override
+                    public EncryptionKey getEncryptionKey() {
+                        return key;
+                    }
+                }));
+    }
+
+    @Test
+    public void sqlCipherKeyForKeyProvider_encodes32BytesCorrectly() {
+        byte[] keyBytes = new byte[] { -123, 53, -22, -15, -123, 53, -22, -15, -123, 53, -22, -15,
```
tomblench: Personally i could understand the decimal/hex conversion if these were unsigned ints or even hex literals.

tomblench: (understand _better_ that is)

sync-android/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLCipherSQLite.java
```diff
@@ -0,0 +1,233 @@
...
+                "Input SQL can not be empty String.");
+        this.database.execSQL(sql);
+    }
+
+    public int status(int operation, boolean reset) {
+        return 0;
+    }
+
+    //TODO: Change password option that will be implemented with key management work items
+    public void changePassword(String password) throws SQLiteException {
```
indisoluble: Just for the record, this method is not part of the iOS API. It is a task recorded in FB but the method is not there yet, anyway it will be probably part of `CDTEncryptionKeychainManager`.

sync-core/src/test/java/com/cloudant/sync/datastore/encryption/KeyUtilsTest.java
mikerhodes: Oh, I'd love that too but Java's byte is always signed.

sync-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
mikerhodes: Will fix in future PR.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
mikerhodes: @emlaver ^

sync-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
mikerhodes: Future PR.

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
mikerhodes: http://docs.oracle.com/javase/7/docs/api/java/lang/UnsupportedOperationException.html?

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
mikerhodes: Our external code is using a combination of open/create and I'd like to review that separately from this PR.

AndroidTest/gradle.properties
mikerhodes: Could you suggest a rewrite, I'm unfamiliar with this. Also, @emlaver may have input.

sync-android/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLCipherSQLite.java
mikerhodes: Removed in f07a75a.

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptionKey.java
mikerhodes: Copy/clone are usually used in reference to the class they are written within. Personally I'd prefer to stick with `get` in this case because it feels more natural when reading code which uses this.

sync-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
emlaver: Fixed and commented in c318b074693332a9a4b86ba0fe3a037a487bbfad

AndroidTest/app/src/main/java/cloudant/com/androidtest/MyActivity.java
emlaver: Fixed in 16a184b09d4eadfd337205740120578fefbc60aa

sync-android/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLCipherSQLite.java
emlaver: Fixed in 2b356e14bf6e78a187a32984049829babbca3b81

AndroidTest/gradle.properties
emlaver: I was having trouble getting the prefix 'systemProp' to be recognized during debugging.  Will try again tomorrow.

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
rhyshort: UnsupportedOp would make the most sense I think

sync-android/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLCipherSQLite.java
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * Constructor for creating SQLCipher-based SQLite database.
+     * @param path full file path of the db file
+     * @param provider Provider object that contains the key to encrypt the SQLCipher database
+     * @return
+     */
+    public static AndroidSQLCipherSQLite createAndroidSQLite(String path, KeyProvider provider) {
+
+        //Call SQLCipher-based method for opening database, or creating if database not found
+        SQLiteDatabase db = SQLiteDatabase.openOrCreateDatabase(path,
```
indisoluble: Does `openOrCreateDatabase ` check if the key is correct? In iOS, I have to execute `SELECT count(*) FROM sqlite_master;` to know if it works (more details [here](https://www.zetetic.net/sqlcipher/sqlcipher-api/#key)).

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
mikerhodes: The exception gets caught and we return null anyway, though we print the stack. Should the UnsupportedOperationException be rethrown?

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
mikerhodes: I wasn't sure. At least, perhaps, rethrow the UnsupportedOperationException? Not sure on others.

I've a feeling I changed the behaviour here, before each of the reflected method calls were protected by this catch, not the JavaSE unsupported thing.

sync-android/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLCipherSQLite.java
mikerhodes: It doesn't, you're right I need to add this somewhere.

AndroidTest/gradle.properties
emlaver: Rhys, I'd like to keep the `org.gradle.jvmargs` property in the doc as it's required for passing the -D arguments to gradle.  And, within gradle, the `convertTestSysPropsToHash` method only takes arguments that start with 'test'.  I'd be OK with changing the prefix -Dtest to -DsystemProp.test but would need to make changes to the gradle method.

rhyshort: I'm not sure I follow, do you have options that aren't prefixed with `test.`?

sync-android/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLCipherSQLite.java
mikerhodes: Added in https://github.com/cloudant/sync-android/commit/3b87e44768f582e8ef5c4525d52cca062405eee2

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
mikerhodes: I added the UOE in https://github.com/cloudant/sync-android/commit/f9854b4f3e11f5b719e7833d9dbcd574267714a3

AndroidTest/gradle.properties
emlaver: Fixed in 432b064848931acd3e040abc033d8f79bc339514

AndroidTest/app/src/main/java/cloudant/com/androidtest/MyActivity.java
```diff
@@ -49,77 +50,113 @@
...
+                            } else if (testsToRun.equals("replication") && c.getPackage().toString().contains("replication")) {
+                                classes.add(c);
+                            } else if (testsToRun.equals("sqlite") && c.getPackage().toString().contains("sqlite")) {
+                                classes.add(c);
+                            } else if (testsToRun.equals("util") && c.getPackage().toString().contains("util")) {
+                                classes.add(c);
+                            }
+                        } else {
+                            classes.add(c);
+                        }
```
rhyshort: The above code for package filtering could be replaced with something much simpler:

```
ArrayList<Class> classes = new ArrayList<Class>();
//Check parameter for specifying which test case package to run
String testsToRun = System.getProperty("test.run-packages");
                    for (Class c : BuildConfig.classToTests) {

                        //Run all classes if JVM parameter for specifying packages does not exist
                        if (testsToRun != null) {
                        
                          if(c.getPackage().toString().contains(testsToRun)){
                              classes.add(c);
                          }

                        } else {
                            classes.add(c);
                       }
```

sync-core/src/test/java/com/cloudant/sync/datastore/encryption/KeyUtilsTest.java
tomblench: Oh yes...I learnt a thing

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptionKey.java
tomblench: We can agree to disagree on the naming, but what's the reason for making a copy? Just return the pointer and if the user wants to their own writable copy they can make one.

sync-core/src/test/java/com/cloudant/sync/datastore/encryption/KeyUtilsTest.java
mikerhodes: Me too!

AndroidTest/app/src/main/java/cloudant/com/androidtest/MyActivity.java
mikerhodes: I'd like to fix this in another PR.

indisoluble:
I have no more comments to do so :+1:

rhyshort:
:+1:

tomblench:
+1 can you track the 'other PR' stuff in a ticket or somewhere else?

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptionKey.java
mikerhodes: I want to avoid the user writing into the copy stored in the EncryptionKey object. It seems like that'd be a recipe for hard to debug issues.

mikerhodes:
I have separate tickets filed for the "other PR" stuff now.

PR #135, opened 2015-05-19
Fix FTS enabled check

alfinkel:
_What_

Text search tests fail during Android testing because of SQLite idiosyncrasies on Android.

_Why_

We need to ensure that text search will work on Android mobile devices as well as we need to ensure that our build tests pass without failures.

_How_

Since the problem is that the `PRAGMA compile_options` command is returning an empty cursor when executed on Android, change the FTS availability check to instead create a SQL FTS virtual table.  Successful table creation means that FTS is available whereas a failed table create means that FTS is not enabled.

_Tests_

- IndexManagerTest: removed the check for `ENABLE_FTS3` and `ENABLE_FTS3_PARENTHESIS` compile options in `validateTextSearchIsAvailable()`, since this check will fail in Android.
- Add conditional logic to two enhanced query syntax tests to check for `ENABLE_FTS3` and `ENABLE_FTS3_PARENTHESIS` and only execute the content of the tests if enhanced query syntax is available.  This conditional logic allows us to still run the enhanced query syntax tests during Java SE testing.

reviewer @mikerhodes 
reviewer @rhyshort 

BugId: 46093

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
@@ -71,6 +71,7 @@ import java.util.regex.Pattern;
 public class IndexManager {
 
     private static final String INDEX_TABLE_PREFIX = "_t_cloudant_sync_query_index_";
+    private static final String FTS_CHECK_TABLE_NAME = "_t_cloudant_sync_query_fts_check";
     public static final String INDEX_METADATA_TABLE_NAME = "_t_cloudant_sync_query_metadata";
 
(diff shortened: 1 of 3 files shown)


rhyshort:
+1

mikerhodes:
Do we need to update any docs where we say that if we say text search is enabled you can use the enhanced syntax?

mikerhodes:
@alfinkel tells me offline that we don't need to update docs.

:+1:

alfinkel:
The docs are already up to date.  They state that which ever syntax is enabled is the syntax the developer can use.  They also state that the likelihood out of the box is that standard query syntax will likely be enabled on the Android platform.

PR #136, opened 2015-05-21
44538 filter pipeline and cookie filter.

rhyshort:
What:

Introduce filter pipeline for modifying requests and replaying requests based on responses. Also add a cookie filter to make use of the filtering mechanism.

Why:

Enables developers to add information to http request their system may need, for example tokens for 
their proxy set up. The most asked for authentication method is cookie, as a result I have directly
implemented the filter for their use and as a way of "dog fooding" our filter API.

How:

Two interfaces, one for interacting with requests and one for responses. Request filters
implement the `HttpConnectionRequestFilter` interface, response filters should implement `HttpConnectionResponseFilter` interface. A class can implement both interfaces without 
collision, the methods to implement are called `filterRequest` and `filterResponse` respectively.  

Cookie filter interacts with both requests and responses to correctly get and set the 
cookie for requests with couch/cloudant. It will instruct the HttpConnection to replay requests in the event of a 401 error.

Testing:

Set up a local couch server with an admin, along with a protected database called "cookie_test". Test should be able to add a document containing "hello:world" into the database.

reviewer @mikerhodes 
reviewer @tomblench

### DIFF
--- a/sync-core/src/main/java/com/cloudant/http/CookieFilter.java
+++ b/sync-core/src/main/java/com/cloudant/http/CookieFilter.java
@@ -1,0 +1,158 @@
+/*
+ * Copyright (c) 2015 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package com.cloudant.http;
+
+import com.cloudant.mazha.json.JSONHelper;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ *
+ * Adds cookie authentication support to http requests.
+ *
+ * It does this by adding the cookie header for CouchDB
+ * using request filtering pipeline in {@link HttpConnection}.
+ *
+ * If a response has a response code of 401, it will fetch a cookie from
+ * the server using provided credentials and tell {@link HttpConnection} to reply
+ * the request by setting {@link HttpConnectionFilterContext#replayRequest} property to true.
+ *
+ * If the request to get the cookie for use in future request fails with a 401 status code
+ * (or any status that indicates client error) cookie authentication will not be attempted again.
+ *
+ *
+ */
+public  class CookieFilter implements HttpConnectionRequestFilter, HttpConnectionResponseFilter {
+
+    private final static Logger logger = Logger.getLogger(CookieFilter.class.getCanonicalName());
+    final String sessionRequestBody;
+    private String cookie = null;
+    private boolean shouldAttemptCookieRequest = true;
+    private final String username;
+
+    /**
+     * Constructs a cookie filter.
+     * @param username The username to use when getting the cookie
+     * @param password The password to use when getting the cookie
+     */
+    public CookieFilter(String username, String password){
+        this.sessionRequestBody = String.format("name=%s&password=%s",username,password);
+        this.username = username;
+    }
+
+    @Override
+    public HttpConnectionFilterContext filterRequest(HttpConnectionFilterContext context) {
+
+        HttpURLConnection connection = context.connection.getConnection();
+
+        if(shouldAttemptCookieRequest) {
+            if (cookie == null) {
+                cookie = getCookie(connection.getURL());
+            }
+            connection.setRequestProperty("Cookie", cookie);
+        }
+
+        return context;
+    }
+
+    @Override
+    public HttpConnectionFilterContext filterResponse(HttpConnectionFilterContext context) {
+        HttpURLConnection connection = context.connection.getConnection();
+        try {
+            if (context.connection.getConnection().getResponseCode() == 401) {
+                //we need to get a new cookie
+                cookie = getCookie(connection.getURL());
+                //don't resend request, failed to get cookie
+                if(cookie != null) {
+                    context.replayRequest = true;
+                } else {
+                    context.replayRequest = false;
+                }
+            }
+        } catch (IOException e) {
+            logger.log(Level.SEVERE, "Failed to get response code from request",e);
+        }
+        return context;
+
+    }
+
+     private String getCookie(URL url){
+        try {
+            URL sessionURL = new URL(String.format("%s://%s:%d/_session",
+                    url.getProtocol(),
+                    url.getHost(),
+                    url.getPort()));
+
+            HttpConnection conn = Http.POST(sessionURL, "application/x-www-form-urlencoded");
+            conn.setRequestBody(sessionRequestBody);
+            HttpURLConnection connection = conn.execute().getConnection();
+            String cookieHeader = connection.getHeaderField("Set-Cookie");
+            int responseCode = connection.getResponseCode();
+
+            if(responseCode / 100 == 2){
+
+                if(sessionHasStarted(connection.getInputStream())) {
+                    return cookieHeader.substring(0, cookieHeader.indexOf(";"));
+                } else {
+                    return null;
+                }
+
+            } else if(responseCode == 401){
+                shouldAttemptCookieRequest  = false;
+                logger.severe("Credentials are incorrect, cookie authentication will not be" +
+                        " attempted again by this filter object");
+            } else if (responseCode / 100 == 5){
+                logger.log(Level.SEVERE,
+                        "Failed to get cookie from server, response code %s, cookie auth",
+                        responseCode);
+            }  else {
+                // catch any other response code
+                logger.log(Level.SEVERE,
+                        "Failed to get cookie from server, response code %s, " +
+                                "cookie authentication will not be attempted again",
+                        responseCode);
+                shouldAttemptCookieRequest = false;
+            }
+
+        } catch (MalformedURLException e) {
+            logger.log(Level.SEVERE,"Failed to create URL for _session endpoint",e);
+        } catch (UnsupportedEncodingException e) {
+            logger.log(Level.SEVERE, "Failed to encode cookieRequest body", e);
+        } catch (IOException e) {
+            logger.log(Level.SEVERE, "Failed to read cookie response header", e);
+        }
+        return null;
+    }
+
+    private boolean sessionHasStarted(InputStream responseStream){
+        //check the response body
+        JSONHelper jsonHelper = new JSONHelper();
+        Map<String,Object> jsonResponse = jsonHelper.fromJson(new InputStreamReader(responseStream));
+
+        // only check for ok:true, https://issues.apache.org/jira/browse/COUCHDB-1356
+        // means we cannot check that the name returned is the one we sent.
+       return jsonResponse.containsKey("ok") && (Boolean) jsonResponse.get("ok");
+
+    }
+}
--- a/sync-core/src/main/java/com/cloudant/http/Http.java
+++ b/sync-core/src/main/java/com/cloudant/http/Http.java
@@ -12,7 +12,6 @@ package com.cloudant.http;
 
 import java.net.MalformedURLException;
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.net.URL;
 
 /**
--- a/sync-core/src/main/java/com/cloudant/http/HttpConnection.java
+++ b/sync-core/src/main/java/com/cloudant/http/HttpConnection.java
@@ -23,11 +23,12 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
-import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Properties;
+import java.util.logging.Logger;
+
 
 /**
  * Created by tomblench on 23/03/15.
@@ -67,6 +68,7 @@ import java.util.Properties;
  */
 public class HttpConnection  {
 
+    private static final Logger logger = Logger.getLogger(HttpConnection.class.getCanonicalName());
     private final String requestMethod;
     public final URL url;
     private final String contentType;
@@ -82,6 +84,12 @@ public class HttpConnection  {
 
     private static String userAgent = getUserAgent();
 
+    public final List<HttpConnectionRequestFilter> requestFilters;
+    public final List<HttpConnectionResponseFilter> responseFilters;
+
+    private int numberOfRetries = 10;
+
+
     public HttpConnection(String requestMethod,
                           URL url,
                           String contentType) {
@@ -89,6 +97,17 @@ public class HttpConnection  {
         this.url = url;
         this.contentType = contentType;
         this.requestProperties = new HashMap<String, String>();
+        this.requestFilters = new LinkedList<HttpConnectionRequestFilter>();
+        this.responseFilters = new LinkedList<HttpConnectionResponseFilter>();
+    }
+
+    /**
+     * Sets the number of times this request can be retried.
+     * This method <strong>must</strong> be called before {@link #execute()}
+     * @param numberOfRetries the number of times this request can be retried.
+     */
+    public void setNumberOfRetries(int numberOfRetries){
+        this.numberOfRetries = numberOfRetries;
     }
 
     /**
@@ -142,72 +161,89 @@ public class HttpConnection  {
      * @throws IOException if there was a problem writing data to the server
      */
     public HttpConnection execute() throws IOException {
-        System.setProperty("http.keepAlive", "false");
-        connection = (HttpURLConnection) url.openConnection();
-        // always read the result, so we can retrieve the HTTP response code
-        connection.setDoInput(true);
-        connection.setRequestMethod(requestMethod);
+            boolean retry = true;
+            int n = numberOfRetries;
+            while (retry && n-- > 0) {
 
-        // Set request headers
-        List<String> lowerCaseExistingHeaders = new ArrayList<String>();
-        for (String key : requestProperties.keySet()) {
-            connection.setRequestProperty(key, requestProperties.get(key));
-            lowerCaseExistingHeaders.add(key.toLowerCase());
-        }
+                System.setProperty("http.keepAlive", "false");
+
+                connection = (HttpURLConnection) url.openConnection();
+                for (String key : requestProperties.keySet()) {
+                    connection.setRequestProperty(key, requestProperties.get(key));
+                }
+
+                connection.setRequestProperty("User-Agent", userAgent);
+                if (url.getUserInfo() != null) {
+                    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+                    OutputStream bos = Base64OutputStreamFactory.get(baos);
+                    bos.write(url.getUserInfo().getBytes());
+                    bos.flush();
+                    bos.close();
+                    String encodedAuth = baos.toString();
+                    connection.setRequestProperty("Authorization", String.format("Basic %s", encodedAuth));
+                }
 
 
-        if (contentType != null && !lowerCaseExistingHeaders.contains("content-type")) {
-            connection.setRequestProperty("Content-type", contentType);
-        }
+                // always read the result, so we can retrieve the HTTP response code
+                connection.setDoInput(true);
+                connection.setRequestMethod(requestMethod);
+                if (contentType != null) {
+                    connection.setRequestProperty("Content-type", contentType);
+                }
 
-        if (!lowerCaseExistingHeaders.contains("user-agent")) {
-            connection.setRequestProperty("User-Agent", userAgent);
-        }
+                HttpConnectionFilterContext currentContext = new HttpConnectionFilterContext(this);
 
-        if (url.getUserInfo() != null && !lowerCaseExistingHeaders.contains("authorization")) {
-            ByteArrayOutputStream baos = new ByteArrayOutputStream();
-            OutputStream bos = Base64OutputStreamFactory.get(baos);
-            bos.write(url.getUserInfo().getBytes());
-            bos.flush();
-            bos.close();
-            String encodedAuth = baos.toString();
-            connection.setRequestProperty("Authorization", String.format("Basic %s", encodedAuth));
-        }
+                for (HttpConnectionRequestFilter requestFilter : requestFilters) {
+                    currentContext = requestFilter.filterRequest(currentContext);
+                }
 
-        if (input != null) {
-            connection.setDoOutput(true);
-            if (inputLength != -1) {
-                // TODO on 1.7 upwards this method takes a long, otherwise int
-                connection.setFixedLengthStreamingMode((int)this.inputLength);
-            } else {
-                // TODO some situations where we can't do chunking, like multipart/related
-                /// https://issues.apache.org/jira/browse/COUCHDB-1403
-                connection.setChunkedStreamingMode(1024);
+                if (input != null) {
+                    connection.setDoOutput(true);
+                    if (inputLength != -1) {
+                        // TODO on 1.7 upwards this method takes a long, otherwise int
+                        connection.setFixedLengthStreamingMode((int) this.inputLength);
+                    } else {
+                        // TODO some situations where we can't do chunking, like multipart/related
+                        /// https://issues.apache.org/jira/browse/COUCHDB-1403
+                        connection.setChunkedStreamingMode(1024);
+                    }
+
+                    // See "8.2.3 Use of the 100 (Continue) Status" in http://tools.ietf.org/html
+                    // /rfc2616
+                    // Attempting to write to the connection's OutputStream may cause an exception to be
+                    // thrown. This is useful because it avoids sending large request bodies (such as
+                    // attachments) if the server is going to reject our request. Reasons for rejecting
+                    // requests could be 401 Unauthorized (eg cookie needs to be refreshed), etc.
+                    connection.setRequestProperty("Expect", "100-continue");
+
+                    int bufSize = 1024;
+                    int nRead = 0;
+                    byte[] buf = new byte[bufSize];
+                    InputStream is = input;
+                    OutputStream os = connection.getOutputStream();
+
+                    while ((nRead = is.read(buf)) >= 0) {
+                        os.write(buf, 0, nRead);
+                    }
+                    os.flush();
+                    // we do not call os.close() - on some JVMs this incurs a delay of several seconds
+                    // see http://stackoverflow.com/questions/19860436
+                }
+
+                for (HttpConnectionResponseFilter responseFilter : responseFilters) {
+                    currentContext = responseFilter.filterResponse(currentContext);
+                }
+
+                // retry flag is set from the final step in the response filterRequest pipeline
+                retry = currentContext.replayRequest;
+
+                if (n == 0) {
+                    logger.info("Maximum number of retries reached");
+                }
             }
-
-            // See "8.2.3 Use of the 100 (Continue) Status" in http://tools.ietf.org/html/rfc2616
-            // Attempting to write to the connection's OutputStream may cause an exception to be
-            // thrown. This is useful because it avoids sending large request bodies (such as
-            // attachments) if the server is going to reject our request. Reasons for rejecting
-            // requests could be 401 Unauthorized (eg cookie needs to be refreshed), etc.
-            connection.setRequestProperty("Expect", "100-continue");
-
-            int bufSize = 1024;
-            int nRead = 0;
-            byte[] buf = new byte[bufSize];
-            InputStream is = input;
-            OutputStream os = connection.getOutputStream();
-
-            while((nRead = is.read(buf)) >= 0) {
-                os.write(buf, 0, nRead);
-            }
-            os.flush();
-            // we do not call os.close() - on some JVMs this incurs a delay of several seconds
-            // see http://stackoverflow.com/questions/19860436
+            // return ourselves to allow method chaining
+            return this;
         }
-        // return ourselves to allow method chaining
-        return this;
-    }
 
     /**
      * <p>
@@ -276,6 +312,7 @@ public class HttpConnection  {
         return connection;
     }
 
+
     private static String getUserAgent() {
         String userAgent;
         String ua = getUserAgentFromResource();
@@ -309,5 +346,4 @@ public class HttpConnection  {
             return defaultUserAgent;
         }
     }
-
 }
--- a/sync-core/src/main/java/com/cloudant/http/HttpConnectionFilterContext.java
+++ b/sync-core/src/main/java/com/cloudant/http/HttpConnectionFilterContext.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2015 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package com.cloudant.http;
+
+/**
+ * Created by tomblench on 30/03/15.
+ */
+public class HttpConnectionFilterContext {
+
+    public boolean replayRequest;
+    public final HttpConnection connection;
+
+    /**
+     * Constructor
+     * @param connection HttpConnection
+     */
+    public HttpConnectionFilterContext (HttpConnection connection) {
+        this.replayRequest = false;
+        this.connection = connection;
+    }
+
+    /**
+     * Shallow copy constructor
+     * @param other Context to copy
+     */
+    public HttpConnectionFilterContext (HttpConnectionFilterContext other) {
+        this.replayRequest = other.replayRequest;
+        this.connection = other.connection;
+
+    }
+
+}
--- a/sync-core/src/main/java/com/cloudant/http/HttpConnectionRequestFilter.java
+++ b/sync-core/src/main/java/com/cloudant/http/HttpConnectionRequestFilter.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2015 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package com.cloudant.http;
+
+/**
+ * Created by tomblench on 30/03/15.
+ */
+
+import java.net.HttpURLConnection;
+
+/**
+
+ A Request Filter is run before the request is made to the server. It can use headers to add support
+ for other authentication methods, for example cookie authentication. See
+ {@link CookieFilter#filterRequest(HttpConnectionFilterContext)} for an example.
+
+
+ Filters are executed in a pipeline and modify the context in a serial fashion.
+ */
+
+
+public interface HttpConnectionRequestFilter {
+
+    /**
+     * Filter the request.
+     * This method <strong>must not</strong> do any of the following:
+     * <ul>
+     *     <li>Return null</li>
+     *     <li>Call methods on the underlying {@link java.net.HttpURLConnection} which
+     *     initiate a request such as {@link HttpURLConnection#getResponseCode()}</li>
+     * </ul>
+     * @param context Input context
+     * @return Output context
+     */
+    HttpConnectionFilterContext filterRequest(HttpConnectionFilterContext context);
+
+}
--- a/sync-core/src/main/java/com/cloudant/http/HttpConnectionResponseFilter.java
+++ b/sync-core/src/main/java/com/cloudant/http/HttpConnectionResponseFilter.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2015 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package com.cloudant.http;
+
+/**
+ A Response Filter is run after the response is obtained from the
+ server but before the output stream is returned to the original client. The Response
+ Filter enables two main behaviours:
+<ul>
+  <li>Modifying the response for every request</li>
+
+ <li> Replaying a (potentially modified) request by reacting to the
+ response.</li>
+ </ul>
+
+ Filters are executed in a pipeline and modify the context in a serial fashion.
+ */
+public interface HttpConnectionResponseFilter {
+
+    /**
+     * Filter the response
+     *
+     * This method <strong>must not</strong> do any of the following
+     * <ul>
+     *     <li>Return null</li>
+     *     <li>Read the response stream</li>
+     * </ul>
+     * @param context Input context
+     * @return Output context
+     */
+    HttpConnectionFilterContext filterResponse(HttpConnectionFilterContext context);
+}
--- a/sync-core/src/test/java/com/cloudant/common/CouchTestBase.java
+++ b/sync-core/src/test/java/com/cloudant/common/CouchTestBase.java
@@ -57,7 +57,7 @@ public abstract class CouchTestBase {
             System.getProperty("test.couch.ignore.compaction",Boolean.FALSE.toString()));
 
     public static final Boolean IGNORE_AUTH_HEADERS = Boolean.valueOf(
-            System.getProperty("test.couch.ignore.auth.headers",Boolean.FALSE.toString()));
+            System.getProperty("test.couch.ignore.auth.headers",Boolean.TRUE.toString()));
 
     public CouchConfig getCouchConfig(String db) {
 
--- a/sync-core/src/test/java/com/cloudant/http/HttpTest.java
+++ b/sync-core/src/test/java/com/cloudant/http/HttpTest.java
@@ -14,13 +14,17 @@ import com.cloudant.common.CouchTestBase;
 import com.cloudant.common.RequireRunningCouchDB;
 import com.cloudant.mazha.CouchClient;
 import com.cloudant.mazha.CouchConfig;
+import com.cloudant.mazha.json.JSONHelper;
 
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.Map;
 
 /**
  * Created by tomblench on 06/05/15.
@@ -118,5 +122,50 @@ public class HttpTest extends CouchTestBase {
         // stream was not read because execute() was not called
         Assert.assertEquals(bis.available(), data.getBytes().length);
         client.deleteDb();
+    }
+
+
+    //NOTE: This test doesn't work with specified couch servers,
+    // the URL will always include the creds specified for the test
+    //
+    // A couchdb server needs to be set and running with the correct
+    // security settings, the database *must* not be public, it *must*
+    // be named cookie_test
+    //
+    @Test
+    public void testCookieAuthWithoutRetry() throws IOException {
+
+        Assume.assumeFalse(IGNORE_AUTH_HEADERS);
+
+        CookieFilter filter = new CookieFilter(System.getProperty("test.couch.username"),
+                System.getProperty("test.couch.password"));
+
+        CouchConfig config = getCouchConfig("cookie_test");
+        HttpConnection conn = new HttpConnection("POST", config.getRootUri().toURL(),
+                "application/json");
+        conn.responseFilters.add(filter);
+        conn.requestFilters.add(filter);
+        ByteArrayInputStream bis = new ByteArrayInputStream(data.getBytes());
+
+        // nothing read from stream
+        Assert.assertEquals(bis.available(), data.getBytes().length);
+
+        conn.setRequestBody(bis);
+        conn.execute();
+
+        // stream was read to end
+        Assert.assertEquals(bis.available(), 0);
+        Assert.assertEquals(2, conn.getConnection().getResponseCode() / 100);
+
+        //check the json
+        JSONHelper helper = new JSONHelper();
+        Map<String,Object> jsonRes = helper.fromJson(new InputStreamReader(conn.getConnection()
+                .getInputStream()));
+
+        Assert.assertTrue(jsonRes.containsKey("ok"));
+        Assert.assertTrue((Boolean)jsonRes.get("ok"));
+        Assert.assertTrue(jsonRes.containsKey("id"));
+        Assert.assertTrue(jsonRes.containsKey("rev"));
+
     }
 }


sync-core/src/main/java/com/cloudant/http/Http.java
```diff
@@ -26,6 +25,7 @@
  */
 public class Http {
 
+
     // high level http operations, URL and URI flavoured
```
rhyshort: I need to revert the changes to file, they shouldn't be here.

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,107 @@
...
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Created by Rhys Short on 15/05/15.
+ */
+public  class CookieFilter implements HttpConnectionRequestFilter, HttpConnectionResponseFilter {
```
tomblench: javadoc comments, on class and methods, what does it do, why etc

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,107 @@
...
+            context = new HttpConnectionFilterContext(context);
+        }
+        return context;
+
+        }
+
+     String getCookie(URL url){
+
+        try {
+            StringBuilder sb = new StringBuilder();
```
tomblench: I can see this URL mangling is ugly but necessary. Maybe string.format instead?

sync-core/src/main/java/com/cloudant/http/HttpConnectionFilterContext.java
```diff
@@ -0,0 +1,30 @@
+package com.cloudant.http;
```
tomblench: Copyright headers (yes, it was my fault in the first place...)

sync-core/src/main/java/com/cloudant/http/HttpConnectionRequestFilter.java
```diff
@@ -0,0 +1,28 @@
+package com.cloudant.http;
```
tomblench: Copyright headers again

rhyshort: Copyrights added in 75346b6

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -133,56 +145,79 @@ public void setRequestBody(InputStream input, long inputLength) {
...
-            }
 
-            // See "8.2.3 Use of the 100 (Continue) Status" in http://tools.ietf.org/html/rfc2616
-            // Attempting to write to the connection's OutputStream may cause an exception to be
-            // thrown. This is useful because it avoids sending large request bodies (such as
-            // attachments) if the server is going to reject our request. Reasons for rejecting
-            // requests could be 401 Unauthorized (eg cookie needs to be refreshed), etc.
-            connection.setRequestProperty("Expect", "100-continue");
+        boolean retry = true;
+        int n = maxRetries;
```
tomblench: Make retries configurable and log a message when they have been exhausted?

rhyshort: Any idea how to expose this to the user, via a setter maybe?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,118 @@
...
+            context = new HttpConnectionFilterContext(context);
+        }
+        return context;
+
+        }
+
+     private String getCookie(URL url){
+
+        try {
+            URL sessionURL = new URL(String.format("%s://%s:%s/_session",
```
tomblench: %d for port?

rhyshort: we can get away with %s it'll wrap the number in a string iirc

tomblench: fine but say what you mean, %d is the conversion for numbers

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,118 @@
...
+
+            HttpConnection conn = Http.POST(sessionURL, "application/json");
+            conn.setRequestBody(cookieRequestBody.getBytes("UTF-8"));
+            String cookieHeader = conn.execute().getConnection().getHeaderField("Set-Cookie");
+            return cookieHeader.substring(0,cookieHeader.indexOf(";"));
+
+
+
+        } catch (MalformedURLException e) {
+            e.printStackTrace();
```
tomblench: Need better error logging and handling than this. If we can't get the cookie maybe we should abandon and set replayRequest to false?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,126 @@
...
+                connection.setRequestProperty("Cookie", cookie);
+
+                context = new HttpConnectionFilterContext(context);
+            } else {
+                context.replayRequest = false;
+            }
+        }
+        return context;
+
+        }
```
tomblench: left indent this brace so it matches the opening brace

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,126 @@
...
+            URL sessionURL = new URL(String.format("%s://%s:%d/_session",
+                    url.getProtocol(),
+                    url.getHost(),
+                    url.getPort()));
+
+            HttpConnection conn = Http.POST(sessionURL, "application/json");
+            conn.setRequestBody(cookieRequestBody.getBytes("UTF-8"));
+            String cookieHeader = conn.execute().getConnection().getHeaderField("Set-Cookie");
+            return cookieHeader.substring(0,cookieHeader.indexOf(";"));
+
```
tomblench: Excessive newlines

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,126 @@
...
+    @Override
+    public HttpConnectionFilterContext filterRequest(HttpConnectionFilterContext context) {
+
+        HttpURLConnection connection = context.connection.getConnection();
+
+        if(cookie == null){
+            cookie = getCookie(connection.getURL());
+        }
+        connection.setRequestProperty("Cookie",cookie);
+
```
tomblench: Excessive newlines

tomblench:
:+1: once you've fixed up my last minor comments

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: done in 2ed5f402d611188dc027c7748a64b9396376ab9b

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: done in 2ed5f402d611188dc027c7748a64b9396376ab9b

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: done in  237f35e

tomblench:
OK, do we think about how to expose the filters for replications etc in a separate PR?

rhyshort:
I would think so, its probably going to turn into a long conversation about where and how we can do it

tomblench:
Hah OK , better merge this one first then :+1:

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,124 @@
...
+     * Constructs a cookie filter.
+     * @param username The username to use when getting the cookie
+     * @param password The password to use when getting the cookie
+     */
+    public CookieFilter(String username, String password){
+        Map<String,String> cookieRequestMap = new HashMap<String, String>();
+        cookieRequestMap.put("name",username);
+        cookieRequestMap.put("password",password);
+        JSONHelper helper = new JSONHelper();
+        cookieRequestBody = helper.toJson(cookieRequestMap);
```
mikerhodes: This is actually the session request body, not the cookie body.

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,124 @@
...
+        try {
+            URL sessionURL = new URL(String.format("%s://%s:%d/_session",
+                    url.getProtocol(),
+                    url.getHost(),
+                    url.getPort()));
+
+            HttpConnection conn = Http.POST(sessionURL, "application/json");
+            conn.setRequestBody(cookieRequestBody.getBytes("UTF-8"));
+            String cookieHeader = conn.execute().getConnection().getHeaderField("Set-Cookie");
+            return cookieHeader.substring(0,cookieHeader.indexOf(";"));
```
mikerhodes: How do you know the cookie you want is the first one?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Override
+    public HttpConnectionFilterContext filterRequest(HttpConnectionFilterContext context) {
+
+        HttpURLConnection connection = context.connection.getConnection();
+
+        if(cookie == null){
+            cookie = getCookie(connection.getURL());
+        }
+        connection.setRequestProperty("Cookie",cookie);
```
mikerhodes: Can't you use `hc.requestProperties.put("x-some-header", "some-value");` as described in HttpConnection javadoc here?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: The documentation for [couchdb] (https://people.apache.org/~dch/snapshots/couchdb/20140715/new-docs-build/html/api/server/authn.html#post--_session) indicates only one cookie is returned

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,124 @@
...
+        HttpURLConnection connection = context.connection.getConnection();
+        if (context.connection.getResponseCode() == 401) {
+            //we need to get a new cookie
+            cookie = getCookie(connection.getURL());
+            //don't resend request, failed to get cookie
+            if(cookie != null) {
+                context.replayRequest = true;
+                connection.setRequestProperty("Cookie", cookie);
+
+                context = new HttpConnectionFilterContext(context);
```
mikerhodes: Why do you have to create a new context here, but not in the `else` case?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -133,56 +155,82 @@ public void setRequestBody(InputStream input, long inputLength) {
...
+                bos.flush();
+                bos.close();
+                String encodedAuth = baos.toString();
+                connection.setRequestProperty("Authorization", String.format("Basic %s", encodedAuth));
+            }
+            
+            HttpConnectionFilterContext currentContext = new HttpConnectionFilterContext(this);
+            for (HttpConnectionRequestFilter requestFilter : requestFilters) {
+                currentContext = requestFilter.filterRequest(currentContext);
+            }
```
mikerhodes: What happens on exceptions? Or, is just giving up on exceptions in filters the correct thing to do?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -133,56 +155,82 @@ public void setRequestBody(InputStream input, long inputLength) {
...
+                }
+                os.flush();
+                // we do not call os.close() - on some JVMs this incurs a delay of several seconds
+                // see http://stackoverflow.com/questions/19860436
+            }
+            this.responseCode = connection.getResponseCode();
+            
+            for (HttpConnectionResponseFilter responseFilter : responseFilters) {
+                currentContext = responseFilter.filterResponse(currentContext);
+            }
```
mikerhodes: What happens if the filter reads the response content?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -133,56 +155,82 @@ public void setRequestBody(InputStream input, long inputLength) {
...
+            
+            // retry flag is set from the final step in the response filterRequest pipeline
+            retry = currentContext.replayRequest;
+        }
+        if(n < 0){
+            logger.info("Maximum number of retries reached");
```
mikerhodes: More of an error?

sync-core/src/main/java/com/cloudant/http/HttpConnectionRequestFilter.java
```diff
@@ -0,0 +1,41 @@
...
+ * and limitations under the License.
+ */
+package com.cloudant.http;
+
+/**
+ * Created by tomblench on 30/03/15.
+ */
+
+/**
+ A <code>HttpConnectionFilter</code> can either be configured as a Request Filter or a Response Filter:
```
mikerhodes: `HttpConnectionFilter` isn't a noun in our system right now, so I'm not sure it should be here -- perhaps this whole sentence should be removed.

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: I could, but I'd have to create a list since requestProperties is defined as `Map<String,List<String>>` [see java doc](https://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#getRequestProperties%28%29) I'd have to create a list object which makes the code harder to read when I can just call `setRequestProperty`

sync-core/src/main/java/com/cloudant/http/HttpConnectionRequestFilter.java
```diff
@@ -0,0 +1,41 @@
...
+
+ A Request Filter is run before the request is made to the server. It can use headers to add support
+ for other authentication methods, for example cookie authentication. See
+ {@link CookieFilter#filterRequest(HttpConnectionFilterContext)} for an example.
```
mikerhodes: Probably want to be more specific here, called after headers X and Y are set, but before body is attached etc. What if the filter wants to replace the body being sent?
What are the things this class can safely do with a context?

sync-core/src/main/java/com/cloudant/http/HttpConnectionResponseFilter.java
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package com.cloudant.http;
+
+/**
+ A <code>HttpConnectionResponseFilter</code>/
```
mikerhodes: Sentence is ... odd, should it be removed?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,124 @@
...
+    @Override
+    public HttpConnectionFilterContext filterResponse(HttpConnectionFilterContext context) {
+        HttpURLConnection connection = context.connection.getConnection();
+        if (context.connection.getResponseCode() == 401) {
+            //we need to get a new cookie
+            cookie = getCookie(connection.getURL());
+            //don't resend request, failed to get cookie
+            if(cookie != null) {
+                context.replayRequest = true;
+                connection.setRequestProperty("Cookie", cookie);
```
mikerhodes: Why does the Response filter need to set the cookie? Isn't that the job of the request filter?

sync-core/src/main/java/com/cloudant/http/HttpConnectionResponseFilter.java
```diff
@@ -0,0 +1,43 @@
...
+ Filter enables two main behaviours:
+
+ - Modifying the response for every request
+
+ - Replaying a (potentially modified) request by reacting to the
+ response. For example, obtaining a cookie on receipt of a 401
+ response code, modifying the "Cookie" header of the original
+ request, then setting replayRequest to <code>true</code> to replay the request
+ with the new "Cookie" header, see {@link CookieFilter#filterResponse(HttpConnectionFilterContext)}
+ for an example of replying requests based on the response code.
```
mikerhodes: Proper HTML needed.

mikerhodes: Example shouldn't be in the bullets.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
mikerhodes: Behaviour if a filter returns null?

sync-core/src/main/java/com/cloudant/http/HttpConnectionRequestFilter.java
```diff
@@ -0,0 +1,41 @@
...
+
+
+ Filters are executed in a pipeline and modify the context in a serial fashion.
+ */
+
+
+public interface HttpConnectionRequestFilter {
+
+    /**
+     * Filter the request or response
```
mikerhodes: Isn't this just the request?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: There was a reason, but it I can't remember right now

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: I think its fine as an info, but its more of a question for @tomblench

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+
+    @Override
+    public HttpConnectionFilterContext filterRequest(HttpConnectionFilterContext context) {
+
+        HttpURLConnection connection = context.connection.getConnection();
+
+        if(cookie == null){
+            cookie = getCookie(connection.getURL());
```
mikerhodes: * Is there some way to stop the request if the creds are not valid?
* If the creds are not valid, which `_session` will tell you, should this class avoid making tons of requests which will always fail?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,124 @@
...
+        connection.setRequestProperty("Cookie",cookie);
+
+        return context;
+
+    }
+
+    @Override
+    public HttpConnectionFilterContext filterResponse(HttpConnectionFilterContext context) {
+        HttpURLConnection connection = context.connection.getConnection();
+        if (context.connection.getResponseCode() == 401) {
```
mikerhodes: Access to the underlying HttpURLConnection -- does it allow the filters to do things which put the HttpConnection into a bad state?

mikerhodes: And should HttpConnection do some checks on the connection it ends up with?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,124 @@
...
+            cookie = getCookie(connection.getURL());
+            //don't resend request, failed to get cookie
+            if(cookie != null) {
+                context.replayRequest = true;
+                connection.setRequestProperty("Cookie", cookie);
+
+                context = new HttpConnectionFilterContext(context);
+            } else {
+                context.replayRequest = false;
+                logger.severe("Cookie is unavailable, cannot replay request");
```
mikerhodes: You can definitely improve this with things like server was down, bad creds etc.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: Response filters probably shouldn't be doing anything that would cause exceptions, so I think its fine to bomb out if a runtime exception gets thrown. The author of any implementation should handle exceptions that arise.

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,124 @@
...
+     private String getCookie(URL url){
+
+        try {
+            URL sessionURL = new URL(String.format("%s://%s:%d/_session",
+                    url.getProtocol(),
+                    url.getHost(),
+                    url.getPort()));
+
+            HttpConnection conn = Http.POST(sessionURL, "application/json");
+            conn.setRequestBody(cookieRequestBody.getBytes("UTF-8"));
```
mikerhodes: There's a version of this which takes a String; use it.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: I don't know, @tomblench did you do any work on this when writing the first pass?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,124 @@
...
+            String cookieHeader = conn.execute().getConnection().getHeaderField("Set-Cookie");
```
mikerhodes: Check response is appropriate.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -81,6 +93,16 @@ public HttpConnection(String requestMethod,
...
+        this.requestFilters = new LinkedList<HttpConnectionRequestFilter>();
+        this.responseFilters = new LinkedList<HttpConnectionResponseFilter>();
+    }
+
+    /**
+     * Sets the number of times this request can be retried.
+     * @param numberOfRetries the number of times this request can be retried.
+     */
+    public void setNumberOfRetries(int numberOfRetries){
+        this.numberOfRetries = numberOfRetries;
```
mikerhodes: I'd consider protecting properties which should be set before the request begins, throw a RuntimeException if the developer tries to call them after execute has been called?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -74,6 +77,15 @@
 
     public final HashMap<String, String> requestProperties;
 
+    public final LinkedList<HttpConnectionRequestFilter> requestFilters;
+    public final LinkedList<HttpConnectionResponseFilter> responseFilters;
```
mikerhodes: Consider making these just `List` objects in the interface, and the requestProperties just a `Map`.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -74,6 +77,15 @@
...
+
+    //we need to store the response code separately so we don't send a request
+    //before we are ready
+    private int responseCode = -1;
```
mikerhodes: This still feels messy, and isn't publicly documented that -1 can be returned by getResponseCode.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -133,56 +155,82 @@ public void setRequestBody(InputStream input, long inputLength) {
...
+                }
+                
+                // See "8.2.3 Use of the 100 (Continue) Status" in http://tools.ietf.org/html/rfc2616
+                // Attempting to write to the connection's OutputStream may cause an exception to be
+                // thrown. This is useful because it avoids sending large request bodies (such as
+                // attachments) if the server is going to reject our request. Reasons for rejecting
+                // requests could be 401 Unauthorized (eg cookie needs to be refreshed), etc.
+                connection.setRequestProperty("Expect", "100-continue");
+                
+                int bufSize = 1024;
```
mikerhodes: 1k buffers are quite small?

sync-core/src/main/java/com/cloudant/http/HttpConnectionRequestFilter.java
rhyshort: removed in 2343454

mikerhodes:
Overall I'd like to see a bunch more docs about what filters are able to safely do.

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
mikerhodes: No, our `HttpConnection` class, not `URLConnection`?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: A filter shouldn't return null as it breaks the contract, an NPE will get thrown somewhere

mikerhodes: Document it.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
tomblench: Obviously if the filter read the response content then the original caller would have missed the data. It does raise an interesting point because we should probably have some hooks which allow response filters to modify/transform the content before the caller sees it.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
mikerhodes: Well, info is fine -- is there a way for the caller to know the thing failed because max retries were reached?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: Oh right, chalk that up to Friday afternoon tiredness, Those properties get applied, before we make the calls to the filters, we could move it, and make that the API to set the properties. But I think it would be better if they applied them straight to the `HttpUrlConnection` object. So we could 

1. Apply Modifiable headers eg auth
2. Run Filters
3. Apply unmodifiable headers, eg content type

So I think we have the hash holding properties we don't want to user to modify and hide it. 

And thinking about it even more The cookie filter should remove the auth header since there is no point for that when the cookie is applied.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: Perhaps we should hide the URLConnection Object and only provide api's via our connection object, that way we can improve the hooks in a safe manner without risking the user messing with things they shouldn't be

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: No, not right now, they just know it failed

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: It does, and it shouldn't, I think we should hide the URLConnection object, and only provide APIs through our connection object

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: Sadly there isn't calling for the response code from the URL connection object will cause a request to get sent (if one hasn't been made already), breaking everything else.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: another @tomblench question, I assume there is a reason behind it

sync-core/src/main/java/com/cloudant/http/HttpConnectionResponseFilter.java
rhyshort: Done in  41ccc69

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: done in e2392de

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: done in 8a7af2b

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: done in 8a7af2b

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: done in 275df31

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: fixed in 52e906df443e1439abce991c2d91cd34e0a3b2a2

sync-core/src/main/java/com/cloudant/http/HttpConnectionRequestFilter.java
rhyshort: fixed in 4416868446fcf5537044b8d25022e479d8dfe277

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: I think if we do throw an exception for that it should be a `ConcurrentModificationException`

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: Done in 4416868

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
tomblench: No reason, it's completely arbitrary and I haven't performance tested it.

Difficult to find good benchmark results out there but 8kiB seems to come up a few times, including here: http://nadeausoftware.com/articles/2008/02/java_tip_how_read_files_quickly

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,147 @@
...
+                return cookieHeader.substring(0, cookieHeader.indexOf(";"));
+            } else {
+                // catch any other response code
+                logger.log(Level.SEVERE,
+                        "Failed to get cookie from server, response code %s, " +
+                                "cookie authentication will not be attempted again",
+                        responseCode);
+                shouldAttemptCookieRequest = false;
+
+            }
```
mikerhodes: Spacing and newlines need fixing in this class.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
mikerhodes: I disagree, what's being concurrently modified?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -79,6 +84,16 @@
 
     private static String userAgent = getUserAgent();
 
+    public final List<HttpConnectionRequestFilter> requestFilters;
+    public final List<HttpConnectionResponseFilter> responseFilters;
+
+    //we need to store the response code separately so we don't send a request
+    //before we are ready
+    private int responseCode = -1;
```
mikerhodes: I don't see anywhere in the code which checks the responseCode to avoid sending a request before something is ready.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -163,36 +188,101 @@ public HttpConnection execute() throws IOException {
...
+                this.responseCode = connection.getResponseCode();
+
+                for (HttpConnectionResponseFilter responseFilter : responseFilters) {
+                    currentContext = responseFilter.filterResponse(currentContext);
+                }
+
+                // retry flag is set from the final step in the response filterRequest pipeline
+                retry = currentContext.replayRequest;
+            }
+            if (n < 0) {
```
mikerhodes: Should this be less-than-or-equal, given the loop above bails when `n==0`?

sync-core/src/main/java/com/cloudant/http/HttpConnectionRequestFilter.java
```diff
@@ -0,0 +1,48 @@
...
+public interface HttpConnectionRequestFilter {
+
+    /**
+     * Filter the request.
+     * This method <strong>must not</strong> do any of the following:
+     * <ul>
+     *     <li>Return null</li>
+     *     <li>Call methods on the underlying {@link java.net.HttpURLConnection} which
+     *     initiate a request such as {@link HttpURLConnection#getResponseCode()}</li>
+     * </ul>
```
mikerhodes: I suggest whitelisting safe things to do, that we've checked are okay.

sync-core/src/main/java/com/cloudant/http/HttpConnectionResponseFilter.java
```diff
@@ -0,0 +1,50 @@
...
+ Filter enables two main behaviours:
+<ul>
+  <li>Modifying the response for every request</li>
+
+ <li> Replaying a (potentially modified) request by reacting to the
+ response. For example, obtaining a cookie on receipt of a 401
+ response code, modifying the "Cookie" header of the original
+ request, then setting replayRequest to <code>true</code> to replay the request
+ with the new "Cookie" header, see {@link CookieFilter#filterResponse(HttpConnectionFilterContext)}
+ for an example of replying requests based on the response code.</li>
```
mikerhodes: Why not just link to the CookieFilter?

sync-core/src/test/java/com/cloudant/common/CouchTestBase.java
```diff
@@ -57,7 +57,7 @@
             System.getProperty("test.couch.ignore.compaction",Boolean.FALSE.toString()));
 
     public static final Boolean IGNORE_AUTH_HEADERS = Boolean.valueOf(
-            System.getProperty("test.couch.ignore.auth.headers",Boolean.FALSE.toString()));
+            System.getProperty("test.couch.ignore.auth.headers",Boolean.TRUE.toString()));
```
mikerhodes: Why is this default changed?

sync-core/src/test/java/com/cloudant/http/HttpTest.java
```diff
@@ -119,4 +122,90 @@ public void testReadBeforeExecute() throws IOException {
...
+    //
+    @Test
+    public void testCookieAuthWithoutRetry() throws IOException {
+
+        if(IGNORE_AUTH_HEADERS){
+            System.err.println("Ignore cookie filters");
+            return;
+        }
+
+        CookieFilter filter = new CookieFilter(System.getProperty("test.couch.username"),System.getProperty("test.couch.password"));
```
mikerhodes: line length

sync-core/src/test/java/com/cloudant/http/HttpTest.java
```diff
@@ -119,4 +122,90 @@ public void testReadBeforeExecute() throws IOException {
...
+
+        //check the json
+        JSONHelper helper = new JSONHelper();
+        Map<String,Object> jsonRes = helper.fromJson(new InputStreamReader(conn.getConnection()
+                .getInputStream()));
+
+        Assert.assertTrue(jsonRes.containsKey("ok"));
+        Assert.assertTrue((Boolean)jsonRes.get("ok"));
+        Assert.assertTrue(jsonRes.containsKey("id"));
+        Assert.assertTrue(jsonRes.containsKey("rev"));
```
mikerhodes: I don't think this should succeed because you've not set a request filter. The response filter shouldn't be able to modify future (retried) requests, only respond to or alter the results of the current one.

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,147 @@
...
+
+            if(responseCode == 401){
+                shouldAttemptCookieRequest  = false;
+                logger.severe("Credentials are incorrect, cookie authentication will not be attempted again");
+            } else if (responseCode / 100 == 5){
+                logger.log(Level.SEVERE,
+                        "Failed to get cookie from server, response code %s, cookie auth",
+                        responseCode);
+            } else if(responseCode / 100 == 2) {
+                return cookieHeader.substring(0, cookieHeader.indexOf(";"));
```
mikerhodes: More validation required here imo, you've no idea at this point whether you have a CouchDB response or just some other `200`.

sync-core/src/test/java/com/cloudant/common/CouchTestBase.java
rhyshort: It changed, because the server needs to be set up not in admin party to be able to run the tests.

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: Such as? Any proxying the user is doing should be transparent, other parts of the code base would  likely break if it wasn't.

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: CookieFilter was using it originally when we had a single interface method for filtering requests and responses, will remove

sync-core/src/test/java/com/cloudant/http/HttpTest.java
rhyshort: removed in 45a5727

sync-core/src/test/java/com/cloudant/http/HttpTest.java
rhyshort: fixed in 38d6be4460111b6a57bdc95cd85b2dfce5e200e5

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: fixed in 1cd1fcf5b883c83b590b30336f6838be69819817

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: removed in de5844e6b2b296248ee6b5d656a15ab5564031c1

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: done in 398b495

sync-core/src/main/java/com/cloudant/http/HttpConnectionResponseFilter.java
rhyshort: @tomblench wrote this irrc but I think we should have the filters use documentation here, rather than being a case of the code or documentation for a particular filter implementation being the place to go to. Generally I think where the interface is defined should be the go to place for documentation on the API.

sync-core/src/main/java/com/cloudant/http/HttpConnectionRequestFilter.java
rhyshort: That would be a small list right now, currently modifying request headers and instructing to a request to replay. Do we want to impose a limit on what can be done and become responsible for testing every possible use case for filters?

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: I'd interpret it as it being modified as we are using it, but thinking about it again, I don't think we need to do either, since the execute method takes a copy of the number of retries before executing the loop. So we could make a documentation change to say that the new value is ignored if called after execute has been called

sync-core/src/test/java/com/cloudant/http/HttpTest.java
mikerhodes: Did you actually change the code in the HTTP classes to prevent the behaviour you are relying on here rather than just removing the test (as your comment says?)

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,143 @@
...
+    public HttpConnectionFilterContext filterResponse(HttpConnectionFilterContext context) {
+        HttpURLConnection connection = context.connection.getConnection();
+        try {
+            if (context.connection.getConnection().getResponseCode() == 401) {
+                //we need to get a new cookie
+                cookie = getCookie(connection.getURL());
+                //don't resend request, failed to get cookie
+                if(cookie != null) {
+                    context.replayRequest = true;
+                    context = new HttpConnectionFilterContext(context);
```
mikerhodes: I'm still not clear why this is required?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
mikerhodes: Check the JSON looks like a valid _session response, at least ok=true and username=<the one we sent>.

```
{
  "ok": true,
  "name": "kimstebel",
  "roles": []
}
```

sync-core/src/test/java/com/cloudant/http/HttpTest.java
rhyshort: No, just the test removal. I'm not sure how we can prevent it, but I will look into it

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
mikerhodes: Seems like a doc change is fine to me. "Set numberOfRetries before executing the request"

sync-core/src/main/java/com/cloudant/http/HttpConnectionResponseFilter.java
mikerhodes: Remove the confusing doc for now then.

sync-core/src/test/java/com/cloudant/common/CouchTestBase.java
mikerhodes: I don't think that answers the question without more context. Why did it work before, for example, why have we changed the requirements on the server?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: I think at the time I was under the impression that it was needed, I don't think it is now the code has evolved some more

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: done in 4bcb1c2

sync-core/src/test/java/com/cloudant/http/HttpTest.java
mikerhodes: The test should probably remain, but it should check for failure?

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: Removed in 513aaee

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
rhyshort: done in d9d823c

sync-core/src/main/java/com/cloudant/http/HttpConnectionResponseFilter.java
rhyshort: done in 477e5ae

sync-core/src/test/java/com/cloudant/common/CouchTestBase.java
rhyshort: The previous authentication header tests didn't rely on the server being set up with the correct credentials. For cookie auth for us to be able to test if we correctly use the `_session` API we need to have valid credentials for couchdb to use.

sync-core/src/test/java/com/cloudant/http/HttpTest.java
rhyshort: I'm not sure if that gains us anything, other than checking we didn't execute a filter we didn't add to the list of filters. If you think there is a value from this kind of test, I'll add it back in with the check for a `401` error code

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -142,72 +161,100 @@ public void setRequestBody(InputStream input, long inputLength) {
...
+                }
+
+
+                // always read the result, so we can retrieve the HTTP response code
+                connection.setDoInput(true);
+                connection.setRequestMethod(requestMethod);
+                if (contentType != null) {
+                    connection.setRequestProperty("Content-type", contentType);
+                }
+                if (url.getUserInfo() != null) {
```
tomblench: This block is a copy and paste of the one above

rhyshort: This is probably caused by rebase conflicts, going to fix

rhyshort: fixed in 92207ce

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,188 @@
...
+        return null;
+    }
+
+    private boolean responseLooksOkay(InputStream responseStream){
+        //check the response bod
+        JSONHelper jsonHelper = new JSONHelper();
+        Map<String,Object> jsonResponse = jsonHelper.fromJson(new InputStreamReader(responseStream));
+        boolean responseLooksOkay = true;
+
+        responseLooksOkay = jsonResponse.containsKey("ok") && responseLooksOkay;
```
tomblench: This method looks messy and it's hard to understand what you're looking for in the JSON.

Why don't you let our JSON library do the mapping for you, by creating a class to capture the response, like here? https://github.com/cloudant/sync-android/blob/master/sync-core/src/main/java/com/cloudant/sync/replication/CouchClientWrapper.java#L88

rhyshort: Maybe, although if we are going to show this as a way of implementing features, shouldn't we use the APIs we expose to the users? I'm reluctant in using CouchClientWrapper etc in the shining example due to its lack of documentation. The current way of doing it isn't perfect, but its better than using API's we haven't documented, in my opinion anyway.

rhyshort: Thinking about it, I think I missunderstood what you meant. So I use the Jackson to stuff the response into an object which has all the fields already?

tomblench: I mean you can do something like this:
```java
MyCookieResponse response = jsonHelper.fromJson(new InputStreamReader(responseStream, MyCookieResponse.class));
```

rhyshort: In light of https://issues.apache.org/jira/browse/COUCHDB-1356 I'm going to leave it how it is, since we are only going to check for the okay field.

sync-core/src/main/java/com/cloudant/http/HttpConnectionFilterContext.java
rhyshort: Copyrights added in 75346b6

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,168 @@
...
+            HttpConnection conn = Http.POST(sessionURL, "application/x-www-form-urlencoded");
+            conn.setRequestBody(sessionRequestBody);
+            HttpURLConnection connection = conn.execute().getConnection();
+            String cookieHeader = connection.getHeaderField("Set-Cookie");
+            int responseCode = connection.getResponseCode();
+
+            if(responseCode == 401){
+                shouldAttemptCookieRequest  = false;
+                logger.severe("Credentials are incorrect, cookie authentication will not be" +
+                        " attempted again");
```
mikerhodes: Suggest for accuracy: "will not be attempted again by this filter object"

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,168 @@
...
+                logger.log(Level.SEVERE,
+                        "Failed to get cookie from server, response code %s, cookie auth",
+                        responseCode);
+            } else if(responseCode / 100 == 2) {
+
+                if(responseLooksOkay(connection.getInputStream())) {
+                    return cookieHeader.substring(0, cookieHeader.indexOf(";"));
+                } else {
+                    return null;
+                }
```
mikerhodes: Suggest starting with the success case, particularly as it's got the lowest response code.

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,168 @@
...
+            logger.log(Level.SEVERE,"Failed to create URL for _session endpoint",e);
+        } catch (UnsupportedEncodingException e) {
+            logger.log(Level.SEVERE, "Failed to encode cookieRequest body", e);
+        } catch (IOException e) {
+            logger.log(Level.SEVERE, "Failed to read cookie response header", e);
+        }
+        return null;
+    }
+
+    private boolean responseLooksOkay(InputStream responseStream){
```
mikerhodes: Looks okay for what? Cleaning sinks? :P Name the method for what it does -- e.g., `loggedIn`.

sync-core/src/main/java/com/cloudant/http/HttpConnectionResponseFilter.java
mikerhodes: See above still.

sync-core/src/test/java/com/cloudant/http/HttpTest.java
mikerhodes: Okay, is fine now I think.

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: done in 8e2a46c

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: done in 8e2a46c

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
rhyshort: renamed to `sessionHasStarted` in 8bb4d99

sync-core/src/main/java/com/cloudant/http/HttpConnectionResponseFilter.java
rhyshort: removed in a151189

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,168 @@
...
+    private boolean sessionHasStarted(InputStream responseStream){
+        //check the response body
+        JSONHelper jsonHelper = new JSONHelper();
+        Map<String,Object> jsonResponse = jsonHelper.fromJson(new InputStreamReader(responseStream));
+        boolean responseLooksOkay = true;
+
+
+        // only check for ok:true, https://issues.apache.org/jira/browse/COUCHDB-1356
+        // means we cannot check that the name returned is the one we sent.
+        responseLooksOkay = jsonResponse.containsKey("ok") && responseLooksOkay;
```
tomblench: You can write this more succinctly because `&&` and is short-circuting:

```java
boolean responseLooksOkay = jsonResponse.containsKey("ok") && (Boolean) jsonResponse.get("ok");
```

rhyshort: done in 8d31d6f

mikerhodes:
+1. We can fix up any niggles as we work out the public API for this feature.

sync-core/src/main/java/com/cloudant/http/CookieFilter.java
```diff
@@ -0,0 +1,159 @@
...
+    private boolean shouldAttemptCookieRequest = true;
+    private final String username;
+
+    /**
+     * Constructs a cookie filter.
+     * @param username The username to use when getting the cookie
+     * @param password The password to use when getting the cookie
+     */
+    public CookieFilter(String username, String password){
+        this.sessionRequestBody = String.format(Locale.ENGLISH,"name=%s&password=%s",username,password);
```
tomblench: `Locale.ENGLISH` not needed here as `%s` is not locale specific

rhyshort: removed in 4e317df

sync-core/src/test/java/com/cloudant/http/HttpTest.java
```diff
@@ -119,4 +122,52 @@ public void testReadBeforeExecute() throws IOException {
         Assert.assertEquals(bis.available(), data.getBytes().length);
         client.deleteDb();
     }
+
+
+    //NOTE: This test doesn't work with specified couch servers,
```
tomblench: Does this mean that this test case will sometimes fail depending on configuration? I'm confused as to when this should and shouldn't run

sync-core/src/test/java/com/cloudant/http/HttpTest.java
```diff
@@ -119,4 +122,52 @@ public void testReadBeforeExecute() throws IOException {
...
+    // the URL will always include the creds specified for the test
+    //
+    // A couchdb server needs to be set and running with the correct
+    // security settings, the database *must* not be public, it *must*
+    // be named cookie_test
+    //
+    @Test
+    public void testCookieAuthWithoutRetry() throws IOException {
```
tomblench: Does this need an annotation to put it into the appropriate category? `RequireRunningCouchDb`?

rhyshort: No, because the class is already annotated with `RequireRunningCouchDB`

sync-core/src/test/java/com/cloudant/http/HttpTest.java
rhyshort: Due to the way the specified couch servers inject credentials, i.e.  through the URI, the tests will always pass even if the cookie filter failed to get a cookie.

tomblench: So it's pointless running it. Can you think of a way of skipping it if it isn't relevant, using one of the `Assume` methods?

sync-core/src/test/java/com/cloudant/http/HttpTest.java
tomblench: Stupid me, that's what happens when you only look at the diff and not the whole file.

sync-core/src/test/java/com/cloudant/http/HttpTest.java
rhyshort: Yeah, there already was a way of skipping, but I've updated it to use `Assume.false` in 9b86a41

tomblench:
OK +1 :panda_face:

PR #140, opened 2015-05-22
47252 public keyprovider

mikerhodes:
This PR simply exposes a Datastore's key provider to the outside world. This is primarily intended so that the Query index code and the Attachments code can get access to the key provider.

reviewer @emlaver 
reviewer @tomblench


tomblench:
Looks reasonable to me, but I have a style question. I seem to recall you had a campaign to abolish 'simple getters' in favour of making the class member `public final`, so which is appropriate?

It seems we have a mixture in `BasicDatastore`, does it depend on whether we expect the getter to perform additional logic/preconditions?

mikerhodes:
Here, I think we need to use the getter approach because of exposing the attribute via the `Datastore` interface.

tomblench:
:+1:

emlaver:
I agree with using the getter approach.  +1

PR #143, opened 2015-05-27
Correctly persist and send length & encoded_length values

tomblench:
*What* persist to the local DB the `length` and `encoded_length` metadata properties when they are sent with attachments.
*Why* currently this data is not persisted or used to check that the actual length equals the indicated length.
*How* update various Attachment classes and related methods to take `length` and `encoded_length` arguments and assert that actual attachment length equals signalled attachment length
*Testing* some covered by existing tests, more TODO

reviewer @rhyshort 
reviewer @mikerhodes


sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -170,6 +171,8 @@ protected PreparedAndSavedAttachments prepareAttachments(Collection<? extends At
 
         for (Attachment a : attachments) {
             if (!(a instanceof SavedAttachment)) {
+                // TODO FIXME the attachment will be an unsaved stream/file attachment
+                // so we potentially need to think about what the different lengths will be before/after zipping, amongst other things
```
rhyshort: Is this going to be handled in this PR or a later PR?

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -58,5 +107,8 @@ public PreparedAttachment(Attachment attachment,
     public final Attachment attachment;
     public final File tempFile;
     public final byte[] sha1;
+    public final long length;
+    public final long encodedLength;
+
```
rhyshort: Can we get these properties pushed to the top of the file?

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -32,6 +32,48 @@
...
+        this.encodedLength = encodedLength;
+        FileInputStream tempFileIS = null;
+        try {
+            FileUtils.copyInputStreamToFile(attachment.getInputStream(), tempFile);
+            long expectedLength;
+            if (this.attachment.encoding == Attachment.Encoding.Plain) {
+                expectedLength = length;
+            } else {
+                expectedLength = encodedLength;
+            }
```
mikerhodes: Shouldn't the replicator code be making this check? I'm not sure PreparedAttachment needs to check this, it's not its job to validate stuff.

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -46,6 +88,13 @@ public PreparedAttachment(Attachment attachment,
         FileInputStream tempFileIS = null;
         try {
             FileUtils.copyInputStreamToFile(attachment.getInputStream(), tempFile);
+            if (this.attachment.encoding == Attachment.Encoding.Plain) {
+                this.length = tempFile.length();
+                this.encodedLength = 0;
```
mikerhodes: Isn't this also tempfile.length()?

sync-core/src/main/java/com/cloudant/sync/datastore/SavedAttachment.java
```diff
@@ -68,14 +78,10 @@ public boolean shouldInline(PushAttachmentsInline inlinePreference) {
...
 
-    public long getSize() {
-        return file.length();
-    }
-
     protected final long revpos;
     protected final long seq;
     protected final byte[] key;  // sha of file, used for file path on disk.
+    protected final long length;
+    protected final long encodedLength;
```
rhyshort: Can we get these properties pushed to the top of the file?

mikerhodes:
Can this commit be split up? It does a whole bunch of stuff. At least the validation logic for incoming replicated attachments can live in a separate commit, but really there could be at least one each for migrating SavedAttachment and PreparedAttachment for using the length from the database and from the replicator respectively.

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -213,13 +240,9 @@ protected Attachment getAttachment(SQLDatabase db, BasicDocumentRevision rev, St
...
             if (c.moveToFirst()) {
-                int sequence = c.getInt(0);
-                byte[] key = c.getBlob(2);
-                String type = c.getString(3);
-                int encoding = c.getInt(4);
-                int revpos = c.getInt(7);
+                byte[] key = c.getBlob(c.getColumnIndex("key"));
                 File file = fileFromKey(key);
-                return new SavedAttachment(attachmentName, revpos, sequence, key, type, file, Attachment.Encoding.values()[encoding]);
+                return new SavedAttachment(file, c);
```
mikerhodes: I'd leave the reading of the values from the database in AttachmentManager, rather than spread database knowledge around. Particularly as you read some values from the cursor here and some in SavedAttachment.

sync-core/src/main/java/com/cloudant/sync/datastore/RevisionHistoryHelper.java
```diff
@@ -195,7 +195,8 @@ public static MultipartAttachmentWriter createMultipartWriter(BasicDocumentRevis
                         mpw = new MultipartAttachmentWriter();
                         mpw.setBody(revision);
                     }
-                    mpw.addAttachment(att);
+                    mpw.addAttachment(savedAtt,
+                            savedAtt.encoding == Attachment.Encoding.Plain ? savedAtt.length : savedAtt.encodedLength);
```
mikerhodes: We have similar logic to this in several places. Could we localise into methods like "dataLength" and "onDiskLength" somewhere?

tomblench: I'm glad you mentioned this because I had similar thoughts. I was trying to keep as closely as possible to the couch metadata model but this nasty pattern arises. As suggested I think I'll add a method to SavedAttachment called "onDiskLength" because that's as literal and clear as you can get.

sync-core/src/main/java/com/cloudant/sync/datastore/SavedAttachment.java
```diff
@@ -54,7 +66,7 @@ public InputStream getInputStream() throws IOException {
     }
 
     public boolean isLarge() {
-        return this.getSize() > largeSizeBytes;
+        return this.encodedLength > largeSizeBytes;
```
mikerhodes: Isn't encodedLength zero if the thing isn't encoded?

PR #144, opened 2015-05-27
Read file manually in PreparedAttachment while calculating SHA1 and length

emlaver:
*What*
Change the current method of saving the temporary file, length, and calculating the SHA1 in PreparedAttachment.
*Why* 
To prepare for encrypting attachments, we need the stream to be read once and the SHA1 digest calculated as we go.
*How*
Using a manual method that includes a while loop that reads the input stream and writes to the temporary file using an output stream. Within the loop, we calculate the SHA1 and the total length.
*Testing* 
setAndGetAttachmentsTest includes a test for asserting the expected SHA1 and saved attachment SHA1.  Also added in the same method a test to compare file length and saved attachment length.

reviewer @mikerhodes 
reviewer @tomblench 

BugId: 47455

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
+++ b/sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2014 Cloudant, Inc. All rights reserved.
+ * Copyright (c) 2015 Cloudant, Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
@@ -14,15 +14,18 @@
 
 package com.cloudant.sync.datastore;
 
-import com.cloudant.sync.util.Misc;
-
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.UUID;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * An attachment which has been been copied to a temporary location and had its sha1 calculated,
@@ -32,6 +35,7 @@ import java.util.UUID;
  */
 public class PreparedAttachment {
 
+    private Logger logger = Logger.getLogger(PreparedAttachment.class.getCanonicalName());
     /**
      * Prepare an attachment by copying it to a temp location and calculating its sha1.
      *
@@ -43,20 +47,46 @@ public class PreparedAttachment {
                               String attachmentsDir) throws AttachmentException {
         this.attachment = attachment;
         this.tempFile = new File(attachmentsDir, "temp" + UUID.randomUUID());
-        FileInputStream tempFileIS = null;
+        InputStream attachmentInStream = null;
+        OutputStream tempFileOutStream = null;
+        MessageDigest calculateSha1 = null;
+        int totalRead = 0;
         try {
-            FileUtils.copyInputStreamToFile(attachment.getInputStream(), tempFile);
-            this.sha1 = Misc.getSha1((tempFileIS = new FileInputStream(tempFile)));
+            attachmentInStream = attachment.getInputStream();
+
+            //Use FileUtils to create folder structure and file if necessary for output stream
+            tempFileOutStream = FileUtils.openOutputStream(this.tempFile);
+
+            calculateSha1 = MessageDigest.getInstance("SHA-1");
+            int bufferSize = 1024;
+            byte[] buffer = new byte[bufferSize];
+            int bytesRead;
+            while ((bytesRead = attachmentInStream.read(buffer)) != -1) {
+                calculateSha1.update(buffer, 0, bytesRead);
+                tempFileOutStream.write(buffer, 0, bytesRead);
+                totalRead += bytesRead;
+            }
         } catch (IOException e) {
+            logger.log(Level.WARNING, "Problem reading from input or writing to output stream ", e);
+            throw new AttachmentNotSavedException(e);
+        } catch (NoSuchAlgorithmException e) {
+            logger.log(Level.WARNING, "Problem calculating SHA1 for attachment stream ", e);
             throw new AttachmentNotSavedException(e);
         } finally {
-            //ensure the temp file is closed after calculating the hash
-            IOUtils.closeQuietly(tempFileIS);
+            //Ensure the attachment input stream and file output stream is closed after calculating the hash
+            IOUtils.closeQuietly(attachmentInStream);
+            IOUtils.closeQuietly(tempFileOutStream);
         }
+        //Set attachment length from bytes read in input stream
+        this.length = totalRead;
+        this.sha1 = calculateSha1.digest();
     }
 
     public final Attachment attachment;
     public final File tempFile;
     public final byte[] sha1;
+    public final long length;
 }
+
+
 
--- a/sync-core/src/test/java/com/cloudant/sync/datastore/AttachmentTest.java
+++ b/sync-core/src/test/java/com/cloudant/sync/datastore/AttachmentTest.java
@@ -20,8 +20,6 @@ import com.cloudant.sync.sqlite.SQLQueueCallable;
 import com.cloudant.sync.util.Misc;
 import com.cloudant.sync.util.TestUtils;
 
-import static org.hamcrest.CoreMatchers.is;
-
 import org.apache.commons.io.IOUtils;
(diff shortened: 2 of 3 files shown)


sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -43,20 +47,47 @@ public PreparedAttachment(Attachment attachment,
...
+            int bytesRead;
+            int totalRead = 0;
+            while ((bytesRead = attachmentInStream.read(buf)) != -1) {
+                sha1.update(buf, 0, bytesRead);
+                tempFileOutStream.write(buf, 0, bytesRead);
+                totalRead += bytesRead;
+            }
+
+            //Set attachment length from bytes read in input stream
+            this.length = totalRead;
```
mikerhodes: Probably set this down by where the sha1 is set for the object, for consistency.

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -43,20 +47,47 @@ public PreparedAttachment(Attachment attachment,
...
         try {
-            FileUtils.copyInputStreamToFile(attachment.getInputStream(), tempFile);
-            this.sha1 = Misc.getSha1((tempFileIS = new FileInputStream(tempFile)));
+            attachmentInStream = attachment.getInputStream();
+
+            //Use FileUtils to create folder structure and file if necessary for output stream
+            tempFileOutStream = FileUtils.openOutputStream(this.tempFile);
+
+            sha1 = MessageDigest.getInstance("SHA-1");
+            int bufSiz = 1024;
```
mikerhodes: Don't abbreviate when it saves you just a character or two.

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -43,20 +47,47 @@ public PreparedAttachment(Attachment attachment,
...
+            byte buf[] = new byte[bufSiz];
```
mikerhodes: Shouldn't this be `byte[] buf`?

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -43,20 +47,47 @@ public PreparedAttachment(Attachment attachment,
...
+                totalRead += bytesRead;
+            }
+
+            //Set attachment length from bytes read in input stream
+            this.length = totalRead;
         } catch (IOException e) {
+            logger.log(Level.WARNING,"Problem reading input or output stream ",e);
+            throw new AttachmentNotSavedException(e);
+        } catch (NoSuchAlgorithmException e) {
+            logger.log(Level.WARNING,"Problem calculating SHA1 for attachment stream ",e);
```
mikerhodes: Spaces after commas.

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -43,20 +47,47 @@ public PreparedAttachment(Attachment attachment,
                               String attachmentsDir) throws AttachmentException {
         this.attachment = attachment;
         this.tempFile = new File(attachmentsDir, "temp" + UUID.randomUUID());
-        FileInputStream tempFileIS = null;
+        InputStream attachmentInStream = null;
+        OutputStream tempFileOutStream = null;
+        MessageDigest sha1 = null;
```
mikerhodes: Don't shadow the instance variable if possible.

sync-core/src/test/java/com/cloudant/sync/datastore/AttachmentTest.java
```diff
@@ -66,6 +62,11 @@ public void setAndGetAttachmentsTest() throws Exception {
             SavedAttachment savedAtt2 = (SavedAttachment) datastore.attachmentsForRevision
                     (newRevision).get(0);
             Assert.assertArrayEquals(expectedSha1, savedAtt2.key);
+
+            //Assert file length and saved attachment length
+            long expectedLength = f.length();
+            Assert.assertEquals(expectedLength,savedAtt.getSize());
+            Assert.assertEquals(expectedLength,savedAtt2.getSize());
```
mikerhodes: I'm not sure what this is meant to test?

emlaver: This is to confirm that the original file (variable f) is the same length as the saved attachment file length.  In the addAttachment method in AttachmentManager, the length from total bytes in PreparedAttachment is put into the ContentValue variable.  SavedAttachment retrieves this length when you call the getSize method. 
This test can also be useful when we add encryption to attachments and need to check that the decrypted length is the same as the original file length.

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
emlaver: Fixed in db71913691e9202f5420a3af5cae52f530b05fc0

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
emlaver: Fixed in db71913691e9202f5420a3af5cae52f530b05fc0

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
emlaver: Can be both ways, but `byte[] buffer` is less confusing and the recommended way.  Fixed: db71913691e9202f5420a3af5cae52f530b05fc0

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
emlaver: Fixed db71913691e9202f5420a3af5cae52f530b05fc0

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
emlaver: Fixed db71913691e9202f5420a3af5cae52f530b05fc0

sync-core/src/test/java/com/cloudant/sync/datastore/AttachmentTest.java
mikerhodes: This doesn't test just your new code, and particularly with the old SavedAttachment code, the length returned by getSize() is for the file on disk anyway.

If you want to check your PreparedAttachment code calculates the right length, instantiate a prepared attachment with an UnsavedAttachment and check the file length of the prepared attachment.

```java
File f = TestUtils.loadFixture("fixture/"+attachmentName);
long expectedLength = f.length();
Attachment att = new UnsavedFileAttachment(f, "text/plain");
PreparedAttachment patt = new PreparedAttachment(att, /* blah */);
Assert.assertEquals(expectedLength, patt.getSize());
```

With a similar implementation, you should add a test that you are calculating the SHA1 correctly.

emlaver: Added test cases in eac97bf8469ddacf68ef93837a45603a67a0313f and 4bacefb14cbfcaaf30e6db9faf57f2f63d6d2e5e

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -43,20 +47,46 @@ public PreparedAttachment(Attachment attachment,
...
+            int bufferSize = 1024;
+            byte[] buffer = new byte[bufferSize];
+            int bytesRead;
+            while ((bytesRead = attachmentInStream.read(buffer)) != -1) {
+                calculateSha1.update(buffer, 0, bytesRead);
+                tempFileOutStream.write(buffer, 0, bytesRead);
+                totalRead += bytesRead;
+            }
         } catch (IOException e) {
+            logger.log(Level.WARNING, "Problem reading input or output stream ", e);
```
tomblench: "Problem reading from input or writing to output stream"

tomblench:
+1 looks good to me. I have some changes I will need to merge in to `PreparedAttachment` when #143 goes in so I'd like this PR to merge soon.

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
emlaver: Fixed in cd31ebee01e255182e27fd7e682e87f63908754d

emlaver:
@tomblench Changed temporary directory in 18693827f5e845b4051f6cb785667239f657ba19

mikerhodes:
+1. My guess is this can all be fixed-up into the first commit.

tomblench:
OK +1

PR #145, opened 2015-05-27
47497 encrypting streams

mikerhodes:
_What:_
Two Stream implementations, one which reads the encrypted attachment blob format and one which writes it.

_How:_
Simple wrappers around the JCE CipherInput/OutputStream classes which read or write the header before delegating to the CipherXStream to write the content.

_Tests:_
Added tests that the streams can read and write pre-canned data, checked against known-good versions checked against the same encryption as done by OpenSSL.
Also add checks for error conditions on reading.

_Questions:_
- Is extending `FilterInputStream` and `FilterOutputStream` the right thing to do here?
- I think reading/writing from the wrapped stream during construction is the right thing to do, but thoughts welcome.

Don't worry about the 4,719 lines of Dostoevsky, it's the encryption test file, taken from Project Gutenburg to avoid having a trivial test file. GitHub collapses it anyway, thankfully.

reviewer @emlaver 
reviewer @ricellis


sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptedAttachmentInputStream.java
```diff
@@ -0,0 +1,259 @@
...
+    public int available()
+            throws IOException {
+        return cipherInputStream.available();
+    }
+
+    @Override
+    /**
+     * <p>Closes this input stream and releases any system resources associated with the stream.</p>
+     *
+     * <p>The EncryptedAttachmentInputStream method of CipherInputStream calls the close method
```
ricellis: I find this comment pretty confusing. Wouldn't it be easier just to say this method calls close on its proxied CipherInputStream?

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptedAttachmentInputStream.java
```diff
@@ -0,0 +1,259 @@
...
+            throws IOException {
+        cipherInputStream.close();
+    }
+
+    @Override
+    /**
+     * <p>Tests if this input stream supports the mark and reset methods, which it does not.</p>
+     *
+     * @return false, since this class does not support the mark and reset methods.
+     */
```
ricellis: This is not what the method does.

ricellis:
Assuming that the Cipher*Streams use the same semantics as the ancestors is there any reason not to use inheritDoc in the Encryption* streams with some augmentation to avoid duplicating so much javadoc? e.g.
```java
/**
* {@inheritDoc}
* <p>This method proxies to the CipherInputStream.</p>
*/
@Override
```

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptedAttachmentOutputStream.java
```diff
@@ -0,0 +1,196 @@
...
+     * stream.</p>
+     *
+     * <p>This method invokes the doFinal method of the encapsulated cipher object, which
+     * causes any bytes buffered by the encapsulated cipher to be processed. The result is
+     * written out by calling the flush method of this output stream.</p>
+     *
+     * <p>This method resets the encapsulated cipher object to its initial state and calls
+     * the close method of the underlying output stream.</p>
+     *
+     * <p>This just proxies to the underlying CipherOutputStream.</p>
```
ricellis: I feel like this would make more sense if the comment order was reversed e.g. "This method proxies to the underlying CipherOutputStream which in turn invokes the doFinal..."
Or maybe after declaring that it proxies to CipherOutputStream just use `@see CipherOutputStream#close()`

ricellis:
The keyless forms of the constructors are documented as essentially useless. If that is really the case is there any reason to have those constructors at all?

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptedAttachmentOutputStream.java
```diff
@@ -0,0 +1,196 @@
...
+        super(out);
+        cipherOutputStream = new CipherOutputStream(out, new NullCipher());
+    }
+
+    /**
+     * <p>Creates an output stream with a key and IV.</p>
+     *
+     * <p>This constructor writes bytes to the out parameter in order to write the header.</p>
+     *
+     * <p>Note: if the specified output stream is null, a NullPointerException may be thrown
```
ricellis: This seems a bit strange, if a null stream isn't acceptable input why not check the parameter and throw during construction? Similarly for the InputStream version.

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptedAttachmentInputStream.java
mikerhodes: I think that was probably what I intended to say, but it somehow slipped betwixt mind and fingers.

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptedAttachmentInputStream.java
mikerhodes: I guess I could instead say it proxies down to cipherInputStream, this javadoc was copied from there (like most of the docs).

ricellis:
Re: Is extending FilterInputStream and FilterOutputStream the right thing to do here?
Currently it doesn't seem to be providing any value as all the methods have been overridden anyway. Constructing the FilterXStream using the CipherXStream would avoid all that overriding, but would necessitate moving the constructor code into some sort of static utility method to init the CipherXStreams. It sounds like your concern with that approach is exposure to later superclass changes. I've no idea how likely that is, but it seems that changes to the Input/OutputStream classes themselves would probably necessitate changes to the CipherXStreams anyway.

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptedAttachmentOutputStream.java
mikerhodes: In fact, we do, when we try to read/write. I'll correct this in both streams.

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptedAttachmentInputStream.java
mikerhodes: Addressed in 538ca2ad1325292bb61d38bced840e117fc8fbd6.

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptedAttachmentInputStream.java
mikerhodes: Addressed in 538ca2ad1325292bb61d38bced840e117fc8fbd6.

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptedAttachmentOutputStream.java
mikerhodes: Addressed in 538ca2ad1325292bb61d38bced840e117fc8fbd6.

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/EncryptedAttachmentOutputStream.java
mikerhodes: Addressed in 538ca2ad1325292bb61d38bced840e117fc8fbd6.

mikerhodes:
@ricellis 

> Use `@inheritDoc`

Not sure this would work as we don't inherit from CipherInputStream

> Keyless constuctors

I include those so we can initialise a NullCipher() object so that the constructed object is somehow valid.

> Extending Filter stream

I think it's still useful as a marker -- some transformation of your input happens here. I think this is why CipherInputStream extends too even though it overrides all the methods. I'm not sure moving to helper methods for the CipherXStream init increases the clarity of the code, I'm not sure where they'd belong.

Possibly in the to-be-written AttachmentStreamFactory class -- I guess I'll revisit this design then.

emlaver:
+1, looks good after the recent doc. updates commit

ricellis:
As discussed I think revisiting the design later to potentially avoid the extra wrapping would be good, but for this right now
+1

PR #151, opened 2015-06-02
Check that pulled attachment is in correct format

tomblench:
Mini-PR adds asserts to pullRevisionsWithBigAttachments to ensure that gzipped attachments are retrieved and stored correctly


mikerhodes:
+1

PR #153, opened 2015-06-03
47760 add attachments key filename

mikerhodes:
## What

This PR adds a table to the database which maps attachment key to attachment file name. It also includes a migration to add the new table and migrate data, from user_version 6 to user_version 7.

This schema change is in preparation for a change to fileFromKey to use this table, which will come in a future PR.

## Change in the way updateSchema works

This migration required Java code in addition to SQL statements, so I changed the way updateSchema works to take a Migration object. This object contains the migration logic, but doesn't handle changing the user version. 

I changed older schema updates to use the Migration interface with the help of a new class RunSchemaOnlyMigration which takes the String[] and runs each statement -- as the old updateSchema method did.

reviewer @tomblench 
reviewer @rhyshort


sync-core/src/main/java/com/cloudant/sync/datastore/migrations/RunSchemaOnlyMigration.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.cloudant.sync.datastore.migrations;
+
+import com.cloudant.sync.sqlite.SQLDatabase;
+import com.cloudant.sync.sqlite.SQLQueueCallable;
+
+/**
+ * Runs a database migration which consists only of SQL statements.
+ */
+public class RunSchemaOnlyMigration implements Migration {
```
tomblench: For consistency should this be called `SchemaOnlyMigration`?

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
```diff
@@ -157,30 +158,7 @@ private static boolean validateOpenedDatabase(SQLDatabase db) {
...
-     *
-     * @param database
-     * @param schema
-     * @param version
-     * @throws SQLException
-     *
-     * @see com.cloudant.sync.sqlite.SQLDatabase#getVersion()
-     */
-    public static void updateSchema(SQLDatabase database, String[] schema, int version)
+    public static void updateSchema(SQLDatabase database, Migration migration, int version)
```
tomblench: javadoc?

sync-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase6To7.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Migration from datastore JSON database version 6 to 7, which requires migration of data
+ * between tables with a transform in Java code along the way.
+ */
+public class MigrateDatabase6To7 implements Migration {
+
+    public void runMigration(SQLDatabase db) throws Exception {
+
+        db.execSQL("CREATE TABLE attachments_key_filename ( " +
```
tomblench: Maybe these should be in `DatastoreConstants` - I find it helpful to see all of the schema/DDL in one file

rhyshort: I agree, they should be in `DatabaseConstants` makes it a lot easier to follow the history of database schema

sync-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase6To7.java
```diff
@@ -0,0 +1,52 @@
+/**
+ * Copyright (c) 2015 IBM Cloudant, Inc. All rights reserved.
```
rhyshort: With the copyrights I'm pretty sure we can't use `IBM Cloudant, Inc.` since IBM Cloudant isn't incorporated

sync-core/src/main/java/com/cloudant/sync/datastore/migrations/RunSchemaOnlyMigration.java
mikerhodes: I thought that too, so I changed it 2955483917d5698afef98fe87949e5cf67b030a3.

sync-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
mikerhodes: Fixed in 8bda1601e9844f9930dfd48bd4f3b80d4c097fb8, where I also re-add the exception thrown if the migration fails.

sync-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase6To7.java
mikerhodes: It really depends. If I did that, the DatastoreConstants would appear to contain the migration because it had the schema, but in fact it wouldn't have the full story of what happened to the database.

But happy to move if this argument doesn't sway you.

sync-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase6To7.java
mikerhodes: Fixed for new files in 4511be406cfce2d7c6aa01a648496d64435bbe2b.

rhyshort:
+1

sync-core/src/test/java/com/cloudant/sync/datastore/DatastoreSchemaTests.java
```diff
@@ -46,4 +66,174 @@ public void testSchemasSame() {
...
+    @Test
+    /**
+     * Test the specific migration of a version 6 database to a version 7 database.
+     * This migration should add the attachments_key_filename table, and populate
+     * it with the existing keys from the attachments table as both key and filename.
+     *
+     * fixture/datastores-user_version6.zip contains a known database, at version 6 and
+     * with two attachments.
+     */
+    public void migrateVersion6To7() throws DatastoreNotCreatedException, ExecutionException,
```
tomblench: It doesn't affect the logic of the test but for consistency you should change 7 to 100 in a few places here

sync-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase6To7.java
tomblench: I see your point but I still would find it easier to get an idea of the whole schema by just looking at one file.

tomblench:
+1 :+1: :ghost:

sync-core/src/test/java/com/cloudant/sync/datastore/DatastoreSchemaTests.java
mikerhodes: Yeah, I noticed this too. Will try to catch them.

mikerhodes:
Going to squash.

sync-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase6To7.java
mikerhodes: Right now that'd entail making DatastoreConstants public, so I'd rather not go there. I think this PR changes enough without reworking the visibility and/or location of a bunch of stuff.

PR #154, opened 2015-06-04
Fix Issues in SavedHTTPAttachment

rhyshort:
## What

Correct SavedHTTPAttachment's expectations of what is present in a JSON object for attachments from couch.

## How
Instead of the code path being based on `stub` being present, it is now based on the presence of `data`. `length` is correctly worked out based on the information provided in the `data` field if present or by
using the `length` field of the attachment objects.

reviewer @tomblench 
reviewer @brynh

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/datastore/SavedHttpAttachment.java
+++ b/sync-core/src/main/java/com/cloudant/sync/datastore/SavedHttpAttachment.java
@@ -34,9 +34,8 @@ public class SavedHttpAttachment extends Attachment {
 
 
     private URI attachmentURI;
-    private int size;
+    private long size;
     private byte[] data;
-    private Encoding encoding;
 
 
     /**
@@ -49,23 +48,26 @@ public class SavedHttpAttachment extends Attachment {
      */
     public SavedHttpAttachment(String name, Map<String,Object> attachmentData, URI attachmentURI)
            throws IOException {
-         super(name, (String)attachmentData.get("content_type"), Encoding.Plain);
-         Boolean stub = (Boolean) attachmentData.get("stub");
-         Number length = (Number)attachmentData.get("length");
+         super(name,
+                 (String)attachmentData.get("content_type"),
+                 Attachment.getEncodingFromString((String)attachmentData.get("encoding")));
+
+         Number length;
          String data = (String)attachmentData.get("data");
-         String encoding =  (String)attachmentData.get("encoding");
-         this.encoding = Attachment.getEncodingFromString(encoding);
-         if(!stub){
+         if(data != null){
             byte[] dataArray =  data.getBytes();
             InputStream is = Base64InputStreamFactory.get(new ByteArrayInputStream(dataArray));
             this.data = IOUtils.toByteArray(is);
+            length = this.data.length;
+         } else {
+            length = (Number)attachmentData.get("length");
          }
 
          this.attachmentURI = attachmentURI;
-         this.size = length.intValue();
-
+         this.size = length.longValue();
 
     }
+
     @Override
     public long getSize() {
         return size;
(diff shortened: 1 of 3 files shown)


sync-core/src/main/java/com/cloudant/sync/datastore/SavedHttpAttachment.java
```diff
@@ -36,7 +36,6 @@
     private URI attachmentURI;
     private int size;
     private byte[] data;
-    private Encoding encoding;
```
tomblench: We should still store `encoding` (and `encoded_length`) if they are available

rhyshort: encoding is stored in the Attachment super class, is encoded length worth storing? I don't think its useful since there isn't an API to ask for it

tomblench: Didn't notice `encoding`, fair enough.

After #143 merges then `length` and `encoded_length` will be stored as `public final` values in the appropriate `Attachment` classes and `getLength()` will be dropped from the `Attachment` interface. So I guess I can revisit this class then.

rhyshort: Okay, I'll open a FB ticket for it so we remember to do it when #143 merges into master

mikerhodes:
Typos in original commit comment, would also be good if you explicitly spelled out what we were mishandling previously in the comment.

sync-core/src/main/java/com/cloudant/sync/datastore/SavedHttpAttachment.java
tomblench: For reference it's https://cloudant.fogbugz.com/f/cases/47818

sync-core/src/main/java/com/cloudant/sync/datastore/SavedHttpAttachment.java
```diff
@@ -49,23 +48,26 @@
...
             byte[] dataArray =  data.getBytes();
             InputStream is = Base64InputStreamFactory.get(new ByteArrayInputStream(dataArray));
             this.data = IOUtils.toByteArray(is);
+            length = this.data.length;
+         } else {
+             length = (Number)attachmentData.get("length");
          }
 
          this.attachmentURI = attachmentURI;
          this.size = length.intValue();
```
tomblench: Is it possible to declare `size` as `long` and use `longValue()`? We store sizes as `long`s in most places although I appreciate users with >4GB attachments are probably asking for trouble.

sync-core/src/main/java/com/cloudant/sync/datastore/SavedHttpAttachment.java
```diff
@@ -49,23 +48,26 @@
...
-         String encoding =  (String)attachmentData.get("encoding");
-         this.encoding = Attachment.getEncodingFromString(encoding);
-         if(!stub){
+         if(data != null){
             byte[] dataArray =  data.getBytes();
             InputStream is = Base64InputStreamFactory.get(new ByteArrayInputStream(dataArray));
             this.data = IOUtils.toByteArray(is);
+            length = this.data.length;
+         } else {
+             length = (Number)attachmentData.get("length");
```
tomblench: Odd indenting

tomblench:
+1 after my latest comments

sync-core/src/main/java/com/cloudant/sync/datastore/SavedHttpAttachment.java
rhyshort: done in cbb1ee7

sync-core/src/main/java/com/cloudant/sync/datastore/SavedHttpAttachment.java
rhyshort: done in cc7bcba

brynh:
+1

PR #155, opened 2015-06-04
47761 use db for blob fnames

mikerhodes:
This PR updates AttachmentManager to use the `attachments_key_filename` mapping table from PR https://github.com/cloudant/sync-android/pull/153.

We update:
1. `fileFromKey` to look up a key in the `attachments_key_filename` table, generating a key if required.
2. `purgeAttachments` to also clean up entries in the `attachments_key_filename` table as well as files on disk.

Questions:
- Should `fileFromKey` take a `create` parameter, and only generate a file if it's set to true (to avoid the `get` methods generating filenames for keys that don't exist?
   - I decided YES here.
- How's the exception and error handling?
- Is there a better purge implementation?
   - Not sure the whole delete row in thing is very performant, but couldn't see a better way (does this make the whole `attachments_key_filename` idea bad?)

reviewer @rhyshort 
reviewer @brynh

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
+++ b/sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
@@ -139,7 +139,18 @@ class AttachmentManager {
             throw new AttachmentNotSavedException("Could not insert attachment " + a + " into database with values " + values + "; not copying to attachments directory");
         }
         // move file to blob store, with file name based on sha1
-        File newFile = fileFromKey(db, sha1);
+        File newFile = null;
+        try {
+            newFile = fileFromKey(db, sha1);
+        } catch (NameGenerationException ex) {
+            // As we couldn't generate a filename, we can't save the attachment. Clean up
+            // temporary file and throw an exception.
+            if (a.tempFile.exists()){
+                a.tempFile.delete();
+            }
+            throw new AttachmentNotSavedException("Couldn't generate name for new attachment", ex);
+        }
+
         try {
             FileUtils.moveFile(a.tempFile, newFile);
         } catch (FileExistsException fee) {
@@ -245,7 +256,7 @@ class AttachmentManager {
         }
     }
 
-    protected Attachment getAttachment(SQLDatabase db, BasicDocumentRevision rev, String attachmentName) {
+    protected Attachment getAttachment(SQLDatabase db, BasicDocumentRevision rev, String attachmentName) throws AttachmentException {
         Cursor c = null;
         try {
              c = db.rawQuery(SQL_ATTACHMENTS_SELECT,
@@ -263,7 +274,10 @@ class AttachmentManager {
 
             return null;
         } catch (SQLException e) {
-            return null;
+            logger.log(Level.SEVERE,
+                    String.format("Failed to get %1$s for doc %2$s", rev.getId(), attachmentName),
+                    e);
+            throw new AttachmentException(e);
         } finally {
             DatabaseUtils.closeCursorQuietly(c);
         }
@@ -425,7 +439,7 @@ class AttachmentManager {
      * @param key key to lookup filename for.
      * @return File object for blob associated with {@code key}.
      */
-    private File fileFromKey(SQLDatabase db, byte[] key) {
+    private File fileFromKey(SQLDatabase db, byte[] key) throws NameGenerationException {
 
         String keyString = keyToString(key);
         String filename = null;
@@ -474,8 +488,16 @@ class AttachmentManager {
         if (filename != null) {
             return new File(attachmentsDir, filename);
         } else {
-            return null;
+            throw new NameGenerationException("Couldn't generate unique filename for attachment");
         }
+    }
+
+    private class NameGenerationException extends AttachmentException {
+
+        NameGenerationException(String msg) {
+            super(msg);
+        }
+
     }
 }
 
--- a/sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
+++ b/sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
@@ -141,7 +141,7 @@ class AttachmentManager {
         // move file to blob store, with file name based on sha1
         File newFile = null;
         try {
-            newFile = fileFromKey(db, sha1, true);
+            newFile = fileFromKey(db, sha1, attachmentsDir, true);
         } catch (AttachmentException ex) {
             // As we couldn't generate a filename, we can't save the attachment. Clean up
             // temporary file and throw an exception.
@@ -267,7 +267,7 @@ class AttachmentManager {
                 String type = c.getString(3);
                 int encoding = c.getInt(4);
                 int revpos = c.getInt(7);
-                File file = fileFromKey(db, key, false);
+                File file = fileFromKey(db, key, attachmentsDir, false);
                 return new SavedAttachment(attachmentName, revpos, sequence, key, type, file,
                         Attachment.Encoding.values()[encoding], this.attachmentStreamFactory);
             }
@@ -295,7 +295,7 @@ class AttachmentManager {
                 String type = c.getString(3);
                 int encoding = c.getInt(4);
                 int revpos = c.getInt(7);
-                File file = fileFromKey(db, key, false);
+                File file = fileFromKey(db, key, attachmentsDir, false);
                 atts.add(new SavedAttachment(name, revpos, sequence, key, type, file,
                         Attachment.Encoding.values()[encoding], this.attachmentStreamFactory));
             }
@@ -449,6 +449,7 @@ class AttachmentManager {
      *
      * @param db database to use.
      * @param key key to lookup filename for.
+     * @param attachmentsDir Root directory for attachment blobs.
      * @param allowCreateName if the is no existing mapping, whether one should be created
      *                        and returned. If false, and no existing mapping, AttachmentException
      *                        is thrown.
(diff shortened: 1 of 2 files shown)


sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -348,47 +382,153 @@ protected void purgeAttachments(SQLDatabase db) {
...
+     * Existing attachments will have an entry in db, so the method just looks this up
+     * and returns the File object for the path.
+     *
+     * For new attachments, the {@code key} doesn't already have an associated filename,
+     * one is generated and inserted into the database before returning a File object
+     * for blob associated with {@code key}.
+     *
+     * @param db database to use.
+     * @param key key to lookup filename for.
+     * @return File object for blob associated with {@code key}.
```
rhyshort: Missing javadoc for `allowCreateName` param

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -348,47 +382,153 @@ protected void purgeAttachments(SQLDatabase db) {
...
+     */
+    File fileFromKey(SQLDatabase db, byte[] key, boolean allowCreateName) throws AttachmentException {
+
+        String keyString = keyToString(key);
+        String filename = null;
+
+        db.beginTransaction();
```
rhyshort: Won't this be running on the queue, which provides the transaction when you call `submitTransaction`? So do we need a transaction within a transaction?

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -348,47 +382,153 @@ protected void purgeAttachments(SQLDatabase db) {
...
+        try {
+            Cursor c = db.rawQuery(SQL_FILENAME_LOOKUP_QUERY, new String[]{ keyString });
+            if (c.moveToFirst()) {
+                filename = c.getString(0);
+                System.out.println(String.format("FOUND filename %1$s for key %2$s", filename, keyString));
```
rhyshort: This should be using the `logger`

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -348,47 +382,153 @@ protected void purgeAttachments(SQLDatabase db) {
...
+            } else if (allowCreateName) {
+                filename = generateFilenameForKey(db, keyString);
+                System.out.println(String.format("ADDED filename %1$s for key %2$s", filename, keyString));
```
rhyshort: This should also being using the `logger`

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -348,47 +382,153 @@ protected void purgeAttachments(SQLDatabase db) {
...
+            }
+            c.close();
+            db.setTransactionSuccessful();
+        } catch (SQLException e) {
+            e.printStackTrace();
```
rhyshort: Appropriate logging for this exception

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: Corrected in dc36c48e162645271cdd5a79cd4004be11dea5a4.

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: I'm not sure it's guaranteed to be the case; I don't think there's harm in the sub-transaction?

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: Corrected in 6f3dafdc0e23ba563f4c0579b67a3e71ca641bb8.

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: Ditto, 6f3dafdc0e23ba563f4c0579b67a3e71ca641bb8.

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: I think I got this right in 6f3dafdc0e23ba563f4c0579b67a3e71ca641bb8?

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -348,47 +382,158 @@ protected void purgeAttachments(SQLDatabase db) {
...
+     *
+     * We try inserting the new record into attachments_key_filename to find a
+     * unique filename rather than checking on disk filenames. This is because we
+     * can make use of the fact that this method is called on a serial database
+     * queue to make sure our name is unique, whereas we don't have that guarantee
+     * for on-disk filenames. This works because filename is declared
+     * UNIQUE in the attachments_key_filename table.
+     *
+     * We allow up to 200 random name generations, which should give us many millions
+     * of files before a name fails to be generated and makes sure this method doens't
```
brynh: Typo `doens't`

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -348,47 +382,172 @@ protected void purgeAttachments(SQLDatabase db) {
...
-                            if (!deleted) {
-
-                                logger.warning("Could not delete file from BLOB store: " +
-                                        f.getAbsolutePath());
-                            }
-                        } catch (SecurityException e) {
-                            logger.log(Level.WARNING, "SecurityException when trying to delete file " +
-
-                                    "from BLOB store: " + f.getAbsolutePath(), e);
+            c.close();
```
tomblench: Don't need to call `close()` since you close it in `finally` block

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -348,47 +382,172 @@ protected void purgeAttachments(SQLDatabase db) {
...
+     * @param db database to use.
+     * @param key key to lookup filename for.
+     * @param allowCreateName if the is no existing mapping, whether one should be created
+     *                        and returned. If false, and no existing mapping, AttachmentException
+     *                        is thrown.
+     * @return File object for blob associated with {@code key}.
+     * @throws AttachmentException if a mapping doesn't exist and {@code allowCreateName} is
+     *          false or if the name generation process fails.
+     */
+    File fileFromKey(SQLDatabase db, byte[] key, boolean allowCreateName) throws AttachmentException {
```
tomblench: I'm a bit uneasy about the `allowCreateName` - is it overly verbose to split this into two methods and have the caller decide whether to create the mapping?

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -348,47 +382,172 @@ protected void purgeAttachments(SQLDatabase db) {
...
+            ContentValues contentValues = new ContentValues();
+            contentValues.put("key", keyString);
+            contentValues.put("filename", candidate);
+            result = db.insert(ATTACHMENTS_KEY_FILENAME, contentValues);
+
+            if (result != -1) {  // i.e., insert worked, filename unique
+                filename = candidate;
+            }
+
+            tries++;
```
tomblench: it's a style thing but you can put `tries++` in your `while` loop

mikerhodes: I tend to prefer it within the loop body rather than condition, but it's good to know. I guess I use `++` in a `for` loop's condition all the time so I don't really have a justification for my preference :)

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: I found the logic more easy to reason about if there was just one call -- allows to enforce a transaction around the "is there a key? no? generate one" logic.

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: Fixed in 7d887b898c2f46c193b01c13276a1f2e59821105.

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: Fixed in 4c41286c87d18830946cbb4d34c8d0177739c00a.

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
tomblench:
```java
for(int tries=0; result == -1 && tries < 200; tries++)
```

is another way

tomblench:
OK +1

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
rhyshort: I'm pretty sure all out insertions are in a transaction, I'm not sure if a sub transaction is a good idea, since it is just a lie in `SQLiteWrapper`

brynh:
+1

sync-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: I'm not sure what you mean by "lie", but if a sub-transaction fails, the overall top level transaction will be rolled back, https://github.com/cloudant/sync-android/blob/master/sync-javase/src/main/java/com/cloudant/sync/sqlite/sqlite4java/SQLiteWrapper.java#L167

PR #161, opened 2015-06-09
Add $mod operator support to Query

alfinkel:
_What:_

This PR adds support for the `$mod` operator for use in Query.  It is the port of `$mod` operator functionality already delivered to the CDTDatastore repository in https://github.com/cloudant/CDTDatastore/pull/152.

Due to PR size, there will be a follow up PR that will contain Query Executor tests for `$mod` and the `$mod` documentation.

_Why:_

In an effort to provide as much functionality to Query as is available elsewhere (such as MongoDB), we are adding support to the $mod operator so that users can write queries containing modulo arithmetic.

_How:_

Add logic to the validator code, the SQLtranslator code, and the unindexed matcher code to support the safe use of `$mod` in queries.

_Tests:_

- Validation tests have been added to QueryValidatorTest exercise argument validation and query normalization for queries containing the $mod operator.
- SQL WHERE clause generation tests have been added to QuerySqlTranslatorTest for queries containing the `$mod` operator.
- Tests have been added to UnindexedMatcherTest to exercise the unindexed matcher engine for queries containing the `$mod` operator.

reviewer @mikerhodes 
reviewer @gadamc

BugId: 44631

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
@@ -469,6 +469,7 @@ class QuerySqlTranslator {
         operatorMap.put(LT, "<");
         operatorMap.put(LTE, "<=");
         operatorMap.put(IN, "IN");
+        operatorMap.put(MOD, "%");
 
         for (Object rawComponent: clause) {
             Map<String, Object> component = (Map<String, Object>) rawComponent;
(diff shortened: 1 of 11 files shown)


sync-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -522,6 +523,15 @@ protected static SqlParts whereSqlForAndClause(List<Object> clause, String index
                         // This was validated during normalization.
                         List<Object> inList = (List<Object>) negatedPredicate.get(operator);
                         placeholder = placeholdersForInList(inList, sqlParameters);
+                    } else if (operator.equals(MOD)) {
+                        // The predicate map value must be a two element list containing integers
+                        // here.  This was validated during normalization.
+                        List<Integer> modulus = (List<Integer>) negatedPredicate.get(operator);
+                        // Casting the remainder as integer ensures proper processing of the modulo
+                        // operation by SQLite.  If left un-cast, unexpected results occur.
+                        placeholder = String.format("? %s CAST(? AS INTEGER)", operatorMap.get(EQ));
```
gadamc: So, I notice that we don't use this CAST AS INTEGER on the obj-c code: https://github.com/cloudant/CDTDatastore/blob/master/Classes/common/query/CDTQQuerySqlTranslator.m#L429

alfinkel: That is correct.  This is an issue with how we communicate with SQLite rather than SQLite itself.  On the Android/Java side it looks like the remainder bind argument is being passed in as a String to SQLite.  So the SQL statement that SQLite tries to process looks something like `SELECT _id FROM table WHERE field % 2 = "1"`, which will never return records.  Adding the CAST ensures that the remainder is handled as an integer.  This does not happen on the iOS/Obj C side.

gadamc:
+1

mikerhodes:
+1

PR #163, opened 2015-06-10
Add EndToEndEncryptionTest to test files encrypted on disk

mikerhodes:
This test creates a database with a key using the DatastoreManager.
It then looks at the JSON database, the indexes database and the
attachment file created to check they are encrypted or not, using
parameterisation to check both encrypted and non-encrypted
datastores.

End to end tests are in sync-android project so they are only
tested on Android as JavaSE doesn't support encryption yet.

It also parameterises the test so we _always_ test both the
encrypted and non-encrypted case when running on Android for
the EndToEndEncryptionTest case.

I copied TestUtils into the sync-android library to avoid
dependency issues with the non-Android build where the class
wasn't available to the sync-android sub-project compilation.

reviewer @tomblench 
reviewer @brynh

### DIFF
--- a/sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
+++ b/sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
@@ -14,6 +14,10 @@
 
 package com.cloudant.sync.datastore.encryption;
 
+import com.cloudant.sync.datastore.Attachment;
+import com.cloudant.sync.datastore.AttachmentException;
+import com.cloudant.sync.datastore.BasicDocumentRevision;
+import com.cloudant.sync.datastore.ConflictException;
 import com.cloudant.sync.datastore.Datastore;
 import com.cloudant.sync.datastore.DatastoreManager;
 import com.cloudant.sync.datastore.DatastoreNotCreatedException;
@@ -21,6 +25,7 @@ import com.cloudant.sync.datastore.DocumentBodyFactory;
 import com.cloudant.sync.datastore.DocumentException;
 import com.cloudant.sync.datastore.MutableDocumentRevision;
 import com.cloudant.sync.datastore.UnsavedFileAttachment;
+import com.cloudant.sync.datastore.UnsavedStreamAttachment;
 import com.cloudant.sync.query.IndexManager;
 
 import net.sqlcipher.database.SQLiteDatabase;
@@ -35,19 +40,27 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 
+import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.StringBufferInputStream;
+import java.io.StringReader;
+import java.nio.charset.StandardCharsets;
+import java.security.InvalidKeyException;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.Map;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
 
 
 /**
@@ -56,6 +69,10 @@ import static org.junit.Assert.assertThat;
  */
 @RunWith(Parameterized.class)
 public class EndToEndEncryptionTest {
+
+    public static final byte[] KEY = new byte[]{-123, 53, -22, -15, -123, 53, -22, -15, 53, -22, -15,
+            -123, -22, -15, 53, -22, -123, 53, -22, -15, -123, 53, -22, -15, 53, -22,
+            -15, -123, -22, -15, 53, -22};
 
     static {
         // Load SQLCipher libraries
@@ -87,9 +104,7 @@ public class EndToEndEncryptionTest {
 
         if(dataShouldBeEncrypted) {
             this.datastore = this.datastoreManager.openDatastore(getClass().getSimpleName(),
-                    new SimpleKeyProvider(new byte[] { -123, 53, -22, -15, -123, 53, -22, -15, 53, -22, -15,
-                            -123, -22, -15, 53, -22, -123, 53, -22, -15, -123, 53, -22, -15, 53, -22,
-                            -15, -123, -22, -15, 53, -22 }));
+                    new SimpleKeyProvider(KEY));
         } else {
             this.datastore = this.datastoreManager.openDatastore(getClass().getSimpleName());
         }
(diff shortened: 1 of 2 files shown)


sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
```diff
@@ -0,0 +1,205 @@
...
+/**
+ * Test that when you open a database with an encryption key, the content on disk
+ * is all encrypted.
+ */
+@RunWith(Parameterized.class)
+public class EndToEndEncryptionTest {
+
+    static {
+        // Load SQLCipher libraries
+        SQLiteDatabase.loadLibs(ProviderTestUtil.getContext());
```
tomblench: Does this have to be static or can it be in `@BeforeClass`? Then you could make it switchable on `dataShouldBeEncrypted`.

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+    @Parameterized.Parameter
+    public boolean dataShouldBeEncrypted;
+
+    String datastore_manager_dir;
+    DatastoreManager datastoreManager;
+    Datastore datastore = null;
+
+    // Magic bytes are "SQLite format 3" + null-terminator
+    byte[] sqlCipherMagicBytes = hexStringToByteArray("53514c69746520666f726d6174203300");
```
tomblench: Took me a while to realise this is plain text, why not `"SQLite format 3".getBytes()` for clarity?

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+        File[] contents = attachmentsFolder.listFiles();
+        assertNotNull("Didn't find expected attachments folder", contents);
+        assertThat("Didn't find expected file in attachments", contents.length, IsEqual.equalTo(1));
+        InputStream in = new FileInputStream(contents[0]);
+
+        if(dataShouldBeEncrypted) {
+
+            byte[] actualContent = new byte[expectedFirstAttachmentByte.length];
+            int readLength = in.read(actualContent);
```
tomblench: If you just want one byte you could do `in.read()` (the version for reading one byte at a time) - is it possible to assert on anything else in the encrypted attachment?

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/TestUtils.java
```diff
@@ -0,0 +1,178 @@
+/**
```
tomblench: Instead of copying this file you should be able to supplement the sourcesets in the gradle file, someting like
```
java {
  sourceSets {
    test {
      // some gradle magic
    }
  }
}

File: sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java [hunk 5]
```diff
@@ -0,0 +1,205 @@
...
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][]{
+                {false}, {true},
+        });
+    }
+
+    @Parameterized.Parameter
+    public boolean dataShouldBeEncrypted;
+
+    String datastore_manager_dir;
```
brynh: Is there any reason this uses underscores instead of camelcase?

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
```diff
@@ -0,0 +1,205 @@
...
+    DatastoreManager datastoreManager;
+    Datastore datastore = null;
```
brynh: Explicitly initialising this to null is unnecessary.

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
```diff
@@ -0,0 +1,205 @@
...
+        assertNotNull("Didn't find expected attachments folder", contents);
+        assertThat("Didn't find expected file in attachments", contents.length, IsEqual.equalTo(1));
+        InputStream in = new FileInputStream(contents[0]);
+
+        if(dataShouldBeEncrypted) {
+
+            byte[] actualContent = new byte[expectedFirstAttachmentByte.length];
+            int readLength = in.read(actualContent);
+            assertEquals("Didn't read full buffer", actualContent.length, readLength);
+            assertArrayEquals("First byte not version byte", actualContent, expectedFirstAttachmentByte);
```
brynh: The expected and actual arguments to `assertArrayEquals` are the wrong way round. This will make the output slightly confusing if the test fails.

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
mikerhodes: It's a good point I'd not thought of. However, once we load the libraries into the JVM, we've loaded the libraries for all the tests, presuming they are all run in the same JVM? I'm not sure we gain from using the `@BeforeClass` method. Happy to change it if you think it's worth while.

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
mikerhodes: Unfortunately it's not possible, the next 16 bytes are the random IV and the next the encrypted data. I could check the blob stored isn't just the plaintext, but I'm not sure that buys us more information.

I'd prefer to stick with the buffer method, just because it's more familiar across the code.

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
mikerhodes: I copied it from elsewhere, where's it's inexplicably underscored. So History. Fixed in 7cf2df9ab856dfb00bfbc961132fa4ad91264a20.

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
mikerhodes: Corrected in 22d630d2554afc64a9c2313b17974d20b499d4a0.

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/EndToEndEncryptionTest.java
mikerhodes: Good spot, corrected in f3f4771720599b5f3a6e1ca5e29c57f621881f83.

tomblench:
+1 :cake: 
😻

brynh:
+1

PR #165, opened 2015-06-12
Stop Base64 encoded strings containing new lines

rhyshort:
Stop Base64 encoding on android adding new lines to large strings when encoding.

reviewer @mikerhodes 
reviewer @tomblench

### DIFF
--- a/sync-core/src/main/java/com/cloudant/android/Base64OutputStreamFactory.java
+++ b/sync-core/src/main/java/com/cloudant/android/Base64OutputStreamFactory.java
@@ -28,7 +28,8 @@ public class Base64OutputStreamFactory {
             if (Misc.isRunningOnAndroid()) {
                 Class c = Class.forName("android.util.Base64OutputStream");
                 Constructor ctor = c.getDeclaredConstructor(OutputStream.class, int.class);
-                return (OutputStream)ctor.newInstance(os, 0);
+                // 2 = android.util.BASE64.NO_WRAP http://developer.android.com/reference/android/util/Base64.html#NO_WRAP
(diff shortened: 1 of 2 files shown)


sync-core/src/main/java/com/cloudant/android/Base64OutputStreamFactory.java
```diff
@@ -28,7 +28,7 @@ public static OutputStream get(OutputStream os) {
             if (Misc.isRunningOnAndroid()) {
                 Class c = Class.forName("android.util.Base64OutputStream");
                 Constructor ctor = c.getDeclaredConstructor(OutputStream.class, int.class);
-                return (OutputStream)ctor.newInstance(os, 0);
+                return (OutputStream)ctor.newInstance(os, 2);
```
mikerhodes: Can we statically use http://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getDeclaredField%28java.lang.String%29 to get the field value for NO_WRAP?

tomblench:
+1

sync-core/src/main/java/com/cloudant/android/Base64OutputStreamFactory.java
rhyshort: We can, but I really don't see why we should. Its a lot of code execution to get the value 2, and it's not going to improve readability. It's not going to change between SDKs since the Java Compiler should treat `Base64.NO_WRAP` as a compile time constant, since its defined as `static final int`

mikerhodes:
+1. Obv. squash up before merging.

PR #166, opened 2015-06-12
Pass customer CouchConfig headers to HttpConnection

mikerhodes:
This is a pretty hacky fix in lieu of a complete fix via the new filters work.

I believe it works, though I'm not sure we've tests which check it.

reviewer @tomblench 
reviewer @rhyshort

### DIFF
--- a/sync-core/src/main/java/com/cloudant/http/HttpConnection.java
+++ b/sync-core/src/main/java/com/cloudant/http/HttpConnection.java
@@ -23,7 +23,10 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 
 /**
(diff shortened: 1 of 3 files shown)


sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -140,17 +143,27 @@ public void setRequestBody(InputStream input, long inputLength) {
...
+
+
+        if (contentType != null && !lowerCaseExistingHeaders.contains("content-type")) {
             connection.setRequestProperty("Content-type", contentType);
         }
-        connection.setRequestProperty("User-Agent", userAgent);
-        if (url.getUserInfo() != null) {
+
+        if (!lowerCaseExistingHeaders.contains("user-agent")) {
+            connection.setRequestProperty("User-Agent", userAgent);
```
rhyshort: Shouldn't we always be setting the user agent field and not allowing users to overwrite it?

mikerhodes: We need to allow users to set their own `User-Agent` in my opinion. The solution I'd advocate for most cases is providing a way to customise part of the big long platform string.

rhyshort: We should probably deny the ability to set the user agent, then open if up for customization, rather than allowing it to be set to anything, then restricting it in the future.

tomblench: I don't see the value in preventing the customer from changing these headers. They may even have a valid use case for setting `user-agent`, and why should we stop them?

rhyshort: It would think it makes it harder for support to support them when they encounter problems when using it with the service.

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -42,26 +42,35 @@
...
     // - map 404 to NoResourceException
-    // - if there's a couch error returned as json, unmarshall and throw
+    // - if there's a couch error returned as json, un-marshall and throw
     // - anything else, just throw the IOException back, use the cause part of the exception?
 
     // it needs to catch eg FileNotFoundException and rethrow to emulate the previous exception handling behaviour
     private InputStream executeToInputStream(HttpConnection connection) throws CouchException {
 
-        // all couchclient requests want to receive application/json responses
+        // all CouchClient requests want to receive application/json responses
```
tomblench: move the comment down

tomblench:
OK +1

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
mikerhodes: Fixed in 5d33a7a68831a179ab8d0d089fca7018ad709225.

rhyshort:
+1

PR #167, opened 2015-06-12
Fix indexing of empty array fields

alfinkel:
_What:_

Fix a bug that, incorrectly, does not index documents that contain an empty array field value.

_Why:_

Documents that contain an empty array should be index-able.

_How:_

Fix logic in the index updater that handles SQL INSERT statement creation for documents to be stored in Query index tables. Logic for handling an empty array field value is now the same as the handling of an index-able field with a missing value in a document.

_Tests:_

Add two additional index updater test cases to the test empty array field values.

reviewer @mikerhodes 
reviewer @gadamc 

BugId: 48139

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
+++ b/sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
@@ -22,6 +22,7 @@ import com.cloudant.sync.util.DatabaseUtils;
 
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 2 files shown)


sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
```diff
@@ -248,53 +249,52 @@ public Boolean call() {
         }
```
gadamc: the msg string above here should say String.format("%s. Only one array field per index allowed", msg)

alfinkel: Fixed in 940c61b

sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
```diff
@@ -248,53 +249,52 @@ public Boolean call() {
...
-                parameter = populateDBParameter(fieldNames,
-                                                initialIncludedFields,
-                                                initialArgs,
-                                                indexName,
-                                                rev);
-                if (parameter == null) {
-                    return null;
+            arrayFieldValues = (List) ValueExtractor.extractValueForFieldName(arrayFieldName,
+                                                                              rev.getBody());
+            if (arrayFieldValues != null && arrayFieldValues.size() > 0) {
```
gadamc: So, I think that if arrayFieldValues is null, then something terrible must have happened, right? It seems impossible. 

But having that check here makes the code confusing, no?

Perhaps my following suggestion could be a bit of overkill, but could you create a Map of Lists, like `mapOfArrayFieldValues` and populate it during the initial loop through `fieldNames`. Then check the number of objects in the map instead of having `arrayCount` and `arrayFieldName` would be the only key in `mapOfArrayFieldValues`. 

But the code does work as written. I was just thrown a bit by the check for null. Or maybe I'm wrong at it can by null here?

sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
```diff
@@ -318,6 +318,12 @@ private DBParameter populateDBParameter(List<String> fieldNames,
...
 
             if (value != null) {
+                // Do not include an empty array field as you would not include a
+                // field with no value.
+                if (value instanceof List && ((List) value).size() == 0) {
+                    continue;
+                }
+
                 includeFieldNames.add(fieldName);
                 args.add(value);
```
gadamc: So, won't an array's value be added to this `args` list twice? It should already be there as part of `initialArgs` on line 309, right? And then  it would get added here again?

gadamc:
This might be a separate PR -- but can we add a test for the case where an element of an indexed array is another object?  As in `pets: [cat, dog, {llama: bob}, fish]`  Does `llama:bob` get indexed and can we successfully query for it?

sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
alfinkel: Actually this line of code is what, in part, fixes our empty array issue.  The check looks to see if there are values in arrayFieldValues.  If there are, then array processing happens as before.  But in the event that the array is empty, processing falls through to the field processing section which then handles an empty array as it would a missing field value.  The check for null followed by the zero size check is standard in Java since it is bad practice to check a size of 0 before knowing you have a non-Null object.  It is essentially the same logic as in CDTDatastore https://github.com/cloudant/CDTDatastore/blob/48139-query-fix-empty-array-field-indexing/Classes/common/query/CDTQIndexUpdater.m#L339.

sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
alfinkel: There is a check above that skips already included fields which means that it skips already included arguments `initialArgs`.  Take a look a few lines up https://github.com/cloudant/sync-android/pull/167/files#diff-0d8c3940dfc1c0e1639ac3ad7ed9a2c5R314.

gadamc: right. thanks.

alfinkel:
Actually here is more info on the topic of sub-documents inside arrays: Dotted notation to index or query sub-documents in arrays. is unsupported. Check out the "Arrays" section under https://github.com/cloudant/CDTDatastore/blob/master/doc/query.md#unsupported-features. FB Case 47341 has been created to track this work.

gadamc:
+1

sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
```diff
@@ -232,69 +233,68 @@ public Boolean call() {
...
+                                                                initialArgs,
+                                                                indexName,
+                                                                rev);
+                    if (parameter == null) {
+                        return null;
+                    }
+                    parameters.add(parameter);
                 }
-                parameters.add(parameter);
+                return  parameters;
```
mikerhodes: This early return statement belies confusing logic. The logic you're expressing is essentially an `if/else` block, with the if condition you want to alternate your logic being `if there was a non-empty array in the fields to be indexed`. However, the code doesn't make this obvious, instead resorting to an early return statement because of using the nested construct for what is logically a single condition.

Can you express this as a single if/else somehow, I think you need to do your checks in serial rather than nesting, like:

```
List<Object> arrayFieldValues = null;
if (arrayCount == 1) {
    arrayFieldValues = (List) ValueExtractor.extractValueForFieldName(arrayFieldName,
                                                                      rev.getBody());
}

if (arrayFieldValues != null && arrayFieldValues.size() > 0) {
    for (Object value: arrayFieldValues) {
      /* ... */
    }
    return parameters;
} else {
    List<String> initialIncludedFields = Arrays.asList("_id", "_rev");
    /* ... */
    return parameters;
}
```

sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
```diff
@@ -318,6 +318,12 @@ private DBParameter populateDBParameter(List<String> fieldNames,
             Object value = ValueExtractor.extractValueForFieldName(fieldName, rev.getBody());
 
             if (value != null) {
+                // Do not include an empty array field as you would not include a
+                // field with no value.
+                if (value instanceof List && ((List) value).size() == 0) {
+                    continue;
+                }
```
mikerhodes: If we ever include logic after the `if` statement, it's easy to miss that this `continue` statement would cause it to be ignored and introduce a bug. I'd reverse the condition and put the logic `includeFieldNames.add(fieldName); args.add(value);` inside the nested if.

sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
alfinkel: Restructured `if` statements in 25918a1

sync-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
alfinkel: Restructured `if` logic in 25918a1

mikerhodes:
+1

PR #168, opened 2015-06-16
Correctly persist and send length & encoded_length values (3rd attempt)

tomblench:
What persist to the local DB the length and encoded_length metadata properties when they are sent with attachments.
Why currently this data is not persisted or used to check that the actual length equals the indicated length.
How update various Attachment classes and related methods to take length and encoded_length arguments and assert that actual attachment length equals signalled attachment length
Testing mostly covered by existing tests

reviewer @rhyshort 
reviewer @mikerhodes

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
+++ b/sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
@@ -43,6 +43,8 @@ public class PreparedAttachment {
     public final long length;
     public final long encodedLength;
 
+    private static final int NO_ENCODED_LENGTH = 0;
+
     /**
      * Prepare an attachment by copying it to a temp location and calculating its sha1.
      *
@@ -97,9 +99,8 @@ public class PreparedAttachment {
         //Set attachment length from bytes read in input stream
         if (this.attachment.encoding == Attachment.Encoding.Plain) {
             this.length = totalRead;
-            // 0 signals "no encoded length" - this is consistent with couch which does not send
-            // encoded_length if the encoding is "plain"
-            this.encodedLength = 0;
+            // couch does not send encoded_length if the encoding is "plain"
+            this.encodedLength = NO_ENCODED_LENGTH;
         } else {
             // the pre-encoded length is known, so store it
             this.length = length;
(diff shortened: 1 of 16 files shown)


rhyshort:
+1, once the commit message reads something sensible

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -81,16 +93,20 @@ public PreparedAttachment(Attachment attachment,
...
             IOUtils.closeQuietly(tempFileOutStream);
         }
+        
         //Set attachment length from bytes read in input stream
-        this.length = totalRead;
+        if (this.attachment.encoding == Attachment.Encoding.Plain) {
+            this.length = totalRead;
+            // 0 signals "no encoded length" - this is consistent with couch which does not send
+            // encoded_length if the encoding is "plain"
+            this.encodedLength = 0;
```
mikerhodes: It'd be nice to use a constant for this, e.g., `NO_ENCODED_LENGTH`?

sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
```diff
@@ -291,7 +291,17 @@ private int processOneChangesBatch(ChangesResultWrapper changeFeeds)
                                 atts.put(new String[]{documentRevs.getId(), documentRevs.getRev()}, preparedAtts);
 
                                 for (String attachmentName : attachments.keySet()) {
-                                    int revpos = (Integer) ((Map<String, Object>) attachments.get(attachmentName)).get("revpos");
+                                    Map attachmentMetadata = (Map)attachments.get(attachmentName);
+                                    int revpos = (Integer) attachmentMetadata.get("revpos");
+                                    String contentType = (String) attachmentMetadata.get("content_type");
+                                    String encoding = (String) attachmentMetadata.get("encoding");
+                                    long length = (Integer) attachmentMetadata.get("length");
+                                    long encodedLength = 0; // encodedLength can default to 0 if it's not encoded
```
mikerhodes: Constant here?

mikerhodes:
Looking good, add the constant if it makes sense. Otherwise +1.

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
tomblench: Agreed. The logic is hard to follow but I think we should be consistent with the naming and meaning of these values in CouchDB. So the constant will at least help with that.

PR #171, opened 2015-06-17
Fix AES key creation for Android versions less then KitKat (API 19)

kholdaway:
Encrypt/Decrypt methods now have a create the SecretKeySpec for the AES key.  This fixes issues with encryption of the PBK on versions of Android less than KitKat (API 19)

reviewer @mikerhodes 
reviewer @tomblench


mikerhodes:
Can you see any of the changes here https://developer.android.com/sdk/api_diff/19/changes.html that might have caused this? (I think `19` is kitkat).

mikerhodes:
Seems to follow http://docs.oracle.com/javase/7/docs/technotes/guides/security/crypto/CryptoSpec.html#SimpleEncrEx, so I think we're good here. Seems like any specific references are hard to find, so I guess I will need to retract my previous request.

+1.

tomblench:
+1
I added the StackOverflow solution URL to help remind our future selves why we did this.

mikerhodes:
Merging.

PR #172, opened 2015-06-22
48351 cache key provider

ddlees:
The CacheKeyProvider wraps another KeyProvider and caches an EncryptionKey. This addresses the impact to performance when multiple encrypted datastores are used, as seen with AndroidKeyProvider.

reviewer @mikerhodes 
reviewer @tomblench

### DIFF
--- a/sync-android/src/test/java/com/cloudant/sync/datastore/encryption/CachingKeyProviderTests.java
+++ b/sync-android/src/test/java/com/cloudant/sync/datastore/encryption/CachingKeyProviderTests.java
@@ -1,0 +1,167 @@
+package com.cloudant.sync.datastore.encryption;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.Random;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.junit.Assume.assumeNotNull;
+
+public class CachingKeyProviderTests {
+    @Before
+    public void beforeMethod() {
+        assumeNotNull(ProviderTestUtil.getContext());
+    }
+
+    @Test
+    public void testWithSimpleKeyProvider() {
+        // Generate 32-byte key
+        byte[] keyBytes = new RandomKeyProvider().getEncryptionKey().getKey();
+
+        // Create provider
+        CachingKeyProvider provider = new CachingKeyProvider(new SimpleKeyProvider(keyBytes));
+
+        // Get the key
+        EncryptionKey encryptionKey = provider.getEncryptionKey();
+        assertNotNull("EncryptionKey was not created", encryptionKey);
+
+        // Assert value is same as original key
+        assertTrue("encryptionKey bytes != keyBytes", Arrays.equals(encryptionKey.getKey(),
+                keyBytes));
+
+        // Get the cached key
+        EncryptionKey cachedKey = provider.getEncryptionKey();
+        assertNotNull("EncryptionKey was not cached", cachedKey);
+
+        // Assert cached value is same as original key
+        assertTrue("encryptionKey bytes != keyBytes", Arrays.equals(encryptionKey.getKey(),
+                keyBytes));
+
+        // Assert cached key and key are the same object reference
+        assertEquals("cachedKey did not match createdKey", encryptionKey,
+                cachedKey);
+    }
+
+    @Test
+    public void testWithAndroidKeyProvider() {
+        CachingKeyProvider provider = new CachingKeyProvider(new AndroidKeyProvider(ProviderTestUtil
+                .getContext(),
+                ProviderTestUtil.password, ProviderTestUtil.getUniqueIdentifier()));
+
+        // This will cause the provider to generate a key using the AndroidKeyProvider. The key
+        // will be encrypted, persisted, and stored in memory.
+        EncryptionKey createdKey = provider.getEncryptionKey();
+        assertNotNull("EncryptionKey was not created", createdKey);
+
+        // This will cause the provider to load the cached key.
+        EncryptionKey cachedKey = provider.getEncryptionKey();
+        assertNotNull("EncryptionKey was not cached", cachedKey);
+
+        // Assert cached key and created key are the same object reference
+        assertEquals("cachedKey did not match createdKey", createdKey,
+                cachedKey);
+
+        AndroidKeyProvider androidKeyProvider = (AndroidKeyProvider) provider.getKeyProvider();
+        androidKeyProvider.getManager().clearKey();
+    }
+
+    @Test
+    public void testTwoProvidersSameKey() {
+        // Generate 32-byte key
+        byte[] keyBytes = new RandomKeyProvider().getEncryptionKey().getKey();
+
+        // Create providers
+        CachingKeyProvider provider1 = new CachingKeyProvider(new SimpleKeyProvider(keyBytes));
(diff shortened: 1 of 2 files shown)


sync-core/src/main/java/com/cloudant/sync/datastore/encryption/CacheKeyProvider.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * This class implements the interface {@link KeyProvider} and it can be used to create an
+ * encrypted datastore.
+ * <p/>
+ * Given an user-provided {@link KeyProvider}, it provides an in-memory cache for retrieving an
+ * {@link EncryptionKey}. This improves performance when multiple encrypted datastores are used.
+ *
+ * @see KeyProvider
+ */
+public class CacheKeyProvider implements KeyProvider {
```
tomblench: How about `CachingKeyProvider` or `CachedKeyProvider`?

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/CacheKeyProviderTests.java
```diff
@@ -0,0 +1,121 @@
...
+        // This will cause the provider to generate a key using the AndroidKeyProvider. The key
+        // will be encrypted, persisted, and stored in memory.
+        EncryptionKey createdKey = provider.getEncryptionKey();
+        assertNotNull("EncryptionKey was not created", createdKey);
+
+        // This will cause the provider to load the cached key.
+        EncryptionKey cachedKey = provider.getEncryptionKey();
+        assertNotNull("EncryptionKey was not cached", cachedKey);
+
+        // Assert cached key and created key are the same
```
tomblench: You can also assert on pointer equality between `createdKey` and `cachedKey`

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/CacheKeyProviderTests.java
```diff
@@ -0,0 +1,121 @@
...
+        assertTrue("cachedKey did not match createdKey", Arrays.equals(createdKey.getKey(),
+                cachedKey.getKey()));
+
+        AndroidKeyProvider androidKeyProvider = (AndroidKeyProvider) provider.getKeyProvider();
```
tomblench: Why do these keys need to be cleared at the end of each method?

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/CacheKeyProvider.java
```diff
@@ -0,0 +1,64 @@
+/**
+ * Copyright (c) 2013 Cloudant, Inc. All rights reserved.
```
tomblench: Copyright header should be the 2015 Cloudant/IBM version

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/CacheKeyProviderTests.java
```diff
@@ -0,0 +1,121 @@
...
+                .getContext(),
+                ProviderTestUtil.password, identifier));
+
+        EncryptionKey provider1Key = provider1.getEncryptionKey();
+        assertNotNull("provider1 EncryptionKey was not created", provider1Key);
+
+        EncryptionKey provider2Key = provider2.getEncryptionKey();
+        assertNotNull("provider2 EncryptionKey was not created", provider2Key);
+
+        assertTrue("provider1 key and provider2 key should be equal", Arrays.equals(provider1Key
```
tomblench: Similar to above, you can assert that the pointers `provider1Key` and `provider2Key` are different here

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/CacheKeyProvider.java
```diff
@@ -0,0 +1,64 @@
...
+ * and limitations under the License.
+ */
+package com.cloudant.sync.datastore.encryption;
+
+import java.util.logging.Logger;
+
+/**
+ * This class implements the interface {@link KeyProvider} and it can be used to create an
+ * encrypted datastore.
+ * <p/>
```
tomblench: When I run `gradle docs` I get `error: self-closing element not allowed` - this may vary across Java versions so you might not get this.

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/CacheKeyProviderTests.java
ddlees: Using the AndroidKeyProvider the EncryptionKey is stored on device. It's cleared to put the device back into a clean state for each test.

tomblench:
+1 :+1:

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/CachingKeyProvider.java
```diff
@@ -0,0 +1,64 @@
...
+ * Given an user-provided {@link KeyProvider}, it provides an in-memory cache for retrieving an
+ * {@link EncryptionKey}. This improves performance when multiple encrypted datastores are used.
+ *
+ * @see KeyProvider
+ */
+public class CachingKeyProvider implements KeyProvider {
+
+    private static final Logger LOGGER = Logger.getLogger(CachingKeyProvider.class
+            .getCanonicalName());
+    private KeyProvider keyProvider;
```
mikerhodes: I suggest making this `final`.

sync-core/src/main/java/com/cloudant/sync/datastore/encryption/CachingKeyProvider.java
```diff
@@ -0,0 +1,64 @@
...
+package com.cloudant.sync.datastore.encryption;
+
+import java.util.logging.Logger;
+
+/**
+ * This class implements the interface {@link KeyProvider} and it can be used to create an
+ * encrypted datastore.
+ *
+ * Given an user-provided {@link KeyProvider}, it provides an in-memory cache for retrieving an
+ * {@link EncryptionKey}. This improves performance when multiple encrypted datastores are used.
```
mikerhodes: I think this paragraph should be the first, introductory paragraph as it's the raison d'etre for the class.

sync-android/src/test/java/com/cloudant/sync/datastore/encryption/CachingKeyProviderTests.java
```diff
@@ -0,0 +1,131 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.junit.Assume.assumeNotNull;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Arrays;
+
+public class CachingKeyProviderTests {
```
mikerhodes: I don't think you need to use AndroidKeyProvider for these tests. I think this will just slow down the test through the key generation and cleanup. I'd suggest running most of the tests with just the SimpleKeyProvider. Then you control the key used, so you can make sure the right key is returned (avoiding the checks that the keys are just "different" or the "same" without checking their values against known good versions).

In addition, you should write a test to make sure the key is actually being cached. I'd do this by creating an implementation of KeyProvider which returns a different key each time its getEncryptionKey method is called (e.g., use Random() to get a new key each time). When wrapped with CachingKeyProvider, if the key is being cached correctly, the same key will be returned by the CachingKeyProvider whereas if the key isn't being cached correctly a different key will be returned each time from the underlying `RandomKeyProvider`.]

I.e.:

```java
private class RandomKeyProvider implements KeyProvider {
  public EncryptionKey getEncryptionKey() {
    byte[] buf = new byte[32];
    new Random().nextBytes(buf);
    return new EncryptionKey(buf);
  }
}
```

mikerhodes:
This looks good to merge, once the commits addressing PR comments are squashed appropriately.

+1.

PR #173, opened 2015-06-23
Restructure encryption code into a new sub-project

brynh:
Restructure the sub-projects to move encryption specific Android code from
the existing sync-android sub-project to a new sync-android-encryption
sub-project.

Reference TestUtils.java from the existing version in sync-core and remove
the duplicate version in the sync-android sub-project.

This means users that don't need encryption don't need to import the encryption
specific code into their project.

reviewer @tomblench
reviewer @mikerhodes


sync-android-encryption/build.gradle
```diff
@@ -0,0 +1,72 @@
...
+    // for unit tests
+    testCompile group: 'org.hamcrest', name: 'hamcrest-all', version:'1.3'
+    testCompile group: 'junit', name: 'junit', version:'4.11'
+
+}
+
+sourceSets {
+    test {
+        java {
+            include '../sync-core/src/test/java/'
```
tomblench: Worth a comment why we do this, to bring in TestUtils.java

tomblench:
Looks pretty good to me, you will need to update README.md `There are currently three jar files for the datastore` and the maven/gradle examples.

Have you done an end-to-end test, something like publish maven repo locally and then consume it from a test app?

sync-android-encryption/build.gradle
brynh: Addressed in 4d2d913.

brynh:
Have updated README.md in 6c78f5e and also performed an end-to-end test using a local maven repo as suggested by @tomblench.

README.md
```diff
@@ -58,8 +59,10 @@ repositories {
 dependencies {
     // Other dependencies
     compile group: 'com.cloudant', name: 'cloudant-sync-datastore-core', version:'0.12.1'
-    // include this if you're targeting Android
+    // include this if you're targeting Android and do not want datastore encryption
     compile group: 'com.cloudant', name: 'cloudant-sync-datastore-android', version:'0.12.1'
+    // include this if you're targeting Android and want datastore encryption
+    compile group: 'com.cloudant', name: 'cloudant-sync-datastore-android-encryption', version:'0.12.1'
```
mikerhodes: To clarify my understanding, what we're saying here is that the `-encryption` jar is a substitute rather than in addition to the `cloudant-sync-datastore-android` jar?

brynh: That text was quite badly worded and definitely gave that incorrect impression. The `-encryption` jar is not a substitute for `cloudant-sync-datastore-android`, but should be used in addition. Improved wording in 1740b57.

sync-android-encryption/build.gradle
```diff
@@ -0,0 +1,73 @@
...
+
+            from components.java
+
+            groupId 'com.cloudant'
+            artifactId 'cloudant-sync-datastore-android-encryption'
+
+            pom.withXml {
+                Node rootNode = asNode()
+
+                rootNode.appendNode('name', 'Cloudant Sync Datastore Android: Android components')
```
mikerhodes: Should this have encryption somewhere in the name?

brynh: Good spot. Fixed in a81e2cb.

tomblench:
Looks good to me +1 - you will need to rebase before you can merge

mikerhodes:
Probably also need an entry in CHANGES

brynh:
Updated CHANGES in 77b464a

mikerhodes:
+1

PR #174, opened 2015-06-23
Fix javadoc warnings

tomblench:
What:

* squash all javadoc warnings
* re-word some javadoc
* remove useless 'public' specifiers on public interface methods
* remove some redundant `throws` clauses

Why:

running `gradle docs` outputs extraneous noise and makes it difficult to spot when there are genuine javadoc errors

Testing:

run `gradle docs` and no errors/warnings are reported; briefly scan HTML javadoc output to ensure documentation looks OK.

reviewer: @brynh 
reviewer: @ricellis

### DIFF
--- a/sync-core/src/main/java/com/cloudant/sync/datastore/DatastoreExtended.java
+++ b/sync-core/src/main/java/com/cloudant/sync/datastore/DatastoreExtended.java
@@ -47,16 +47,18 @@ public interface DatastoreExtended extends Datastore {
      * @param docId      The document id for the document
      * @param body       JSON body for the document
      * @return {@code DocumentRevision} of the newly created document
+     * @throws DocumentException if there is an error inserting the local document into the database
      */
-    public LocalDocument insertLocalDocument(String docId, DocumentBody body) throws DocumentException;
+    LocalDocument insertLocalDocument(String docId, DocumentBody body) throws DocumentException;
 
     /**
      * <p>Returns the current winning revision of a local document.</p>
      *
      * @param documentId id of the local document
      * @return {@code LocalDocument} of the document
+     * @throws DocumentNotFoundException if the document ID doesn't exist
      */
-    public LocalDocument getLocalDocument(String documentId) throws DocumentNotFoundException;
+    LocalDocument getLocalDocument(String documentId) throws DocumentNotFoundException;
 
     /**
      * <p>Deletes a local document.</p>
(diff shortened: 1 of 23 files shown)


ricellis:
+1

sync-core/src/main/java/com/cloudant/sync/datastore/DatastoreExtended.java
```diff
@@ -119,13 +121,26 @@
...
      *                        including the rev ID of {@code rev}. This list
-     *                        needs to be sorted in ascending order.
-     * @param preparedAttachments Attachments that have already been prepared, this is a
-     *                            Map of String[docId,revId], list of attachments
+     *                        needs to be sorted in ascending order
+     * @param attachments Attachments metadata and optionally data if {@code pullAttachmentsInline} true
+     * @param preparedAttachments Non-empty if {@code pullAttachmentsInline} false.
+     *                            Attachments that have already been prepared, this is a
+     *                            Map of String[docId,revId] → list of attachments
+     * @param pullAttachmentsInline
```
brynh: The description of this parameter is detached from the parameter name - it should probably be 2 separate sentences attached to the parameter name rather than 2 paragraphs.

tomblench: Agreed. The jdk 8 javadoc renders all of these tags in monospace which makes it look even worse!

sync-core/src/main/java/com/cloudant/sync/datastore/PreparedAttachment.java
```diff
@@ -41,7 +41,10 @@
      *
      * @param attachment The attachment to prepare
      * @param attachmentsDir The 'BLOB store' or location where attachments are stored for this database
-     * @throws AttachmentNotSavedException
+     * @param attachmentStreamFactory The @{link AttachmentStreamFactory} used for writing attachment
```
brynh: Typo `@{link` -> `{@link`

sync-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -89,6 +89,7 @@
 
     /**
      *  Constructs a new IndexManager which indexes documents in 'datastore'
+     *  @param datastore The @{code Datastore} to index
```
brynh: `@{code` is incorrect javadoc, and `{@link` would probably be better.

brynh:
+1 once previous comments are addressed.

PR #176, opened 2015-06-25
Fix query without any index capability

alfinkel:
_What:_

There is a bug in Query that requires at least one index be created before any query can execute.

_Why:_

The desired behavior is to allow for querying of documents regardless of the existence of any indexes.

_How:_

Remove the logic that causes a query to fail when no indexes are found and replace it with logic that catches the condition and instead retrieves a full list of document ids from the main datastore for the post-hoc matcher to use when generating the query results.

This logic is found in the SQL translator and the query executor.  Additionally a new method to retrieve just a list of document ids needs to be added to the datastore API.

reviewer @mikerhodes 
reviewer @gadamc 

BugId: 45362

### DIFF
--- a/sync-core/src/test/java/com/cloudant/sync/datastore/BasicDatastoreCRUDTest.java
+++ b/sync-core/src/test/java/com/cloudant/sync/datastore/BasicDatastoreCRUDTest.java
@@ -14,6 +14,10 @@
 
 package com.cloudant.sync.datastore;
 
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.hasSize;
+
 import com.cloudant.sync.sqlite.Cursor;
 import com.cloudant.sync.sqlite.SQLDatabase;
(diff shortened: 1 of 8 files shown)


gadamc:
Logic seems the same as iOS. Looks okay to me. +1

sync-core/src/test/java/com/cloudant/sync/datastore/BasicDatastoreCRUDTest.java
```diff
@@ -558,6 +555,13 @@ public void getAllDocuments() throws Exception {
     }
 
     @Test
+    public void getAllDocumentIds() throws Exception {
+        Assert.assertTrue(datastore.getAllDocumentIds().isEmpty());
+        createTwoDocuments();
+        Assert.assertEquals(2, datastore.getAllDocumentIds().size());
```
mikerhodes: Would like to see a check that these docIds are correct.

alfinkel: Changed in 480ae21

mikerhodes:
+1

PR #177, opened 2015-06-26
Fix pickWinnerOfConflicts to select current revision:

tomblench:
* What: fix pickWinnerOfConflicts which had incorrect behaviour with regard to deleted revisions, depending on the order of documents being inserted via replication.

Apply the CouchDB sorting criteria (highest rev wins, if there
is a tie then do a lexicographical compare of the revision id strings)
to all non-deleted leafs

* Why: previous behaviour was incorrect and could sometimes return deleted revisions as being 'current', as noticed by a customer

* Testing: new test cases `forceInsert_pickWinnerOfConflicts_Simple`, `forceInsert_pickWinnerOfConflicts_Complex` - see tests and comments for details


ricellis:
As discussed I'd like to see:
a second forceInsert_pickWinnerOfConflicts_Simple test that does the deletion/insert in the opposite order
some output from the asserts in forceInsert_pickWinnerOfConflicts_Complex so that if it does fail it is possible to try and unpick the deletion order that caused a failure

After that +1

rhyshort:
+1 :)

PR #178, opened 2015-06-29
Fix url encoding of local document urls

rhyshort:
Local documents should not be url encoded, remove / encoding so url
is correctly generated as _local/mylocal instead of _local&2FmyLocal

reviewer @tomblench 
reviewer @ricellis


ricellis:
Is a new test needed?

rhyshort:
Well I don't think so because we have test coverage for _local doc urls by making use of them during replication for checkpoint docs. But we can add some specifically to test for this for peace of mind

tomblench:
Agree that a test is needed to check that the URL string is the expected value

rhyshort:
@tomblench @ricellis test added in 69db968

ricellis:
+1

tomblench:
+1

PR #181, opened 2015-07-21
Expose HTTP Filters

rhyshort:
## What 
Expose HTTP request and response filters to developers using the library.

## Why
Allow developers to implement their own custom logic for authenticating e.g. OAuth or using a Cookie.

## How
Make Push and pull replication objects follow the builder pattern.
```java
new PullReplication()
.addFilters(...)
.source(...)
.target(...)
.start();
```
Use CouchConfig to pass the list of filters down. As a result the ```Replicator.createCouchConfig(URI uri```is now final, to ensure that when a config object is created it will have the filters.

## Reviewers

reviewer @tomblench 
reviewer @mikerhodes


.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: java
+sudo: false
```
tomblench: What's this for? Was this left in from something else - I know you talked about doing this in travis but I can't remember why and it doesn't seem relevant here.

sync-core/src/main/java/com/cloudant/sync/replication/Replication.java
```diff
@@ -29,9 +30,8 @@
 /**
  * <p>Abstract class which provides configuration for a replication.</p>
  *
- * <p>This class is abstract. Concrete classes
- * {@link com.cloudant.sync.replication.PullConfiguration}
- * and {@link com.cloudant.sync.replication.PushReplication} are used to
+ * <p>This class is abstract. Concrete class {@link com.cloudant.sync.replication.PushReplication}
```
tomblench: This doesn't make sense any more - the link is for PushReplication but the following text mentions push and pull

rhyshort: fixed in f0aece6

sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
```diff
@@ -69,4 +72,53 @@ public void start_StartedThenComplete() throws InterruptedException {
...
+        Assert.assertTrue(listener.exception instanceof CouchException);
+        CouchException couchException = (CouchException) listener.exception;
+        Assert.assertTrue(couchException.getStatusCode() == 400);
+        Assert.assertEquals(couchException.getReason(), "invalid_json");
+
+    }
+
+    @Test
+    public void testResponseFilters() throws Exception {
+        //to test the filters we make an invalid request
```
tomblench: comment doesn't seem to match code - it's a response filter so how can it make an invalid request?

sync-core/src/test/java/com/cloudant/sync/replication/ResponseStreamReaderFilter.java
```diff
@@ -0,0 +1,41 @@
...
+import com.cloudant.http.HttpConnection;
+import com.cloudant.http.HttpConnectionFilterContext;
+import com.cloudant.http.HttpConnectionResponseFilter;
+
+import java.io.IOException;
+
+/**
+ * Created by Rhys Short on 20/07/15.
+ */
+public class ResponseStreamReaderFilter implements HttpConnectionResponseFilter {
```
tomblench: Some documentation might be useful here. It looks like you are reading the response stream which will cause the caller to error because the stream has been exhausted at the point it wants to read the response.

I'm not sure if this is a useful test and as discussed, it might highlight the fact that response filters are fairly useless in their current guise. Maybe it's worth leaving out anything to do with response filters in this PR until they can do useful things like replace the response stream etc

.travis.yml
rhyshort: Travis gave me a nice big warning that the build was running on the legacy infrastructure, this removes the warning and makes the builds faster too

sync-core/src/test/java/com/cloudant/sync/replication/ResponseStreamReaderFilter.java
rhyshort: Its not perfect, but it gives us the knowledge that the filter added to the replication object gets passed down. I would side on leaving it in, and modifying it to use new and better APIs as they come available

rhyshort: Doc added in 0682bc6

sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
rhyshort: removed in 85e6b07

.travis.yml
tomblench: Fine, but that's not the purpose of this PR :)

sync-core/src/main/java/com/cloudant/sync/replication/PullReplication.java
```diff
@@ -54,6 +60,66 @@ public PullReplication() {
         /* Does nothing but we can now document it */
     }
 
+    /**
+     * Adds request filters to run for every request made by this replication
+     * @param requestFilters The filters to run
+     * @return The current instance of PullReplication
+     */
+    public PullReplication addRequestFilters(List<HttpConnectionRequestFilter> requestFilters){
```
tomblench: Can these go in the base class, `Replication`? They repeat the same code for Push and Pull

rhyshort: No, Well we could, but the subclass would still need an implementation so you can seamless use the builder pattern. So we'd have something like:

```
public abstract Replication {

    public Replication addRequestFilters(HttpConnectionRequestFilter ... filters){
        this.filters.addAll(Arrays.asList(filters));
    }
}

public class PullReplication {

    @Override
     public PullReplication addRequestFilters(HttpConnectionRequestFilter ... filters){
         super.addRequestFilters(filters);
     }
}

```
or we could end up with some really complicated reflection so we could type infer whats going on using `T` it would look something like: 

```java
public abstract Replication<T extends Replication> {

    public T addRequestFilters(HttpConnectionRequestFilter ... filters){
        this.filters.addAll(Arrays.asList(filters));
    }
}

public class PullReplication extends Replication<PullReplication> {

}

```
(I haven't actually tried it, but my research suggests this would work)

I though implementing it completely twice was the lesser of three evils.

rhyshort: Basically, I'd love to have Objc's instancetype right now

sync-core/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -148,7 +148,6 @@ public void setRequestBody(InputStream input, long inputLength) {
         this.input = input;
         this.inputLength = inputLength;
     }
-
```
tomblench: Looks like you have a whitespace diff but this file is otherwise untouched - probably best to revert it?

rhyshort: done in 20349b9

sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
```diff
@@ -81,12 +91,20 @@
...
+                DEFAULT_PULL_ATTACHMENTS_INLINE);
     }
 
     public BasicPullStrategy(PullReplication pullReplication,
                              ExecutorService executorService,
-                             PullConfiguration config) {
+                             int changeLimitPerBatch,
+                             int batchLimitPerRun,
+                             int insertBatchSize,
+                             boolean pullAttachmentsInline ) {
```
mikerhodes: Don't allow publicly changing these from the public interface, in this PR at least.

sync-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java
```diff
@@ -75,23 +83,29 @@
...
+                    DEFAULT_PUSH_ATTACHMENTS_INLINE);
     }
 
-    public BasicPushStrategy(PushReplication pushReplication,
-                             PushConfiguration config) {
+    public BasicPushStrategy (PushReplication pushReplication,
+                              int changeLimitPerBatch,
+                              int batchLimitPerRun,
+                              int insertBatchSize,
+                              PushAttachmentsInline pushAttachmentsInline){
```
mikerhodes: Again, avoid exposing these as configuration options.

sync-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
rhyshort: The class is package visible, so the API isn't public, I can reduce the scope of the constructor to package as well if you prefer

rhyshort:
This has been split into multiple PRs closing

PR #183, opened 2015-07-22
Clean up CouchConfig class

rhyshort:
Clean up CouchConfig class by removing options that were not honoured.

Split out from PR #181 

reviewer @mikerhodes 
reviewer @tomblench

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 - [BREAKING CHANGE] Moved code for encryption on Android to the
   `sync-android-encryption` subproject. See the instructions in the [README](https://github.com/cloudant/sync-android/blob/master/README.md)
   file for how to include `sync-android-encryption` in your project.
+- [BREAKING CHANGE] Removed Configuration options from `CouchConfig` which are not used
+   by the HTTP Layer. See commit [815026c](https://github.com/cloudant/sync-android/commit/815026c14caf86a8cbb105af0b30f7fb73a46871)
+   for details.
 - [FIX] Fixed issue where at least one index had to be created before a query would execute.  
   You can now query for documents without the existence of any indexes.
 - [NEW] New fields `documentsReplicated` and `batchesReplicated` added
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -2,6 +2,8 @@
 - [BREAKING CHANGE] Moved code for encryption on Android to the
   `sync-android-encryption` subproject. See the instructions in the [README](https://github.com/cloudant/sync-android/blob/master/README.md)
   file for how to include `sync-android-encryption` in your project.
+- [BREAKING CHANGE] Removed Configuration options from CouchConfig which are not used
```
mikerhodes: Can you link to the commit here so people can see what changed?

mikerhodes:
Looks generally good -- though how difficult would it be to instead make the options used? Only worth doing, ofc, if the options were public in the first place.

tomblench:
`HTTPUrlConnection` doesn't have as many low level settings, but it's possible to adjust timeouts etc  through that class, and a few other low-level settings through properties: http://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html#MiscHTTP

Perhaps these could be addressed in a separate ticket?

CHANGES.md
rhyshort: done in commit cfd7730

rhyshort:
I'd side on handling it in a separate ticket, that way we can look at which options we can actually provide, for example the connection per host can't be provided until we deal with the keep alive issue.

mikerhodes:
Sounds good -- let's add some of the options we can do (I guess things like number of retries for network failures might go here).

+1 otherwise.

CHANGES.md
tomblench: Configuration and CouchConfig should be in backticks to make it clearer.

'see commit' should either be in its own sentence or in brackets

tomblench: also 'see commit xxx for details' is clearer

tomblench:
+1 after my minor comments addressed

PR #186, opened 2015-07-23
Expose HttpConnectionFilters

rhyshort:
Expose HttpConnectionFilters via the ReplicatorBuilder.
Replication.createCouchConfig is now a final method, to guarantee that
the filters are added to CouchClient.

reviewer @tomblench 
reviewer @mikerhodes

### DIFF
--- a/sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
+++ b/sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
@@ -21,6 +21,8 @@ package com.cloudant.mazha;
 
 import com.cloudant.http.Http;
 import com.cloudant.http.HttpConnection;
+import com.cloudant.http.HttpConnectionRequestFilter;
+import com.cloudant.http.HttpConnectionResponseFilter;
 import com.cloudant.mazha.json.JSONHelper;
 import com.cloudant.sync.datastore.MultipartAttachmentWriter;
 import com.fasterxml.jackson.core.type.TypeReference;
@@ -32,6 +34,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.lang.reflect.Type;
 import java.net.URI;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
@@ -44,11 +47,23 @@ public class CouchClient  {
     protected final JSONHelper jsonHelper;
     private final Map<String, String> customHeaders;
     private CouchURIHelper uriHelper;
+    private List<HttpConnectionRequestFilter> requestFilters;
+    private List<HttpConnectionResponseFilter> responseFilters;
 
     public CouchClient(CouchConfig config) {
         this.jsonHelper = new JSONHelper();
         this.uriHelper = new CouchURIHelper(config.getRootUri());
         this.customHeaders = config.getCustomHeaders();
+        this.requestFilters = new ArrayList<HttpConnectionRequestFilter>();
+        this.responseFilters = new ArrayList<HttpConnectionResponseFilter>();
+
+        if(config.getRequestFilters() != null) {
+            this.requestFilters.addAll(config.getRequestFilters());
+        }
+
+        if(config.getResponseFilters() != null) {
+            this.responseFilters.addAll(config.getResponseFilters());
+        }
     }
 
     public URI getRootUri() {
@@ -76,6 +91,8 @@ public class CouchClient  {
         // all CouchClient requests want to receive application/json responses
         connection.requestProperties.put("Accept", "application/json");
         this.addCustomHeaders(connection);
+        connection.responseFilters.addAll(this.responseFilters);
+        connection.requestFilters.addAll(this.requestFilters);
         InputStream is = null; // input stream - response from server on success
         InputStream es = null; // error stream - response from server for a 500 etc
         String response = null;
--- a/sync-core/src/main/java/com/cloudant/sync/replication/Replication.java
+++ b/sync-core/src/main/java/com/cloudant/sync/replication/Replication.java
@@ -14,6 +14,8 @@
 
 package com.cloudant.sync.replication;
 
+import com.cloudant.http.HttpConnectionRequestFilter;
+import com.cloudant.http.HttpConnectionResponseFilter;
 import com.cloudant.mazha.CouchConfig;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
@@ -40,6 +42,12 @@ import java.util.Map;
  */
 @Deprecated
 public abstract class Replication {
+
+    final List<HttpConnectionRequestFilter> requestFilters = new ArrayList
+            <HttpConnectionRequestFilter>();
+    final List<HttpConnectionResponseFilter> responseFilters = new ArrayList
+            <HttpConnectionResponseFilter>();
+
 
     /**
      * <p>Provides the name and parameters for a filter function to be used
@@ -179,8 +187,11 @@ public abstract class Replication {
      */
     abstract ReplicationStrategy createReplicationStrategy();
 
-    CouchConfig createCouchConfig(URI uri) {
-        return new CouchConfig(uri);
+    final CouchConfig createCouchConfig(URI uri) {
+        CouchConfig config = new CouchConfig(uri);
+        config.setResponseFilters(responseFilters);
+        config.setRequestFilters(requestFilters);
+        return config;
     }
 
     void checkURI(URI uri) {
--- a/sync-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
+++ b/sync-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
@@ -14,9 +14,14 @@
 
 package com.cloudant.sync.replication;
 
+import com.cloudant.http.HttpConnectionRequestFilter;
+import com.cloudant.http.HttpConnectionResponseFilter;
 import com.cloudant.sync.datastore.Datastore;
 
 import java.net.URI;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 /**
  * A Builder to create a {@link Replicator Object}
@@ -26,6 +31,10 @@ public abstract class ReplicatorBuilder<S, T, E> {
 
     private T target;
     private S source;
+    private List<HttpConnectionRequestFilter> requestFilters = new ArrayList
+            <HttpConnectionRequestFilter>();
+    private List<HttpConnectionResponseFilter> responseFilters = new ArrayList
+            <HttpConnectionResponseFilter>();
 
     /**
      * A Push Replication Builder
@@ -64,6 +73,8 @@ public abstract class ReplicatorBuilder<S, T, E> {
             PullReplication pullReplication = new PullReplication();
             pullReplication.source = super.source;
             pullReplication.target = super.target;
+            pullReplication.responseFilters.addAll(super.responseFilters);
+            pullReplication.requestFilters.addAll(super.requestFilters);
             //convert the new filter to the old one.
             //this is to avoid invasive changes for now.
             Replication.Filter filter = new Replication.Filter(this.pullPullFilter.getName(),
@@ -94,6 +105,7 @@ public abstract class ReplicatorBuilder<S, T, E> {
      */
     public E to(T target) {
         this.target = target;
+        //noinspection unchecked
         return (E) this;
     }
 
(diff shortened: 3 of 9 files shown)


sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
```diff
@@ -69,4 +72,52 @@ public void start_StartedThenComplete() throws InterruptedException {
...
+
+        TestReplicationListener listener = new TestReplicationListener();
+        PullReplication pullReplication = createPullReplication();
+        pullReplication.requestFilters.add(new InvalidJSONFilter());
+        Replicator replicator = ReplicatorFactory.oneway(pullReplication);
+        replicator.getEventBus().register(listener);
+        replicator.start();
+
+        while(replicator.getState() != Replicator.State.COMPLETE && replicator.getState() != Replicator.State.ERROR) {
+            Thread.sleep(1000);
```
mikerhodes: Don't sleep for a whole second, why not just 50ms? Or fewer?

sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
```diff
@@ -69,4 +72,52 @@ public void start_StartedThenComplete() throws InterruptedException {
...
+
+        TestReplicationListener listener = new TestReplicationListener();
+        PullReplication pullReplication = createPullReplication();
+        pullReplication.responseFilters.add(new ResponseStreamReaderFilter());
+        Replicator replicator = ReplicatorFactory.oneway(pullReplication);
+        replicator.getEventBus().register(listener);
+        replicator.start();
+
+        while(replicator.getState() != Replicator.State.COMPLETE && replicator.getState() != Replicator.State.ERROR) {
+            Thread.sleep(1000);
```
mikerhodes: Again, a long pause which could be shorter?

sync-core/src/test/java/com/cloudant/sync/replication/ResponseStreamReaderFilter.java
```diff
@@ -0,0 +1,42 @@
...
+
+
+        HttpConnection connection = context.connection;
+        try {
+            connection.responseAsString();
+        } catch (IOException e){
+            e.printStackTrace();
+        }
+
+
```
mikerhodes: Excess new lines.

sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
rhyshort: I was following the lead of the other tests which wait for replications to complete, we can probably lower the wait time, I'll investigate

rhyshort: tests still work fine I made the change in 6fe42ac

sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
rhyshort: done in 6fe42ac

sync-core/src/test/java/com/cloudant/sync/replication/ResponseStreamReaderFilter.java
rhyshort: done in 68a83bd

sync-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
```diff
@@ -109,6 +120,44 @@ public E from(S source) {
...
+    }
+
+    /**
+     * Adds filters to the list of request filters to use for each request made by this replication.
+     * @param filters The request filters to add.
+     * @return The current instance of {@link ReplicatorBuilder}
+     */
+    public E addRequestFilters(List<HttpConnectionRequestFilter> filters){
+        this.requestFilters.addAll(filters);
+        return (E)this;
```
mikerhodes: Intellij issues warnings for these casts, which I don't think we can avoid. Maybe add statement level ignores (`//noinspection unchecked`)?

rhyshort: done in 9febb5c

sync-core/src/test/java/com/cloudant/sync/replication/ResponseStreamReaderFilter.java
```diff
@@ -0,0 +1,40 @@
...
+import com.cloudant.http.HttpConnectionFilterContext;
+import com.cloudant.http.HttpConnectionResponseFilter;
+
+import java.io.IOException;
+
+/**
+ * This response filter reads the data stream from the server in order to
+ * force an IO error when Jackson attempts to read the stream to handle the content.
+ */
+public class ResponseStreamReaderFilter implements HttpConnectionResponseFilter {
```
tomblench: I've had a think about this and the only legitimate thing a `HttpConnectionResponseFilter` can currently do is set the `replayRequest` flag to true. This may change in the future but it will require some work to enable it (things like rewriting/replacing the response stream are tricky to get right).

So I think our tests should stick to doing legitimate things which we would expect the users to do.

Also we have talked before about documenting exactly what the user can and can't do in the filters - we should make it explicit that setting the `replayRequest` is pretty much the only thing they can do (for now).

sync-core/src/test/java/com/cloudant/sync/replication/FilterCallCounter.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package com.cloudant.sync.replication;
+
+import com.cloudant.http.HttpConnectionFilterContext;
+import com.cloudant.http.HttpConnectionRequestFilter;
+import com.cloudant.http.HttpConnectionResponseFilter;
+
+/**
+ * This request filter replaces the request body with an invalid json body. To cause
```
tomblench: Comment is now out of date

tomblench:
OK +1

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -44,11 +46,15 @@
...
     private CouchURIHelper uriHelper;
+    private List<HttpConnectionRequestFilter> requestFilters;
+    private List<HttpConnectionResponseFilter> responseFilters;
 
     public CouchClient(CouchConfig config) {
         this.jsonHelper = new JSONHelper();
         this.uriHelper = new CouchURIHelper(config.getRootUri());
         this.customHeaders = config.getCustomHeaders();
+        this.requestFilters = config.getRequestFilters();
+        this.responseFilters = config.getResponseFilters();
```
mikerhodes: Create new lists rather than assigning them.

sync-core/src/main/java/com/cloudant/sync/replication/Replication.java
```diff
@@ -41,6 +43,12 @@
 @Deprecated
 public abstract class Replication {
 
+    final List<HttpConnectionRequestFilter> requestFilters = new ArrayList
+            <HttpConnectionRequestFilter>();
+    final List<HttpConnectionResponseFilter> responseFilters = new ArrayList
+            <HttpConnectionResponseFilter>();
```
mikerhodes: I'd be tempted to make all these public request/responseFilter variables (in all the various classes) private, and then just have `addRequestFilters`, `addResponseFilters` methods in each class. This is breaking encapsulation, to my mind.

sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
```diff
@@ -69,4 +70,47 @@ public void start_StartedThenComplete() throws InterruptedException {
...
+        replicator.getEventBus().register(listener);
+        replicator.start();
+
+        while(replicator.getState() != Replicator.State.COMPLETE && replicator.getState() != Replicator.State.ERROR) {
+            Thread.sleep(50);
+        }
+
+        Assert.assertEquals(Replicator.State.COMPLETE, replicator.getState());
+        Assert.assertFalse(listener.errorCalled);
+        Assert.assertTrue(listener.finishCalled);
```
mikerhodes: Lines 81 to 91 beg to be pulled into a `runReplicationUntilComplete(Replication replication)` method.

sync-core/src/test/java/com/cloudant/sync/replication/FilterCallCounter.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.cloudant.http.HttpConnectionFilterContext;
+import com.cloudant.http.HttpConnectionRequestFilter;
+import com.cloudant.http.HttpConnectionResponseFilter;
+
+/**
+ * This request and response filter counts the number of times the filterRequest and filterResponse
+ * methods get called.
+ */
+public class FilterCallCounter implements HttpConnectionRequestFilter, HttpConnectionResponseFilter{
```
mikerhodes: Space before `{`.

sync-core/src/test/java/com/cloudant/sync/replication/FilterCallCounter.java
```diff
@@ -0,0 +1,42 @@
...
+
+
+    public int filterRequestTimesCalled = 0;
```
mikerhodes: Extra newline.

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -120,6 +127,15 @@ private InputStream executeToInputStream(HttpConnection connection) throws Couch
         }
     }
 
+    private void addFilters(HttpConnection connection) {
+        if (this.responseFilters != null) {
+            connection.responseFilters.addAll(this.responseFilters);
+        }
+        if (this.requestFilters != null) {
+            connection.requestFilters.addAll(this.requestFilters);
+        }
```
mikerhodes: If you create lists in the constructor you don't need the `if` checks.

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
rhyshort: done in 38016a0

sync-core/src/main/java/com/cloudant/sync/replication/Replication.java
rhyshort: I think it does too. However that being said I implemented it this way because the subclasses use public instance variables to interact with the source and target for the replication so I reasoned doing it this way is more in keeping with what we already had, which I think makes the classes less confusing

sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
rhyshort: done in 4cdc4b3

sync-core/src/test/java/com/cloudant/sync/replication/FilterCallCounter.java
rhyshort: done in c9a86ca

sync-core/src/test/java/com/cloudant/sync/replication/FilterCallCounter.java
rhyshort: done in 3e68c42

sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
rhyshort: done in 5a3c9e3

sync-core/src/main/java/com/cloudant/sync/replication/Replication.java
mikerhodes: I guess we can fix this up when the Replication is removed.

sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
mikerhodes: Can `runReplicationUntilComplete` be extended to a utility method used by PushReplicatorTest too? The method can be push/pull agnostic.

rhyshort: I think so. I think maybe this should be pushed up to the replication base class rather than be in a separate test utils one

rhyshort: done in 9e4b641

mikerhodes:
We're probably good to go here now, we can fix up the internals if we so desire later.

+1, good to squash.

PR #188, opened 2015-07-24
Remove some of the unused CouchClient methods

tomblench:
- What: Remove unused and unnecessary `CouchClient` methods

- Why: More preperation for the 49297 'HTTP connection resilience'. This helps us to identify which methods need to be wrapped in a `retry` loop by ensuring that each method has a clear purpose and that there are not multiple overloaded methods which do very similar tasks. Also some methods were only ever called by test code.

- Testing: One test which is no longer relevant has been removed: `getDocument_docDeleted_exception`. This fetched a document without specifying a revid which never happens in production code. All tests pass.

reviewer: @mikerhodes 
reviewer: @rhyshort

### DIFF
--- a/sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
+++ b/sync-core/src/main/java/com/cloudant/mazha/CouchClient.java
@@ -221,13 +221,6 @@ public class CouchClient  {
         }
     }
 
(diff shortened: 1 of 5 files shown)


rhyshort:
+1

mikerhodes:
+!

PR #192, opened 2015-08-17
Fix NullPointerException in ReplicatorBuilder

rhyshort:
## What
Do not attempt to generate a `Replication.Filter` when `PullFilter` is `null` in `ReplictorBuilder.`

## How 
Check if the `PullFilter` is `null` before attempting to generate the `Replication.Filter ` used by the Replicator.

## Why
Currently if a PullReplication is used without setting a filter, the application will crash.

## Testing
To test, build the sample application using the latest release. (it crashes during start up).
Apply patch and run again, replication should be able to complete successfully and the app,
launches.

## Reviewers

reviewer @tomblench 
reviewer @mikerhodes 

BugzId: 50635

### DIFF
--- a/sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
+++ b/sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
@@ -23,6 +23,8 @@ import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
 import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
 
 @Category(RequireRunningCouchDB.class)
 public class PullReplicatorTest extends ReplicationTestBase {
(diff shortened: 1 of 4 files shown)


mikerhodes:
+1.

sync-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
```diff
@@ -92,4 +92,27 @@ public void testResponseInterceptors() throws Exception {
...
+        Assert.assertNotNull(replicator);
+    }
+
+    @Test
+    public void testPullReplicationCreatedSuccessfullyWitFilter() throws Exception {
+
+        Replicator replicator = ReplicatorBuilder.pull()
+                .from(this.source)
+                .to(this.datastore)
+                .filter(new PullFilter("a_filter"))
```
tomblench: Can we test with and without filter params?

rhyshort: done in dcb6392

tomblench:
+1

PR #193, opened 2015-08-17
Fix gradle install artifacts

tomblench:
Rather than customise the artifactId in two places for every sub-project (https://docs.gradle.org/current/userguide/maven_plugin.html#sec:maven_pom_generation and https://docs.gradle.org/current/userguide/publishing_maven.html#N17DA8), change the sub-project names so that the default values are correct.


rhyshort:
+1

ricellis:
+1

PR #194, opened 2015-08-19
Fix encoding for document IDs containing colons

tomblench:
- Introduce `HierarchicalUriComponents` for URI encoding, based on
  code from the Spring Framework. See file for full attribution.
- Add tests for replication of documents containing the colon (`:`)
  character. Previously this was incorrectly handled.
- Update tests to reflect encoding of `[` and `]`
  characters. Previously these were not encoded. RFC 3986 suggests
  that it is correct to encode these characters.


mikerhodes:
LGTM +1

tomblench:
@ricellis updated to use UTF-8

ricellis:
+1

PR #195, opened 2015-09-07
48607 replication policies

brynh:
_What?_

_Why?_

_How?_

reviewer @mikerhodes


REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,292 @@
+Replication policies work quite differently on Android from Java. Please see the relevant section of this guide 
+for [Android](#android-replication-policies) or [Java](#java-replication-policies).
+
+# Android replication policies
+Replication policies on Android run in a `Service` to allow them to run independently of your main application and to allow
+them to be restarted if they are killed by the operating system.  To use replication policies, you need to create a subclass
```
ricellis: Minor grammar issue here
>you need to create a subclass one of

I think needs to be either:
* you need to create a subclass of one of
* you need to subclass one of

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,292 @@
...
+one of the existing replication Service components and implement the specifics of your required replications.
+
+##Creating your service
+There are two options when creating your service. 
+
+1. If you don't require periodic replications, you should create a subclass of `ReplicationService`.
+2. If you want your replications to repeat at (roughly) fixed time intervals, as well as possibly adding other criteria that
+are required for replication, you should create a subclass of `PeriodicReplicationService`. This class takes care of the
+scheduling of periodic alarms to trigger the replications. This also requires you to implement a `BroadcastReceiver` that is a subclass of `PeriodicReplicationReceiver`.
```
ricellis: Assuming `BroadcastReceiver` is an Android class I think a FQN would be useful here. Although the IDE will probably help people end up with the correct class it would be nice to distinguish it from our own library classes.

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,292 @@
...
+
+If you want to start replications from anywhere other than a `PeriodicReplicationReceiver` you would need to replace
+`startWakefulService(context, intent);` with `startService(intent);` in the above example and do any `WakeLock`
+management you require yourself.
+
+###WifiPeriodicReplicationReceiver
+This class is an example of how to extend `PeriodicReplicationReceiver` to add logic that triggers periodic replications
+only when our device is connected to Wifi and stops periodic replications when we disconnect from Wifi.
+This is done by extending the `PeriodicReplicationReceiver` so that the BroadcastReceiver responds to changes in
+network connectivity.  When we detect that the device has connected to Wifi we start our PeriodicReplicationService with
```
ricellis: ~~with~~

AndroidTest/app/build.gradle
```diff
@@ -64,14 +64,14 @@ dependencies {
...
-    compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.3.2'
+    compile 'com.fasterxml.jackson.core:jackson-databind:2.1.1'
+    compile 'com.google.guava:guava:15.0'
+    compile 'commons-io:commons-io:2.4'
+    compile 'commons-codec:commons-codec:1.9'
+    compile 'org.apache.commons:commons-lang3:3.3.2'
     compile 'com.google.code.findbugs:jsr305:3.0.0' //this is needed for some versions of android
     compile files('../../cloudant-sync-datastore-android-encryption/libs/sqlcipher.jar') //required sqlcipher lib
-
+    compile files('../../cloudant-sync-datastore-android/libs/android-support-v4.jar')
```
ricellis: Is it not possible to get this dependency from a repository?

brynh:
This PR has been split into multiple PRs to make reviewing easier, so this PR is being closed without merging. See PRs #196 - #201.

PR #196, opened 2015-09-08
Add ID to Replicator to enable replicators to be identified in callba…

brynh:
This is part 1 of a 6 part change to implement replication policies. Each of the 6 parts of the implementation of replication policies are each in a separate branch.

The 6 branches include:

_What?_

_Why?_

_How?_

reviewer @mikerhodes
reviewer @ricellis

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicReplicator.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicReplicator.java
@@ -22,9 +22,11 @@ import com.google.common.eventbus.Subscribe;
 
 class BasicReplicator implements Replicator {
 
+    public static final int NULL_ID = -1;
     protected final Replication replication;
     protected Thread strategyThread;
     protected ReplicationStrategy strategy;
+    protected int id = NULL_ID;
 
     // Writes need synchronising.
     private State state = null;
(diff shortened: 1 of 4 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicReplicator.java
```diff
@@ -36,6 +38,12 @@ public BasicReplicator(Replication replication) {
         this.state = State.PENDING;
```
ricellis: Any reason not to do: `this(replication, NULL_ID)`?

brynh: Fixed in e2d375f.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
```diff
@@ -129,6 +129,12 @@ public E from(S source) {
         return (E) this;
     }
 
+    public E withId(int id) {
+        this.id = id;
+        //noinspection unchecked
```
ricellis: There is a common `returnThis()` pattern that can avoid this unchecked cast. I don't this use warrants a change, just mentioning it for future reference.

ricellis:
+1

mikerhodes:
+1

brynh:

PR #197, opened 2015-09-08
Add replication policies for Java.

brynh:
This is part 2 of a 6 part change to implement replication policies. Each of the 6 parts of the implementation of replication policies are each in a separate branch.

The 6 branches include:

_What?_

_Why?_

_How?_

reviewer @mikerhodes
reviewer @ricellis

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
@@ -83,28 +83,34 @@ public abstract class ReplicationPolicyManager {
     public abstract void stop();
 
     protected void startReplications() {
-        for (Replicator replicator : replicators) {
-            if (!replicationListener.inProgress(replicator)) {
-                replicationListener.add(replicator);
-                replicator.getEventBus().register(replicationListener);
+        synchronized (replicators) {
+            for (Replicator replicator : replicators) {
+                if (!replicationListener.inProgress(replicator)) {
+                    replicationListener.add(replicator);
+                    replicator.getEventBus().register(replicationListener);
 
-                replicator.start();
+                    replicator.start();
+                }
             }
         }
     }
 
     protected void stopReplications() {
-        for (Replicator replicator : replicators) {
-            replicationListener.remove(replicator);
-            replicator.stop();
+        synchronized (replicators) {
+            for (Replicator replicator : replicators) {
+                replicationListener.remove(replicator);
+                replicator.stop();
+            }
         }
     }
 
     public void addReplicators(Replicator... replicators) {
-        if (this.replicators == null) {
-            this.replicators = new ArrayList<Replicator>();
+        synchronized (this.replicators) {
+            if (this.replicators == null) {
+                this.replicators = new ArrayList<Replicator>();
+            }
+            this.replicators.addAll(Arrays.asList(replicators));
         }
-        this.replicators.addAll(Arrays.asList(replicators));
     }
 
     public void setReplicationsCompletedListener(ReplicationsCompletedListener listener) {
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
@@ -1,0 +1,110 @@
+package com.cloudant.sync.replication;
+
+import com.cloudant.sync.notifications.ReplicationCompleted;
+import com.cloudant.sync.notifications.ReplicationErrored;
+import com.google.common.eventbus.Subscribe;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public abstract class ReplicationPolicyManager {
+
+    private List<Replicator> replicators;
+    private ReplicationListener replicationListener;
+    private ReplicationsCompletedListener mReplicationsCompletedListener;
+
+    public interface ReplicationsCompletedListener {
+        void allReplicationsCompleted();
+
+        void replicationCompleted(int id);
+
+        void replicationErrored(int id);
+    }
+
+    private class ReplicationListener {
+
+        Set<Replicator> replicatorsInProgress;
+
+        ReplicationListener() {
+            replicatorsInProgress = new HashSet<Replicator>();
+        }
+
+        synchronized void add(Replicator replicator) {
+            replicatorsInProgress.add(replicator);
+        }
+
+        synchronized void remove(Replicator replicator) {
+            if (replicatorsInProgress.remove(replicator)) {
+                replicator.getEventBus().unregister(this);
+            }
+        }
+
+        boolean inProgress(Replicator replicator) {
+            return replicatorsInProgress.contains(replicator);
+        }
+
+        @Subscribe
+        public void complete(ReplicationCompleted event) {
+            finishedReplication(event.replicator);
+            if (mReplicationsCompletedListener != null) {
+                mReplicationsCompletedListener.replicationCompleted(event.replicator.getId());
+            }
+        }
+
+        @Subscribe
+        public void error(ReplicationErrored event) {
+            finishedReplication(event.replicator);
+            if (mReplicationsCompletedListener != null) {
+                mReplicationsCompletedListener.replicationErrored(event.replicator.getId());
+            }
+        }
+
+        public void finishedReplication(Replicator replicator) {
+            remove(replicator);
+            if (replicatorsInProgress.size() == 0 && mReplicationsCompletedListener != null) {
+                mReplicationsCompletedListener.allReplicationsCompleted();
+            }
+        }
+    }
+
+    public ReplicationPolicyManager() {
+        replicators = new ArrayList<Replicator>();
+        replicationListener = new ReplicationListener();
+    }
+
+    public abstract void start();
+
+    public abstract void stop();
+
+    protected void startReplications() {
+        for (Replicator replicator : replicators) {
+            if (!replicationListener.inProgress(replicator)) {
+                replicationListener.add(replicator);
+                replicator.getEventBus().register(replicationListener);
+
+                replicator.start();
+            }
+        }
+    }
+
+    protected void stopReplications() {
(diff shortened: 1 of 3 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
```diff
@@ -0,0 +1,110 @@
...
+
+                replicator.start();
+            }
+        }
+    }
+
+    protected void stopReplications() {
+        for (Replicator replicator : replicators) {
+            replicationListener.remove(replicator);
+            replicator.stop();
```
ricellis: Is there a timing window here where the replicator is removed from the listener, but then completes before `stop()` is called?
Isn't it possible that we'd miss a finished event in that case?

ricellis:
I don't think the example code should be packaged with the implementation, it might be nice to have it in a separate package.

ricellis:
I'm a bit concerned about the lack of synchronization on the collections used in the `ReplicationPolicyManager` and `ReplicationListener` particularly because the `inProgress()` check is based on the replicator being contained in a list. I might be missing something, but it seems to me that the addition and removal of replicators will almost inevitably happen on separate threads to a `Timer` execution of `startReplications()` or `stopReplications()`.

brynh:
> I'm a bit concerned about the lack of synchronization on the collections used in the ReplicationPolicyManager and ReplicationListener particularly because the inProgress() check is based on the replicator being contained in a list. I might be missing something, but it seems to me that the addition and removal of replicators will almost inevitably happen on separate threads to a Timer execution of startReplications() or stopReplications().

Added synchronisation of `replicatorsInProgress` in 11bc9fd.

brynh:
> I don't think the example code should be packaged with the implementation, it might be nice to have it in a separate package.

The comments describe the `IntervalTimerReplicationPolicyManager` as a sample policy, but should probably just describe it as a policy. This isn't really example code - it is an actual policy. Updated comments in 4777e38.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
```diff
@@ -0,0 +1,113 @@
...
+                mReplicationsCompletedListener.replicationErrored(event.replicator.getId());
+            }
+        }
+
+        public void finishedReplication(Replicator replicator) {
+            synchronized (replicator) {
+                remove(replicator);
+                if (replicatorsInProgress.size() == 0 && mReplicationsCompletedListener != null) {
+                    mReplicationsCompletedListener.allReplicationsCompleted();
+                }
```
mikerhodes: I don't understand how having this block synchronized on a given replicator is useful -- could you explain? It seems like you could get the event on multiple replicators concurrently, which might be problematic as you'll be going through this concurrently.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
```diff
@@ -0,0 +1,113 @@
...
+import com.google.common.eventbus.Subscribe;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public abstract class ReplicationPolicyManager {
```
mikerhodes: This class should probably have a note somewhere that it isn't threadsafe.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/ReplicationPolicyManagerTest.java
```diff
@@ -0,0 +1,349 @@
...
+        when(mockReplicator.getEventBus()).thenReturn(mockEventBus);
+        ReplicationPolicyManager rpm = new TestPolicy();
+        rpm.addReplicators(mockReplicator);
+        verify(mockReplicator, never()).start();
+        rpm.start();
+        verify(mockReplicator, times(1)).start();
+    }
+
+    /**
+     * Check that when the {@link }ReplicationPolicyManager} is setup with multiple
```
mikerhodes: `{@link }` extra `}`

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/ReplicationPolicyManagerTest.java
```diff
@@ -0,0 +1,349 @@
...
+        verify(mockReplicator2, never()).start();
+        verify(mockReplicator3, never()).start();
+        rpm.start();
+        mockEventBus1.post(new ReplicationCompleted(mockReplicator1, 1, 1));
+        mockEventBus2.post(new ReplicationCompleted(mockReplicator2, 1, 1));
+        mockEventBus3.post(new ReplicationCompleted(mockReplicator3, 1, 1));
+        rpm.start();
+        verify(mockReplicator1, times(2)).start();
+        verify(mockReplicator2, times(2)).start();
+        verify(mockReplicator3, times(2)).start();
```
mikerhodes: Perhaps add a test where if you don't emit the complete event for one of the replicators, that doesn't have start called again? Similarly for the error test below.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/ReplicationPolicyManagerTest.java
brynh: Fixed in 44e8988.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
brynh: This was a mistake - it should have been synchronized on `replicatorsInProgress`. Fixed in 56c2d35.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/ReplicationPolicyManagerTest.java
brynh: Tests added in a193c8a.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
brynh: Made thread safe in 8b30015.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
```diff
@@ -0,0 +1,119 @@
...
+            for (Replicator replicator : replicators) {
+                replicationListener.remove(replicator);
+                replicator.stop();
+            }
+        }
+    }
+
+    public void addReplicators(Replicator... replicators) {
+        synchronized (this.replicators) {
+            if (this.replicators == null) {
```
ricellis: is this block necessary? Line 77 in the constructor does `replicators = new ArrayList<Replicator>();` and I don't see the object being set back to `null` anywhere else.
Maybe just do `private final List<Replicator> replicators = new ArrayList<Replicator>();` if it isn't expected to change. Are there any concerns about this list growing, I don't see any removal of the `Replicator` elements, presumably if one wants to change the `Replicator`s associated with a policy the approach is to create a new `ReplicationPolicyManager` instance.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+    }
+
+    public ReplicationPolicyManager() {
+        replicators = new ArrayList<Replicator>();
+        replicationListener = new ReplicationListener();
+    }
+
+    public abstract void start();
```
brynh: Abstract methods `start()` and `stop()` now only serve to enforce an unnecessary naming convention on how to start/stop the policy. These should be removed and the class made concrete.

brynh: Done in 1664f7f.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
brynh: Good point, plus it can never be true anyway as if `this.replicators == null` there would be a NPE on the previous line.  Fixed in 72fbc91.  My intention was that you'd create a new `ReplicationPolicyManager` if you want to change the `Replicator`s associated with a policy, so I don't see the list growing as an issue. I could easily add ability to clear all and/or remove individual replicators if you have a preference for that.

ricellis: I don't know how frequently people might want to alter the set of replicators associated with a policy. I think it is ok for now, we could always add them later if there is demand.

ricellis:
+1

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
```diff
@@ -0,0 +1,115 @@
...
+    private class ReplicationListener {
+
+        Set<Replicator> replicatorsInProgress;
+
+        ReplicationListener() {
+            replicatorsInProgress = Collections.synchronizedSet(new HashSet<Replicator>());
+        }
+
+        void add(Replicator replicator) {
+            replicatorsInProgress.add(replicator);
```
mikerhodes: Possibly use just `synchronized(replicatorsInProgress)` and avoid `synchronizedSet`?  Though the documentation for the `synchronizedSet` call suggests you can use `synchronized` on a `synchronizedSet` returned set.

brynh: Done in 4c75bb0.

mikerhodes:
+1

brynh:

PR #198, opened 2015-09-08
Add simple Android replication Service

brynh:
This is part 3 of a 6 part change to implement replication policies. Each of the 6 parts of the implementation of replication policies are each in a separate branch.

The 6 branches include:

_What?_

_Why?_

_How?_

reviewer @mikerhodes
reviewer @ricellis


AndroidTest/app/build.gradle
```diff
@@ -64,14 +64,14 @@ dependencies {
...
-    compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.3.2'
+    compile 'com.fasterxml.jackson.core:jackson-databind:2.1.1'
+    compile 'com.google.guava:guava:15.0'
+    compile 'commons-io:commons-io:2.4'
+    compile 'commons-codec:commons-codec:1.9'
+    compile 'org.apache.commons:commons-lang3:3.3.2'
     compile 'com.google.code.findbugs:jsr305:3.0.0' //this is needed for some versions of android
     compile files('../../cloudant-sync-datastore-android-encryption/libs/sqlcipher.jar') //required sqlcipher lib
-
+    compile files('../../cloudant-sync-datastore-android/libs/android-support-v4.jar')
```
ricellis: Can this be obtained from a repository like the rest of the dependencies?

ricellis:
It is always upsetting to see test specific code in classes. If the notifier is useful for automated testing are there other circumstances where it might be useful to an application? Would it be worth making a real interface/plug-point for that functionality?

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    @Override
+    public IBinder onBind(Intent intent) {
+        return mBinder;
+    }
+
+    @Override
+    public boolean onUnbind(Intent intent) {
+        // Ensure onRebind is called when new clients bind to the service.
```
ricellis: Does returning true here ensure `onRebind` is called?

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
```diff
@@ -0,0 +1,319 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * This abstract class forms the basis for creating replication policies on Android.
+ * The replications run in a {@link Service} so that they can properly manage the
+ * lifecycle and handle being killed or restarted by the operating system
+ */
+public abstract class ReplicationService extends Service implements ReplicationPolicyManager.ReplicationsCompletedListener {
```
ricellis: Looks like there are some long lines here, has the code formatter been applied?

brynh: Fixed in 4c7c308.

AndroidTest/app/build.gradle
brynh: Sadly, I don't believe it is oficially uploaded to any repository, so AFAIK it can't.

ricellis: [This page](https://developer.android.com/tools/support-library/features.html) suggests that `com.android.support:support-v4:21.0.0` style dependencies can be used. I think the caveat is that the `Android Support Libraries` or `Android Support Repository` has been downloaded as part of the studio SDK.
I'm guessing that means it probably doesn't work as soon as we try to build outside of studio like in Jenkins?
@rhyshort any thoughts?

brynh:
> It is always upsetting to see test specific code in classes. If the notifier is useful for automated testing are there other circumstances where it might be useful to an application? Would it be worth making a real interface/plug-point for that functionality?

I agree that it's not nice having the test specific code here and was reluctant to do it, but (other than having a timed wait and hoping that the operation had completed, which I dislike even more) I couldn't think of a better way. I can't really think how it would be useful to an application as it's only notifying that replications have been either kicked off or the stop has been requested, not that replications are complete or the stop is complete, so I'm not convinced making it a real part of the API makes sense.

AndroidTest/app/build.gradle
brynh: Yes, it can be used from a local repository, but I think as you say the issue is that building outside Android Studio then becomes an issue. Unfortunately it doesn't seem to be uploaded to any official public repository.

ricellis:
What about broadcasting additional `Intent`s for the kick-off and stop? The tests could consume these broadcasts via a mock.  Someone might even find a valid use case for consuming them.
Presumably android is reasonably well optimised for broadcast/receive of intents, but I'm not sure how much overhead this would add, obviously the current no-op method call is very light. it hopefully wouldn't be that much of an increased performance hit over the no-op call

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
```diff
@@ -0,0 +1,323 @@
...
+                case COMMAND_STOP_REPLICATION:
+                    stopReplications();
+                    notifyTestOperationComplete();
+                    break;
+            }
+
+            // Get the Intent used to start the service and release the WakeLock if there is one.
+            // Calling completeWakefulIntent is safe even if there is no wakelock held.
+            Intent intent = msg.getData().getParcelable(EXTRA_INTENT);
+            WakefulBroadcastReceiver.completeWakefulIntent(intent);
```
ricellis: I wonder if this should be in a `finally` with the other parts of the method in a `try` to ensure that the `WakeLock` is released even if there is some kind of `RuntimeException` from any of the starting and stopping code.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
```diff
@@ -0,0 +1,323 @@
...
+     */
+    public void setReplicationPolicyManager(ReplicationPolicyManager replicationPolicyManager) {
+        mReplicationPolicyManager = replicationPolicyManager;
+    }
+
+    /**
+     * Start the set of replications specified in the set of {@link Replicator}s returned by
+     * {@link #getReplicators(Context)}.
+     */
+    protected void startReplications() {
```
mikerhodes: Does this always run on the same thread? If not, it would appear we could lose the reference to a held wifi lock.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
```diff
@@ -0,0 +1,323 @@
...
+                    }
+
+                    @Override
+                    public void stop() {
+                    }
+                };
+            }
+            mReplicationPolicyManager.addReplicators(replicators);
+            mReplicationPolicyManager.setReplicationsCompletedListener(this);
+            mReplicationPolicyManager.start();
```
mikerhodes: Could you explain why this conditionless policy manager is created, and how it interacts with the `setReplicationPolicyManager` below? (e.g., should `setReplicationPolicyManager` call `stop` on any existing policy manager before re-assigning?)

mikerhodes:
Aside from the conditionless policy manager, my question is whether this class needs to be thread-safe or not. I don't understand enough of the semantics around services to know that.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
brynh: Seems like a sensible precaution. Added in ccfaec0.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
brynh: Fixed in 761c29d.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
brynh: The conditionless policy manager is used to take care of the general replication policy management - e.g. ensuring that replications are not restarted if they are already in progress.  On Android, the main business logic of the replication policy is quite likely to be in a BroadcastReceiver in many cases rather than in the `start()` or `stop()` methods of `ReplicationPolicyManager`.

I'd originally intended that on Java you'd set up the policy in `start()` and stop it in `stop()` and the actual replications would be started by when the policy calls `startReplications()` and stopped when it calls `stopReplications()` (see `IntervalTimerPolicyManager`). I think at one point in the evolution of the code I probably had the framework calling the abstract `start()` and `stop()` methods to start/stop the policy, but now that no longer happens the abstract methods are pretty redundant and only serve to force an unnecessary naming convention for how to start or stop the policy on Java. I think, therefore, it maybe a good idea to remove the `start()` and `stop()` methods from `ReplicationPolicyManager` and make the class concrete.

Fixed in da16267.

brynh:
> Aside from the conditionless policy manager, my question is whether this class needs to be thread-safe or not. I don't understand enough of the semantics around services to know that.
> 
> After reading up, it looks like your worker thread is going to handle all the communications, so things should always be executed in that thread from intents, but I'm not sure about the completion handlers from the policy nor the add/removeListener code.

Access to the `mListeners` set needs to be made thread safe as listeners can be added/removed from different threads. Fixed in 86400d8.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
```diff
@@ -0,0 +1,319 @@
...
+    public static final String EXTRA_INTENT = "intent";
+    public static final String EXTRA_COMMAND = "command";
+    public static final int COMMAND_NONE = -1;
+    public static final int COMMAND_START_REPLICATION = 0;
+    public static final int COMMAND_STOP_REPLICATION = 1;
+
+    private Handler mServiceHandler;
+    private ReplicationPolicyManager mReplicationPolicyManager;
+
+    private final Set<ReplicationCompleteListener> mListeners = new HashSet<ReplicationCompleteListener>();
```
mikerhodes: You should add a note here that all modifications or iterations over `mListeners` should be protected by a synchronized block on the `mListeners` object.

brynh: Done in 3e3f8e7.

brynh:
> What about broadcasting additional Intents for the kick-off and stop? The tests could consume these broadcasts via a mock. Someone might even find a valid use case for consuming them.
> Presumably android is reasonably well optimised for broadcast/receive of intents, but I'm not sure how much overhead this would add, obviously the current no-op method call is very light. it hopefully wouldn't be that much of an increased performance hit over the no-op call

As discussed, I've now instead made the callback mechanism a more general way of being notified when an operation has been kicked off within the service rather than having it as something test specific. Done in a187a46.

ricellis:
+1

mikerhodes:
+1

brynh:

PR #199, opened 2015-09-08
Add replications at periodic intervals.

brynh:
This is part 4 of a 6 part change to implement replication policies. Each of the 6 parts of the implementation of replication policies are each in a separate branch.

The 6 branches include:

_What?_

_Why?_

_How?_

reviewer @mikerhodes
reviewer @ricellis

### DIFF
--- a/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationReceiver.java
+++ b/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationReceiver.java
@@ -1,0 +1,46 @@
+package com.cloudant.sync.replication;
+
+import android.content.Context;
+import android.content.Intent;
+import android.support.v4.content.WakefulBroadcastReceiver;
+
+/**
+ * This class implements a {@link WakefulBroadcastReceiver} that handles events related to
+ * periodic replication.  It responds to reboot of the device to trigger the resetting of
+ * the {@link android.app.AlarmManager} used to trigger periodic replication, and it
+ * handles the periodic alarms sent by the {@link android.app.AlarmManager} to trigger
+ * the periodic replications.
+ *
+ * The resetting of periodic alarms after reboot and the handling of the periodic alarms
+ * are delegated to the {@link PeriodicReplicationService} associated with this
+ * {@link android.content.BroadcastReceiver}.
+ *
+ * @param <T> The {@link PeriodicReplicationService} component triggered by this
+ * {@link android.content.BroadcastReceiver}
+ */
+public class PeriodicReplicationReceiver<T extends PeriodicReplicationService> extends WakefulBroadcastReceiver {
+
+    static final String ALARM_ACTION = "com.cloudant.sync.replication.PeriodicReplicationReceiver.Alarm";
+
+    Class<T> clazz;
+
+    protected PeriodicReplicationReceiver(Class<T> clazz) {
+        this.clazz = clazz;
+    }
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        int command = PeriodicReplicationService.COMMAND_NONE;
+        if (ALARM_ACTION.equals(intent.getAction())) {
+            command = PeriodicReplicationService.COMMAND_START_REPLICATION;
+        } else if (Intent.ACTION_BOOT_COMPLETED.equals(intent.getAction())) {
+            command = PeriodicReplicationService.COMMAND_DEVICE_REBOOTED;
+        }
+
+        if (command != PeriodicReplicationService.COMMAND_NONE) {
+            Intent serviceIntent = new Intent(context.getApplicationContext(), clazz);
+            serviceIntent.putExtra(PeriodicReplicationService.EXTRA_COMMAND, command);
+            startWakefulService(context, serviceIntent);
+        }
+    }
+}
--- a/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationReceiver.java
+++ b/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationReceiver.java
@@ -30,17 +30,19 @@ public class PeriodicReplicationReceiver<T extends PeriodicReplicationService> e
 
     @Override
     public void onReceive(Context context, Intent intent) {
-        int command = PeriodicReplicationService.COMMAND_NONE;
-        if (ALARM_ACTION.equals(intent.getAction())) {
-            command = PeriodicReplicationService.COMMAND_START_REPLICATION;
-        } else if (Intent.ACTION_BOOT_COMPLETED.equals(intent.getAction())) {
-            command = PeriodicReplicationService.COMMAND_DEVICE_REBOOTED;
-        }
+        if (intent != null) {
+            int command = PeriodicReplicationService.COMMAND_NONE;
+            if (ALARM_ACTION.equals(intent.getAction())) {
+                command = PeriodicReplicationService.COMMAND_START_REPLICATION;
+            } else if (Intent.ACTION_BOOT_COMPLETED.equals(intent.getAction())) {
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationReceiver.java
```diff
@@ -0,0 +1,46 @@
...
+    Class<T> clazz;
+
+    protected PeriodicReplicationReceiver(Class<T> clazz) {
+        this.clazz = clazz;
+    }
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        int command = PeriodicReplicationService.COMMAND_NONE;
+        if (ALARM_ACTION.equals(intent.getAction())) {
```
ricellis: Do you think it is worth an `intent != null` check here? It was unclear to me whether the framework and filters will always protect us from receiving a `null` intent.

ricellis:
It took me a while to understand why the alarms are not just set and cancelled and thus why the `*ALARM_DUE*` prefs needed to be stored. I think it is so if I start my periodic replication for once every 24 hours then reboot after 8 hours I will only have to wait ~16 hours after the reboot until the replication, is that correct?

The simpler alternative would be to just set the elapsed trigger time to one interval following a reboot. However, I think this feature is worth the additional complexity of needing to use of both clocks and storing the due time, but I just wanted to check that my understanding was correct. Perhaps an additional comment with the an example would make it more clear why the complexity is needed.

ricellis:
I noticed that nearly all the methods in `PeriodicReplicationService` are `protected` and I know the class is designed to be sub-classed, but it seems to me that overriding many of them in a subclass could break the behaviour. For example the alarm time getters and setters should probably be made either `private` or `final` depending on whether access is required in the sub-classes.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationReceiver.java
```diff
@@ -0,0 +1,46 @@
...
+import android.content.Intent;
+import android.support.v4.content.WakefulBroadcastReceiver;
+
+/**
+ * This class implements a {@link WakefulBroadcastReceiver} that handles events related to
+ * periodic replication.  It responds to reboot of the device to trigger the resetting of
+ * the {@link android.app.AlarmManager} used to trigger periodic replication, and it
+ * handles the periodic alarms sent by the {@link android.app.AlarmManager} to trigger
+ * the periodic replications.
+ *
```
ricellis: If this is expected to be read in the generated HTML javadoc then it will need `<P></P>` tags or something.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
```diff
@@ -0,0 +1,300 @@
...
+import android.os.SystemClock;
+import android.util.Log;
+
+/**
+ * This {@link android.app.Service} is an abstract class that is the basis for creating a service
+ * that performs periodic replications (i.e. replications that occur at regular intervals). The
+ * period between replications may be varied depending on whether other application components
+ * are bound to the service or not, so as to allow for more frequent replications when an app
+ * is in active use and less frequent replications the rest of the time.
+ *
```
ricellis: Doc paragraphing again

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
```diff
@@ -0,0 +1,300 @@
...
+            return getUnboundIntervalInSeconds();
+        }
+    }
+
+    /** Update the SharedPreferences to store the time the next alarm is due at using the
+     *  bound time interval between replications.
+     */
+    protected void updateAlarmTimeOnBind() {
+        long dueTime = getNextAlarmDueElapsedTime();
+        long newDueTime = dueTime - (getUnboundIntervalInSeconds() * MILLISECONDS_IN_SECOND) + (getBoundIntervalInSeconds() * MILLISECONDS_IN_SECOND) - SystemClock.elapsedRealtime();
```
ricellis: Line length formatting again

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationReceiver.java
brynh: I think it's worth it. Although I don't think having a null `Intent` makes much sense I can't see there's any guarantee that it can't happen, so probably best to be defensive. Fixed in c12068f.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationReceiver.java
brynh: Fixed in 99a4714.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
brynh: Reformatted long lines in 37753b0.

brynh:
You are correct regarding the reasons for storing the alarms in SharedPreferences. One reason to not trigger at one interval following a reboot would be a scenario where the replications are scheduled for every 24hrs, and someone turns off their phone every night when they go to bed and switches it back on when they wake up - then it would never trigger, so I think the more complicated mechanism is necessary.  I've added some comments in e045445, which hopefully clarify things a bit.

brynh:
> I noticed that nearly all the methods in PeriodicReplicationService are protected...

Agreed. Made them `private` or `final` as appropriate in 0271e02.

ricellis:
+1

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
```diff
@@ -0,0 +1,329 @@
...
+ *           the intervals when replication is required and handles resetting of alarms after
+ *           reboot of the device.
+ */
+public abstract class PeriodicReplicationService<T extends PeriodicReplicationReceiver>
+    extends ReplicationService {
+
+    /* Name of the SharedPreferences file used to store alarm times. We store the alarm
+     * times in preferences so we can reset the alarms as accurately as possible after reboot
+     * and so we can adjust alarm times when components bind to or unbind from this Service. */
+    private static final String PREFERENCES_FILE_NAME = "com.cloudant.preferences";
```
mikerhodes: I wonder if this or the field names should include the class name somehow -- to both help people track down what's creating preferences if they come across them and to help if we add other preferences-using classes in future?

mikerhodes:
As usual, my questions are around threading. Is it possible for stop/start replications to be called from more than one thread? If so, it looks like it's possible to schedule two alarms.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
brynh: Good idea. My preference would be to add the fully qualified package name to the field names in the file rather than changing the filename. My personal feeling is that developers are more likely to be accepting of a single Cloudant specific preferences file than multiple files, so having a generic name like `com.cloudant.preferences` allows us to use it if we ever need to for anything else, but by looking at the fields inside you get the traceability to where the preference is used.
Done in c7b0975.

brynh:
> As usual, my questions are around threading. Is it possible for stop/start replications to be called from more than one thread? If so, it looks like it's possible to schedule two alarms.

Made thread safe in a69a5e0.

mikerhodes:
+1

brynh:

PR #200, opened 2015-09-08
Add policy for periodic replication only on Wifi.

brynh:
This is part 5 of a 6 part change to implement replication policies. Each of the 6 parts of the implementation of replication policies are each in a separate branch.

_The following text is common to all 6 PRs._

The 6 branches include:
1. _48607-policies-1-replicator-id_: Adding an ID to replicators so we can easily identify which replicators are completed or have errored in subsequent callbacks.
2. _48607-policies-2-java_: Adding replication policies for Java.
3. _48607-policies-3-android-service_:Adding basic replication policies for Android.
4. _48607-policies-4-android-periodic_: Adding replication at regular intervals for Android.
5. _48607-policies-5-android-wifi_: Adding replication only when on Wifi.
6. _48607-policies-6-docs_: Documentation updates.

_What?_
Add mechanism to enable users to more easily specify the circumstances under which they want replication to occur. This allows relatively simple definition of policies such as:
* Do a pull replication every 2 hours;
* Do a full sync every hour, but only when we have a Wifi connection;
* Sync whenever we connect to Wifi.

_Why?_
Prior to this PR, it was very difficult to implement such policies, particularly on Android where many aspects of the way the Android operating system works would need to be taken into account in order for such policies to work reliably.

_How?_
On Android we want the replication policy management to be reliable, battery efficient and not hog system resources unnecessarily. To achieve this we typically might trigger our replications from a `BroadcastReceiver`. This allows our code to be completely inactive on the device and be started when events of interest are broadcast.  We can also use a `BroadcastReceiver` to trigger periodic replications via the `AlarmManager` mechanism in an efficient way.  However, a `BroadcastReceiver` is intended to only run for a short time in response to receiving a broadcast, so the replication is carried on inside a `Service` that may be started from a `BroadcastReceiver`.

Running replications in a `Service` allows them to run independently of the lifecycle of other application components and be started easily via other application components.  We must also handle the fact that when we wish replication to occur, the device may be asleep, so we need to hold a `WakeLock` for the duration of the replication to prevent the device going back to sleep once the `onReceive()` method of the `BroadcastReceiver` triggering the replications has completed.  Also, if our replications are over Wifi, we need to hold a `Wifi` lock to keep the Wifi radio awake.

On Java, replication policies are much simpler than on Android. The new `ReplicationPolicyManager` class takes care of ensuring that if replications are in progress they are not restarted.

[...]

PR #201, opened 2015-09-08
Add replication policy user guide.

brynh:
This is part 6 of a 6 part change to implement replication policies. Each of the 6 parts of the implementation of replication policies are each in a separate branch.

The 6 branches include:

_What?_

_Why?_

_How?_

reviewer @mikerhodes
reviewer @ricellis

### DIFF
--- a/REPLICATION_POLICIES.md
+++ b/REPLICATION_POLICIES.md
@@ -32,7 +32,8 @@ This abstract class is a child of `ReplicationService` and requires you to imple
 * `protected abstract int getUnboundIntervalInSeconds();` This method should return the interval (in seconds) you wish to have between replications when components are not bound to the Service.
 * `protected abstract boolean startReplicationOnBind();` This should return `true` if you wish to have replications triggered immediately when a component binds to the Service.
 
-Note that internally this uses `AlarmManager.setInexactRepeating()` to schedule the repetition of the replications at the given intervals in a battery efficient way, this means the intervals between replications will not be exact.
+Note that internally this uses [android.app.AlarmManager.setInexactRepeating()](http://developer.android.com/reference/android/app/AlarmManager.html#setInexactRepeating(int, long, long, android.app.PendingIntent)) 
+to schedule the repetition of the replications at the given intervals in a battery efficient way, this means the intervals between replications will not be exact.
 
 ###PeriodicReplicationReceiver
 This uses a `WakefulBroadcastReceiver` to trigger periodic replications at the intervals specified by your
--- a/REPLICATION_POLICIES.md
+++ b/REPLICATION_POLICIES.md
@@ -13,9 +13,9 @@ one of the existing replication Service components and implement the specifics o
 
 There are two options when creating your service:
 
-1. If you don't require periodic replications, you should create a subclass of `ReplicationService`.
+1. If you don't require periodic replications, create a subclass of `ReplicationService`.
 2. If you want your replications to repeat at (roughly) fixed time intervals, as well as possibly adding other criteria that
-are required for replication, you should create a subclass of `PeriodicReplicationService`. This class takes care of the
+are required for replication, create a subclass of `PeriodicReplicationService`. This class takes care of the
 scheduling of periodic alarms to trigger the replications. This also requires you to implement a `BroadcastReceiver` that is a subclass of `PeriodicReplicationReceiver`. 
 
 Both Services allow other application components to bind to them. This allows components to be notified when
--- a/REPLICATION_POLICIES.md
+++ b/REPLICATION_POLICIES.md
@@ -87,7 +87,8 @@ management you require yourself.
 
 #### WifiPeriodicReplicationReceiver
 
-This class is an example of how to extend `PeriodicReplicationReceiver` to add logic that triggers periodic replications
+[`WifiPeriodicReplicationReceiver`](cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/WifiPeriodicReplicationReceiver.java)
+is an example of how to extend `PeriodicReplicationReceiver` to add logic that triggers periodic replications
 only when our device is connected to Wifi and stops periodic replications when we disconnect from Wifi.
 This is done by extending the `PeriodicReplicationReceiver` so that the 
 [`BroadcastReceiver`](http://developer.android.com/reference/android/content/BroadcastReceiver.html) responds to changes in
--- a/REPLICATION_POLICIES.md
+++ b/REPLICATION_POLICIES.md
@@ -1,12 +1,16 @@
+# Replication Policies
+
 Replication policies work quite differently on Android from Java. Please see the relevant section of this guide 
 for [Android](#android-replication-policies) or [Java](#java-replication-policies).
 
-# Android replication policies
+## Android replication policies
+
 Replication policies on Android run in a `Service` to allow them to run independently of your main application and to allow
 them to be restarted if they are killed by the operating system.  To use replication policies, you need to subclass
 one of the existing replication Service components and implement the specifics of your required replications.
 
-##Creating your service
+### Creating your service
+
 There are two options when creating your service. 
 
 1. If you don't require periodic replications, you should create a subclass of `ReplicationService`.
@@ -21,11 +25,13 @@ triggers replications every 24 hours in the background when the app is not activ
 used and a particular `Activity` is displayed it may be desirable to replicate every few minutes and then update the
 `Activity` to show the data when replication has completed.
 
-###ReplicationService
+#### ReplicationService
+
 `ReplicationService` is an abstract class and your subclass must implement the following abstract method:
 * `protected abstract Replicator[] getReplicators(Context context);` This method should return an array of `Replicator` objects that define the replications you want to invoke.
 
-###PeriodicReplicationService
+#### PeriodicReplicationService
+
 This abstract class is a child of `ReplicationService` and requires you to implement the following abstract methods:
 * `protected abstract Replicator[] getReplicators(Context context);` This method should return an array of `Replicator` objects that define the replications you want to invoke.
 * `protected abstract int getBoundIntervalInSeconds();` This method should return the interval (in seconds) you wish to have between replications when components are bound to the Service.
@@ -35,7 +41,8 @@ This abstract class is a child of `ReplicationService` and requires you to imple
 Note that internally this uses [android.app.AlarmManager.setInexactRepeating()](http://developer.android.com/reference/android/app/AlarmManager.html#setInexactRepeating(int, long, long, android.app.PendingIntent)) 
 to schedule the repetition of the replications at the given intervals in a battery efficient way, this means the intervals between replications will not be exact.
 
-###PeriodicReplicationReceiver
+#### PeriodicReplicationReceiver
+
 This uses a `WakefulBroadcastReceiver` to trigger periodic replications at the intervals specified by your
 `PeriodicReplicationService`. This means your application does not have to keep running to trigger replications at the
 intervals you require, but will be restarted at the time a replication is required. This class also handles resetting of
@@ -47,7 +54,8 @@ the periodic replications after a reboot of the device. You must add the followi
 <action android:name="android.intent.action.BOOT_COMPLETED" />
 ```
 
-##Controlling the replication service
+### Controlling the replication service
+
 To control the replication service you start the service passing an Extra in the `Intent` used to start the service whose
 key is `ReplicationService.EXTRA_COMMAND`, and whose value is one of:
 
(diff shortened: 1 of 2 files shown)


REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,292 @@
...
+
+###PeriodicReplicationService
+This abstract class is a child of `ReplicationService` and requires you to implement the following abstract methods:
+* `protected abstract Replicator[] getReplicators(Context context);` This method should return an array of `Replicator` objects that define the replications you want to invoke.
+* `protected abstract int getBoundIntervalInSeconds();` This method should return the interval (in seconds) you wish to have between replications when components are bound to the Service.
+* `protected abstract int getUnboundIntervalInSeconds();` This method should return the interval (in seconds) you wish to have between replications when components are not bound to the Service.
+* `protected abstract boolean startReplicationOnBind();` This should return `true` if you wish to have replications triggered immediately when a component binds to the Service.
+
+Note that internally this uses `AlarmManager.setInexactRepeating()` to schedule the repetition of the replications at the given intervals in a battery efficient way, this means the intervals between replications will not be exact.
```
ricellis: What did you think about using a fully qualified name for this? Maybe a link to the Android doc so people can find the information easily about how inexact this is.

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,292 @@
...
+management you require yourself.
+
+###WifiPeriodicReplicationReceiver
+This class is an example of how to extend `PeriodicReplicationReceiver` to add logic that triggers periodic replications
+only when our device is connected to Wifi and stops periodic replications when we disconnect from Wifi.
+This is done by extending the `PeriodicReplicationReceiver` so that the BroadcastReceiver responds to changes in
+network connectivity.  When we detect that the device has connected to Wifi we start our PeriodicReplicationService
+by sending it the command `PeriodicReplicationService.COMMAND_START_PERIODIC_REPLICATION` and when the device disconnects
+from Wifi we stop the periodic replications by sending `PeriodicReplicationService.COMMAND_STOP_PERIODIC_REPLICATION` to the
+Service. Other Intent actions received by the broadcast receiver are passed to parent class so that periodic alarms and
```
ricellis: Elsewhere Intent has been formatted as `Intent`

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,292 @@
...
+            return new Replicator[]{pullReplicator, pushReplicator};
+        } catch (URISyntaxException e) {
+            e.printStackTrace();
+        }
+        return null;
+    }
+
+    @Override
+    protected int getBoundIntervalInSeconds() {
+        return 5 * 60; // 5 minutes
```
ricellis: Does Android have `java.util.concurrent.TimeUnit`?
I find it can be quite readable to use it such as:
`TimeUnit.SECONDS.convert(5l, TimeUnit.MINUTES);`

REPLICATION_POLICIES.md
brynh: Added link to doc in 1f77614. I think using FQNs for some less frequently used classes like AlarmManager is useful (so I've done it).  However, personally, I think adding them for the 4 fundamental Android app components (`Service`, `BroadcastReceiver`, `Activity` and `ContentProvider`) is less useful and makes things less readable. Android developers should really know what these 4 are and I'd expect them not to be confused.

REPLICATION_POLICIES.md
brynh: Fixed in 6992948.

REPLICATION_POLICIES.md
brynh: Android does have `java.util.concurrent.TimeUnit`, so it could be done as you suggest (which I agree is quite nice), but in the context of this example, I'm not sure it will make it more readable, particularly for people unfamiliar with `TimeUnit`. The current implementation doesn't preclude people from doing it your way in real code if they want to (although they'd have to cast the return value to `int`).

REPLICATION_POLICIES.md
ricellis: Sounds good.

REPLICATION_POLICIES.md
ricellis: Agreed, I don't think a change would add value in this example.

ricellis:
+1

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,293 @@
+Replication policies work quite differently on Android from Java. Please see the relevant section of this guide
```
mikerhodes: You need your h1 above this, "Replication Policies" with the other headings shifted down a level.

Otherwise it's too easy to skip over this important sentence.

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,293 @@
+Replication policies work quite differently on Android from Java. Please see the relevant section of this guide 
+for [Android](#android-replication-policies) or [Java](#java-replication-policies).
+
+# Android replication policies
+Replication policies on Android run in a `Service` to allow them to run independently of your main application and to allow
+them to be restarted if they are killed by the operating system.  To use replication policies, you need to subclass
+one of the existing replication Service components and implement the specifics of your required replications.
+
+##Creating your service
```
mikerhodes: Spaces after the `#` characters. I also prefer new lines after headings.

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,293 @@
...
+There are two options when creating your service.
```
mikerhodes: Colon not full stop.

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,293 @@
...
+# Android replication policies
+Replication policies on Android run in a `Service` to allow them to run independently of your main application and to allow
+them to be restarted if they are killed by the operating system.  To use replication policies, you need to subclass
+one of the existing replication Service components and implement the specifics of your required replications.
+
+##Creating your service
+There are two options when creating your service. 
+
+1. If you don't require periodic replications, you should create a subclass of `ReplicationService`.
+2. If you want your replications to repeat at (roughly) fixed time intervals, as well as possibly adding other criteria that
```
mikerhodes: You can remove `you should` from `you should create`.

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,293 @@
...
+* `protected abstract Replicator[] getReplicators(Context context);` This method should return an array of `Replicator` objects that define the replications you want to invoke.
+* `protected abstract int getBoundIntervalInSeconds();` This method should return the interval (in seconds) you wish to have between replications when components are bound to the Service.
+* `protected abstract int getUnboundIntervalInSeconds();` This method should return the interval (in seconds) you wish to have between replications when components are not bound to the Service.
+* `protected abstract boolean startReplicationOnBind();` This should return `true` if you wish to have replications triggered immediately when a component binds to the Service.
+
+Note that internally this uses [android.app.AlarmManager.setInexactRepeating()](http://developer.android.com/reference/android/app/AlarmManager.html#setInexactRepeating(int, long, long, android.app.PendingIntent)) 
+to schedule the repetition of the replications at the given intervals in a battery efficient way, this means the intervals between replications will not be exact.
```
mikerhodes: This comma should be a full stop, I think.

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,293 @@
...
+
+###PeriodicReplicationReceiver
+This uses a `WakefulBroadcastReceiver` to trigger periodic replications at the intervals specified by your
```
mikerhodes: Switch `this` for the classname, it's not super-clear it means the class in the heading.

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,293 @@
...
+###ReplicationService
+`ReplicationService` is an abstract class and your subclass must implement the following abstract method:
+* `protected abstract Replicator[] getReplicators(Context context);` This method should return an array of `Replicator` objects that define the replications you want to invoke.
+
+###PeriodicReplicationService
+This abstract class is a child of `ReplicationService` and requires you to implement the following abstract methods:
+* `protected abstract Replicator[] getReplicators(Context context);` This method should return an array of `Replicator` objects that define the replications you want to invoke.
+* `protected abstract int getBoundIntervalInSeconds();` This method should return the interval (in seconds) you wish to have between replications when components are bound to the Service.
+* `protected abstract int getUnboundIntervalInSeconds();` This method should return the interval (in seconds) you wish to have between replications when components are not bound to the Service.
+* `protected abstract boolean startReplicationOnBind();` This should return `true` if you wish to have replications triggered immediately when a component binds to the Service.
```
mikerhodes: I found the trailing semi-colons on the method names distracting. I'd remove them.

REPLICATION_POLICIES.md
mikerhodes: Can you link to the Android classes referenced throughout the doc (service, receiver classes etc.)?

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,293 @@
...
+<action android:name="android.intent.action.BOOT_COMPLETED" />
+```
+
+##Controlling the replication service
+To control the replication service you start the service passing an Extra in the `Intent` used to start the service whose
+key is `ReplicationService.EXTRA_COMMAND`, and whose value is one of:
+
+* `ReplicationService.COMMAND_START_REPLICATION` This is used to start the replicators.
+* `ReplicationService.COMMAND_STOP_REPLICATION` This is used to stop replicators in progress.
+* `PeriodicReplicationService.COMMAND_START_PERIODIC_REPLICATION` This only applies if you're using the `PeriodicReplicationService`. This starts the periodic replications.
```
mikerhodes: You could maybe shorten these to `This stops the periodic replications when you are using the `PeriodicReplicationService`.

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,293 @@
...
+startWakefulService(context, intent);
+```
+Note the use of `startWakefulService()` which gets a `WakeLock` to ensure the device does not go to sleep while your replication is in progress.
+
+If you want to start replications from anywhere other than a `PeriodicReplicationReceiver` you would need to replace
+`startWakefulService(context, intent);` with `startService(intent);` in the above example and do any `WakeLock`
+management you require yourself.
+
+###WifiPeriodicReplicationReceiver
+This class is an example of how to extend `PeriodicReplicationReceiver` to add logic that triggers periodic replications
```
mikerhodes: Is it possible to link to the file in the repo?

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,293 @@
...
+protected void onStop() {
+    super.onStop();
+    if (mIsBound) {
+        unbindService(mConnection);
+        mIsBound = false;
+    }
+}
+```
+
+# Java replication policies
```
mikerhodes: I wonder if we could shift this up to the top, as it's so short. The counter-argument is that this is the less-used case.

REPLICATION_POLICIES.md
brynh: Fixed in bd814ba.

REPLICATION_POLICIES.md
brynh: Fixed in bd814ba.

REPLICATION_POLICIES.md
brynh: Fixed in 3a3ab9c.

REPLICATION_POLICIES.md
brynh: Fixed in c042e61.

REPLICATION_POLICIES.md
brynh: Fixed in d48385f.

REPLICATION_POLICIES.md
brynh: Fixed in 9ecf841.

REPLICATION_POLICIES.md
brynh: > Switch this for the classname, it's not super-clear it means the class in the heading.

Fixed in d662fc4.

brynh: > Can you link to the Android classes referenced throughout the doc (service, receiver classes etc.)?

Fixed in 2608838.

REPLICATION_POLICIES.md
brynh: Fixed in 97ae83.

REPLICATION_POLICIES.md
brynh: Fixed in 53f29ae.

REPLICATION_POLICIES.md
brynh: Moved in 351cc1a.

REPLICATION_POLICIES.md
```diff
@@ -0,0 +1,334 @@
...
+to schedule the repetition of the replications at the given intervals in a battery efficient way. This means the intervals between replications will not be exact.
+
+#### PeriodicReplicationReceiver
+
+`PeriodicReplicationReceiver` uses a 
+[`WakefulBroadcastReceiver`](http://developer.android.com/reference/android/support/v4/content/WakefulBroadcastReceiver.html)
+to trigger periodic replications at the intervals specified by your
+`PeriodicReplicationService`. This means your application does not have to keep running to trigger replications at the
+intervals you require, but will be restarted at the time a replication is required. This class also handles resetting of
+the periodic replications after a reboot of the device. You must add the following `intent-filter`s to your
```
brynh: This should say `<action> elements to the <intent-filter> of the PeriodicReplicationReceiver's entry in the 
AndroidManifest.xml.

brynh: Fixed in e9b6ade.

mikerhodes:
+1

brynh:

PR #202, opened 2015-09-18
Add updateOneFieldIndexMultithreaded test

tomblench:
- What: Check that a non-trivial number of threads (currently 20) can concurrently create documents and update the query index for those documents
- Why: Previous threading implementations of our database access layer were prone to error or deadlock when accessed by non-trivial numbers of threads
- Testing: Add new `updateOneFieldIndexMultithreaded` test to check the above

reviewers: @ricellis @brynh 

FB ticket 34396

### DIFF
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
@@ -234,8 +234,8 @@ public class IndexUpdaterTest extends AbstractIndexTestBase {
                 }
             }
             // check that the number of values on the name field is correct
-            assertThat(mikes, is(nDocs*nThreads/2));
-            assertThat(toms, is(nDocs*nThreads/2));
+            assertThat(mikes, is(nDocs/2*nThreads + nDocs%2*nThreads));
+            assertThat(toms, is(nDocs/2*nThreads));
         }catch (SQLException e) {
             Assert.fail(String.format("SQLException occurred executing %s: %s", sql, e));
         } finally {
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
@@ -159,6 +159,8 @@ public class IndexUpdaterTest extends AbstractIndexTestBase {
         class PopulateThread extends Thread {
             @Override
             public void run() {
+                // try to get all threads to start simultaneously
+                latch.countDown();
                 try {
                     latch.await();
                 } catch (InterruptedException e) {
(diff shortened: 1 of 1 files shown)


cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
```diff
@@ -125,6 +127,123 @@ public void updateOneFieldIndex() throws Exception {
...
+        } finally {
+            DatabaseUtils.closeCursorQuietly(cursor);
+        }
+
+        final int nDocs = 50;
+        final int nThreads = 8;
+        final int nUpdates = 10; // update index after how many docs?
+
+        // populate nDocs documents per thread, across nThreads simultaneously
+        // and update index after each insert
```
ricellis: This comment is no longer valid as the index updates now happen after each `nUpdates` batch

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
```diff
@@ -125,6 +127,123 @@ public void updateOneFieldIndex() throws Exception {
...
+        List<Thread> threads = new ArrayList<Thread>();
+
+        // Create, start and wait for the threads to complete
+        for (int i = 0; i < nThreads; i++) {
+            threads.add(new PopulateThread());
+        }
+        for (Thread t : threads) {
+            t.start();
+            // try to get all threads to start simultaneously
+            latch.countDown();
```
ricellis: I think moving this inside the `run()` method gives a slightly tighter window of the threads all being notified at the same time because the last thread will be guaranteed to be in its `run` method before the last `countdown` happens. As we discussed earlier all of this is going to be very subject to thread scheduling anyway so it probably doesn't make much difference.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
```diff
@@ -125,6 +127,123 @@ public void updateOneFieldIndex() throws Exception {
...
+                if (docNum % 2 == 0) {
+                    assertThat(cursor.getString(2), is("mike"));
+                    mikes++;
+                } else {
+                    assertThat(cursor.getString(2), is("tom"));
+                    toms++;
+                }
+            }
+            // check that the number of values on the name field is correct
+            assertThat(mikes, is(nDocs*nThreads/2));
```
ricellis: Whilst these assertions are ok now, this first one could easily break if someone changes `nDocs` or `nThreads` to produce an odd multiple. I think 
`(nDocs*nThreads/2) + (nDoc*nThreads % 2)` always works.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
ricellis: @tomblench did you miss this note? I think it needs to say "and update index after nUpdates" now

tomblench: thanks, will amend it

ricellis:
+1

brynh:
Just an observation - this PR mentions 20 threads, but the code uses 8. However, that doesn't appear to materially affect the test.

brynh:
+1

PR #203, opened 2015-09-22
Correctly handle interceptors for Push Replications

rhyshort:
Fixed issue where interceptors were not set on the Replication object backing the
replication builder, So when `build` was called HTTP interceptors were not added
to the replicator.

BugzId: 52511

reviewer @ricellis 
reviewer @tomblench


ricellis:
+1

ricellis:
As discussed I'd prefer to see a common `build()` method for both types, but understand there will be some refactoring in this area soon.

tomblench:
+1

PR #204, opened 2015-09-23
Remove setCustomHeaders

tomblench:
- What: Remove `setCustomHeaders` from `CouchConfig`

- How: Remove methods and associated code. Add documentation explaining how to set headers via Request Interceptors

- Testing: `HttpRequestsTest` modified to set headers via Request Interceptors

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchConfig.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchConfig.java
@@ -22,9 +22,8 @@ import com.cloudant.http.HttpConnectionRequestInterceptor;
 import com.cloudant.http.HttpConnectionResponseInterceptor;
 
 import java.net.URI;
-import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
 
 public class CouchConfig {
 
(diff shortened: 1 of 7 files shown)


rhyshort:
+1

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchConfig.java
```diff
@@ -35,37 +38,23 @@
...
-    // Optional custom headers
-    private Map<String, String> customHeaders;
+    private List<HttpConnectionRequestInterceptor> requestInterceptors = new ArrayList
+            <HttpConnectionRequestInterceptor>();
+    private List<HttpConnectionResponseInterceptor> responseInterceptors = new ArrayList
+            <HttpConnectionResponseInterceptor>();
 
-    private List<HttpConnectionRequestInterceptor> requestInterceptors;
-    private List<HttpConnectionResponseInterceptor> responseInterceptors;
+    private HttpConnectionRequestInterceptor headerInterceptor;
```
ricellis: Is this field actually used? Isn't the headerInterceptor just added into the list of `requestInterceptors`?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullReplication.java
```diff
@@ -33,8 +33,8 @@
     /**
      * URI for this replication's remote database.
      *
-     * Include username and password in the URL, or supply an Authorization header using
-     * setCustomHeaders() in CouchConfig.
+     * Include username and password in the URL, or supply an Authorization header using an
+     * HttpConnectionRequestInterceptor
```
ricellis: Isn't the `CookieInterceptor`added by default? In which case this statement could be pretty misleading or is it more of an internal note hidden behind the `ReplicatorBuilder`?

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchConfig.java
```diff
@@ -35,37 +38,23 @@
...
-        this.customHeaders = customHeaders;
-    }
-
     public List<HttpConnectionRequestInterceptor> getRequestInterceptors() {
         return requestInterceptors;
     }
 
     public void setRequestInterceptors(List<HttpConnectionRequestInterceptor> requestInterceptors) {
-        this.requestInterceptors = requestInterceptors;
+        this.requestInterceptors.addAll(requestInterceptors);
```
ricellis: This change makes the method name controversial because it doesn't actually set the field any more. It should probably be renamed `addRequestInterceptors` (and similarly for `Response`).
This also means the interceptors cannot be replaced using the "setter". Is the expected pattern that if you want to change interceptors you manipulate the list from the `get*` method or that you create a new `CouchConfig`?

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchConfig.java
tomblench: Thanks for spotting, fixed in 39dd54b1c3b291b83b5f475a47aed1a7bfa870ce

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchConfig.java
tomblench: I think you're right about this, it was a vestige left over from some previous commits - I'm going to revert it.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullReplication.java
rhyshort: Nope, `CookieInterceptor` is available by default, but we don't replace the creds yet, the plan is to add that in the future, not sure when it is planned for though

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchConfig.java
tomblench: Done in e68a8d3 and a9b5f75

ricellis:
+1

PR #206, opened 2015-09-28
Call executeToInputStream in revsDiff

tomblench:
This method erroneously called execute() directly on the underlying connection which meant that the interceptors did not run.

reviewer: @rhyshort


rhyshort:
+1 LGTM

PR #207, opened 2015-10-02
Feature 48607 policies

brynh:
This feature branch is the merge of all the following branches following completion of their PR reviews:
1. 48607-policies-1-replicator-id (PR #196): Adding an ID to replicators so we can easily identify which replicators are completed or have errored in subsequent callbacks.
2. 48607-policies-2-java (PR #197): Adding replication policies for Java.
3. 48607-policies-3-android-service (PR #198):Adding basic replication policies for Android.
4. 48607-policies-4-android-periodic (PR #199): Adding replication at regular intervals for Android.
5. 48607-policies-5-android-wifi (PR #200): Adding replication only when on Wifi.
6. 48607-policies-6-docs (PR #201): Documentation updates.

What?

Do a pull replication every 2 hours;
Do a full sync every hour, but only when we have a Wifi connection;
Sync whenever we connect to Wifi.
Why?

How?
On Android we want the replication policy management to be reliable, battery efficient and not hog system resources unnecessarily. To achieve this we typically might trigger our replications from a BroadcastReceiver. This allows our code to be completely inactive on the device and be started when events of interest are broadcast. We can also use a BroadcastReceiver to trigger periodic replications via the AlarmManager mechanism in an efficient way. However, a BroadcastReceiver is intended to only run for a short time in response to receiving a broadcast, so the replication is carried on inside a Service that may be started from a BroadcastReceiver.

Running replications in a Service allows them to run independently of the lifecycle of other application components and be started easily via other application components. We must also handle the fact that when we wish replication to occur, the device may be asleep, so we need to hold a WakeLock for the duration of the replication to prevent the device going back to sleep once the onReceive() method of the BroadcastReceiver triggering the replications has completed. Also, if our replications are over Wifi, we need to hold a Wifi lock to keep the Wifi radio awake.

On Java, replication policies are much simpler than on Android. The new ReplicationPolicyManager class takes care of ensuring that if replications are in progress they are not restarted.

See the Replication Policies User Guide for further information on how replication policies are implemented on Android and on Java.

reviewer @mikerhodes
reviewer @ricellis


ricellis:
+1

mikerhodes:
+1

PR #208, opened 2015-10-05
Run tests with cookie authentication

tomblench:
Allow tests to run with `CookieInterceptor`s enabled for replication
etc with cookie authentication switched on.

To run tests:

- Configure CouchDB instance with appropriate admin user and password
- Run test with flags as follows:

`gradle -Dtest.with.specified.couch=true -Dtest.couch.username=admin
-Dtest.couch.password=admin -Dtest.couch.use.cookies=true
integrationTest`

### DIFF
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/common/CouchTestBase.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/common/CouchTestBase.java
@@ -50,18 +50,13 @@ public abstract class CouchTestBase {
         }
     }
 
-    private static final Boolean SPECIFIED_COUCH = Boolean.valueOf(
-            System.getProperty("test.with.specified.couch",Boolean.FALSE.toString()));
 
-    public static final Boolean IGNORE_COMPACTION = Boolean.valueOf(
-            System.getProperty("test.couch.ignore.compaction",Boolean.FALSE.toString()));
-
-    public static final Boolean IGNORE_AUTH_HEADERS = Boolean.valueOf(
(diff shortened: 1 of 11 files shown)


cloudant-sync-datastore-core/src/test/java/com/cloudant/common/CouchTestBase.java
```diff
@@ -50,18 +50,13 @@
...
-
-    public static final Boolean IGNORE_AUTH_HEADERS = Boolean.valueOf(
-            System.getProperty("test.couch.ignore.auth.headers",Boolean.TRUE.toString()));
 
     public CouchConfig getCouchConfig(String db) {
 
-        if (SPECIFIED_COUCH) {
+        // setting COUCH_URI implies SPECIFIED_COUCH but means we are giving the full URI rather
+        // than the component parts
+        if (TestOptions.COUCH_URI != null || TestOptions.SPECIFIED_COUCH) {
```
rhyshort: Shouldn't we still require specified_couch=true as a command line option to get a custom couchconfig?

tomblench: I don't see the value in it. It means you always have to say specified_couch=true and couch_uri=http://blah so why not just make the specified bit implicit? Otherwise you're needlessly causing people to do more typing.

rhyshort: yeah true

rhyshort:
Okay we need to update CONTRIBUTING.md with the new options, and explanations about the overrides too

rhyshort:
+1 once comment on the last commit is addressed

PR #210, opened 2015-10-09
Fix replication policy tests

brynh:
This fixes some test failures in Android specific tests for replication policies.

This PR addresses the following issues:
1. During the original PR reviews a bug was introduced in ffc605e, which caused the time of the next alarm due to be updated always rather than only when the time was not within the interval used for the alarm. An explicit test that the time was not updated therefore failed as a result.
2. During the original PR reviews a bug was introduced to the tests in 09e4eb8 because the keys used to identify the alarms in `SharedPreferences` were changed, but these changes were not reflected in the tests.
3. The mocked `WifiManager` was not being setup before it was being used in the tests `testOnStartCommandStartReplication()` and `testOnStartCommandStopReplication()`. Also, in a real system it cannot be guaranteed that `getSystemService()` will be successful before the `Service`s `onCreate()` method is called, so the call is therefore moved from the field initialisation to the `onCreate()` method.
4. The code relating to the `WifiLock` is updated so on a device without Wifi it doesn't cause a null pointer exception.

Reviewers: @tomblench, @ricellis


ricellis:
+1

tomblench:
+1 these tests now pass on our jenkins android builds

PR #211, opened 2015-10-14
Initial work to support _bulk_get

tomblench:
This PR adds the necessary internal support for the new `_bulk_get` endpoint.

We can switch between `_bulk_get` and "one document at a time" whilst providing the same interface to the replication code. Currently only the "one document at a time" code path works. A later PR will add the functionality for calling `_bulk_get` and deserialising responses in a streaming/iterable manner.

* Rename existing CouchClient bulkXXX methods to bulkPostXXX
* Add dummy CouchClient.bulkGet method (implementation to come...)
* Add CouchClientWrapper.bulkGetRevisions - returns an object which
  can be iterated over to get each revision in turn
* Convert GetRevisionTask to use above iterator

reviewer: @mikerhodes 
reviewer: @ricellis


cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/BulkGetRequest.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * This is in the format which the _bulk_get endpoint understands, where &lt;doc id, rev id&gt;
+ * pairs are given. If multiple rev ids are required, then the doc id needs to be repeated across
+ * multiple requests.
+ */
+public class BulkGetRequest {
+    public String id;
+    public String rev;
+    public List<String> atts_since;
+}
```
ricellis: \n

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/BulkGetRequest.java
```diff
@@ -0,0 +1,34 @@
...
+ * Represents a bulk GET request for documents
+ *
+ * This is in the format which the _bulk_get endpoint understands, where &lt;doc id, rev id&gt;
+ * pairs are given. If multiple rev ids are required, then the doc id needs to be repeated across
+ * multiple requests.
+ */
+public class BulkGetRequest {
+    public String id;
+    public String rev;
+    public List<String> atts_since;
```
ricellis: If this represents the _bulk_get request does it need additional properties for the query parameters?
`boolean revs`
`boolean attachments`

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -541,7 +542,7 @@ private void closeQuietly(InputStream is) {
         }
     }
 
-    InputStream bulkInputStream(List<?> objects) {
+    private InputStream bulkPostInputStream(List<?> objects) {
```
ricellis: This is slightly misleading from the aspect that a `_bulk_get` request is also a `POST`, but maybe that is just because I know that endpoint detail. Maybe `bulkInsert` & `bulkRetrieve` make the intent more clear without the ambiguity of the HTTP request type?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
```diff
@@ -269,14 +269,16 @@ private int processOneChangesBatch(ChangesResultWrapper changeFeeds)
...
 
-            List<Callable<DocumentRevsList>> tasks = createTasks(batch, missingRevisions);
+            Callable<Iterable<DocumentRevsList>> tasks = createTask(batch, missingRevisions);
             try {
-                List<Future<DocumentRevsList>> futures = executor.invokeAll(tasks);
-                for(Future<DocumentRevsList> future : futures) {
-                    DocumentRevsList result = future.get();
+                Future<Iterable<DocumentRevsList>> future = executor.submit(tasks);
 
+                Iterable<DocumentRevsList> result = future.get();
```
ricellis: `future.get()` is a blocking call and this now appears to be a single task submitted for execution on the previous line, so if there is no parallel processing to be gained what is the value of using the `executor` here?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTask.java
```diff
@@ -57,55 +56,50 @@
...
 
     @Override
     public String toString() {
-        StringBuilder s = new StringBuilder("GetRevisionTask: ")
-                .append("{ documentId : \"").append(this.documentId).append("\", ")
-                .append("openRevisions : ").append(Arrays.asList(this.openRevisions)).append(" }");
-        return s.toString();
+        return "GetRevisionTask{" +
+                "sourceDb=" + sourceDb +
+                ", requests=" + requests +
```
ricellis: I think the `toString()` of `List` calls `toString()` on each member, so wouldn't this be more useful of `BulkGetRequest` also had a `toString()`?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/BasicPullStrategyMockTest.java
```diff
@@ -170,29 +171,26 @@ public void testReplicationDocWithEmptyId() throws Exception {
...
-        revs = new ArrayList<String>();
-        revs.add("1-13d33701a0954729ad029adf8fdc5a04");
-        when(mockRemoteDb.getRevisions("4d3b3f01362649d79b31d9092799a7e0", revs, new HashSet<String>(),false)).then(new Answer<Object>() {
-            @Override
-            public Object answer(InvocationOnMock invocation) throws Throwable {
-
-                return loadOpenRevsResponseFromFixture("fixture/testReplicationDocWithEmptyId_open_revs_2.json");
-
+                List<BulkGetRequest> requests = ((List<BulkGetRequest>)invocation.getArguments()[0]);
+                System.out.println(requests.get(0).id);
```
ricellis: test logging or remove?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/BasicPullStrategyMockTest.java
```diff
@@ -170,29 +171,26 @@ public void testReplicationDocWithEmptyId() throws Exception {
...
+                if (requests.get(0).id.equals("")) {
+                    // well this will never get called with "" because empty ids get skipped - should we fail?
+                    return loadOpenRevsResponseFromFixture("fixture/testReplicationDocWithEmptyId_open_revs_1.json");
+                }
+                else if (requests.get(0).id.equals("4d3b3f01362649d79b31d9092799a7e0")) {
+                    return loadOpenRevsResponseFromFixture("fixture/testReplicationDocWithEmptyId_open_revs_2.json");
+                }
+                System.out.println("This shouldn't happen");
```
ricellis: test logging?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/BasicPullStrategyMockTest.java
```diff
@@ -170,29 +171,26 @@ public void testReplicationDocWithEmptyId() throws Exception {
...
-        when(mockRemoteDb.getRevisions("4d3b3f01362649d79b31d9092799a7e0", revs, new HashSet<String>(),false)).then(new Answer<Object>() {
-            @Override
-            public Object answer(InvocationOnMock invocation) throws Throwable {
-
-                return loadOpenRevsResponseFromFixture("fixture/testReplicationDocWithEmptyId_open_revs_2.json");
-
+                List<BulkGetRequest> requests = ((List<BulkGetRequest>)invocation.getArguments()[0]);
+                System.out.println(requests.get(0).id);
+                if (requests.get(0).id.equals("")) {
+                    // well this will never get called with "" because empty ids get skipped - should we fail?
```
ricellis: If our expected behaviour is that empty IDs should be skipped then yes, we should fail.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
```diff
@@ -269,14 +269,16 @@ private int processOneChangesBatch(ChangesResultWrapper changeFeeds)
 
             if (this.cancel) { break; }
 
-            List<Callable<DocumentRevsList>> tasks = createTasks(batch, missingRevisions);
+            Callable<Iterable<DocumentRevsList>> tasks = createTask(batch, missingRevisions);
```
ricellis: Why no more multiple tasks here? Isn't there still an advantage to operating in parallel for the case where `_bulk_get` isn't supported?

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/BulkGetRequest.java
tomblench: This is meant to capture the fields needed for one entry in the `docs` array - as in https://github.com/couchbase/sync_gateway/wiki/Bulk-GET

So maybe this class should be called `BulkGetRequestItem`.

`revs` and `attachments` is for the whole request. I'm minded to keep this structure as it is for now, but I might re-visit this (in a later PR) and change to a structure which captures the whole list of requests as well as the flags.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BulkGetRequest.java
```diff
@@ -0,0 +1,49 @@
...
+package com.cloudant.sync.replication;
+
+import java.util.List;
+
+/**
+ * Created by tomblench on 14/10/15.
+ */
+
+/**
+ * Represents a bulkCreateDocs GET request for documents
```
ricellis: Isn't this for `bulkGetRevisions` rather than `bulkCreateDocs`?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
```diff
@@ -85,10 +83,16 @@ public BasicPullStrategy(PullReplication pullReplication) {
...
+                             PullConfiguration config) {
+        this(pullReplication, null, config);
+    }
+
+    public BasicPullStrategy(PullReplication pullReplication,
                              ExecutorService executorService,
                              PullConfiguration config) {
         Preconditions.checkNotNull(pullReplication, "PullReplication must not be null.");
 
+        // TODO pass this executor service down to the GetRevisionTask
```
ricellis: I assume this TODO and the associated TODOs in the `GetRevisionTaskThreaded` for executor configuration are going to be covered in a future PR?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
```diff
@@ -0,0 +1,210 @@
...
+        Preconditions.checkNotNull(requests, "requests cannot be null");
+        for(BulkGetRequest request : requests) {
+            Preconditions.checkNotNull(request.id, "id cannot be null");
+            Preconditions.checkNotNull(request.revs, "revs cannot be null");
+        }
+
+        this.sourceDb = sourceDb;
+        this.requests = requests;
+        this.pullAttachmentsInline = pullAttachmentsInline;
+        this.executorService = new ThreadPoolExecutor(threads, threads, 1, TimeUnit.MINUTES,
```
ricellis: Given that this `ThreadPoolExecutor` has identical core and max thead number configuration I think this is essentially a fixed size thread pool where the keep alive time has no effect (because it won't reduce below core threads anyway). As such I think `Executors.newFixedThreadPool(threads)` may be more concise.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
```diff
@@ -0,0 +1,210 @@
...
+            }
+
+            return next;
+        }
+
+        @Override
+        public DocumentRevsList next() {
+            // can't advance iterator as there was a problem in call()
+            if (!iteratorValid) {
+                return null;
```
ricellis: I think this should throw a `NoSuchElementException`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
```diff
@@ -0,0 +1,210 @@
...
+            }
+
+            try {
+                requestsOutstanding.decrementAndGet();
+                return completionService.take().get();
```
ricellis: Do you think we should use `completionService.poll(5, TimeUnit.MINUTES).get()` to match up with the `offer`? Otherwise I'm concerned that this iterator could block forever (say if the network vanishes) unless we have another timeout somewhere that guarantees the `InterruptedException` will get thrown.

tomblench: That sounds like a good idea, code which blocks forever is not a good idea in our library. Does it behave exactly the same way but with a timeout? I have to say I was put off by the name `poll` which sounded like something more CPU intensive.

ricellis: I guess the implementation varies depending on the type of queue behind the `poll` method, but I would be very surprised if it actually did a poll my guess would be that it waits for a notify when an item is added to the queue within the timeout.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BulkGetRequest.java
tomblench: Thanks for spotting, this was on over-zealous refactor. Fixed in dc50066.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
tomblench: I've decided to remove this, see 039220e

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
tomblench: Good idea, done in 9f7bd62

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
tomblench: Also done in 9f7bd62

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
tomblench: Done in 3a485d4

tomblench:
Thanks for the feedback @ricellis - hopefully everything has been addressed. I've raised a follow-on for some small items which aren't urgent (FB 56304).

ricellis:
+1

PR #212, opened 2015-10-28
Jenkins build improvements

tomblench:
A bundle of small changes to help get jenkins CI reliably running:

- Exit early instead of `Assume.assume*` - the Android runner will fail rather than skip these tests
- Exclude some tests from Android emulator testing by source path because we can't use Categories for that purpose
- Increase timeouts for ADB responses
- Upgrade android gradle tooling

reviewer: @rhyshort


rhyshort:
+1

PR #213, opened 2015-11-03
55027: Handle missing open revision error

tomblench:
See ticket for details - this is a fix for _transient_ "Missing open revision for document" exceptions thrown by `CouchClientWrapper#getRevisions`.

**What**: The customer originating the ticket has experienced this error, but have not provided details to reproduce. In the absence of hard information, it is possible that this error arises as the result of a race condition as follows:
```
        // - changes feed tells us that 1-rev is a new rev we don't have
        // - on remote db, 2-rev gets added and becomes the new leaf
        // - compact() is called and 1-rev is compacted away
        // - we try to pull 1-rev but it is no longer there
```

**How**: We alter `BasicPullStrategy#replicate` to use a `RetriableTask` to fetch the batch from the changes feed and process it. Note that it is OK to call nextBatch() repeatedly for each retry without missing any batches - this is because we know that the checkpoint has not been updated. 

**Testing**: See `BasicPullStrategyTest.pull_changesNewerThanOpenRevs`

reviewer @ricellis
reviewer @emlaver

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
@@ -14,6 +14,7 @@
 
 package com.cloudant.sync.replication;
 
+import com.cloudant.common.RetriableTask;
 import com.cloudant.mazha.ChangesResult;
 import com.cloudant.mazha.CouchConfig;
 import com.cloudant.mazha.DocumentRevs;
@@ -179,6 +180,36 @@ class BasicPullStrategy implements ReplicationStrategy {
         
     }
 
+    // Inner class used to fetch and process a batch of changes. See comment in replicate() for
+    // more details.
+    private class BatchProcessor implements Callable<BatchProcessor> {
+
+        ChangesResultWrapper changeFeeds;
+        int batchChangesProcessed;
+
+        public BatchProcessor call() throws ExecutionException, InterruptedException, DocumentException, DatastoreException {
+
+            if (cancel) {return this;}
+
+            changeFeeds = nextBatch();
+            batchChangesProcessed = 0;
+
+            // So we can check whether all changes were processed during
+            // a log analysis.
+            String msg = String.format(
+                    "Batch %s contains %s changes",
+                    batchCounter,
+                    changeFeeds.size()
+            );
+            logger.info(msg);
+
+            if (changeFeeds.size() > 0) {
+                batchChangesProcessed = processOneChangesBatch(changeFeeds);
+            }
+            return this;
+        }
+    }
+
     private void replicate()
             throws DatabaseNotFoundException, ExecutionException, InterruptedException, DocumentException, DatastoreException {
         logger.info("Pull replication started");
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
```diff
@@ -179,6 +180,36 @@ public void run() {
         
     }
 
+    // Inner class used to fetch and process a batch of changes. See comment in replicate() for
+    // more details.
+    private class BatchProcessor implements Callable<BatchProcessor> {
```
ricellis: The returned `BatchProcessor` doesn't appear to be used (i.e. there is no assignment after the `call()`, is there any reason not to make this a `Callable<Void>`? Is there some future use case in mind?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
```diff
@@ -204,36 +235,48 @@ private void replicate()
...
+            // we can access changeFeeds and batchChangesProcessed fields after successful execution
+            //
+            // Note that it is OK to call nextBatch() repeatedly for each retry without missing any
+            // batches - this is because we know that the checkpoint has not been updated.
+            //
+            // Since RetriableTask#call can throw Exception, we have to re-throw the exceptions we
+            // are expecting, plus wrapping and throwing any that we aren't.
+            try {
+                new RetriableTask<BatchProcessor>(batchProcessor).call();
+            } catch (ExecutionException e) {
```
ricellis: :( really want Java 7's multi-catch about now!

tomblench: I agree. I'm at a loss for a better strategy - my guess is that anything using `Callable` or similar has to do the same because the `call()` method is capable of throwing any `Exception`.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/BasicPullStrategyTest.java
```diff
@@ -379,4 +390,86 @@ public void pull_indexesUpdated()
...
+        final ChangesResult changesResult1 = remoteDb.changes(null, 1000);
+
+        // upload some more docs
+        Bar bar2 = BarUtils.updateBar(remoteDb, bar1.getId(), "Jerry", 41);
+
+        // capture 'real' changes feed - after second rev
+        final ChangesResult changesResult2 = remoteDb.changes(null, 1000);
+
+        // compact away the first rev
+        System.out.println("compact");
```
ricellis: I'm not sure what the sync tests use, would `java.util.Logger` be more appropriate?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/BasicPullStrategyTest.java
```diff
@@ -379,4 +390,86 @@ public void pull_indexesUpdated()
...
+
+        CouchDbInfo info = couchClient.getDbInfo();
+
+        while(info.isCompactRunning()) {
+            Thread.sleep(1000);
+            info = couchClient.getDbInfo();
+        };
+
+
+        System.out.println("done compact");
```
ricellis: logging

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/BasicPullStrategyTest.java
```diff
@@ -379,4 +390,86 @@ public void pull_indexesUpdated()
...
+        // inject our mocked changesresult via a 'spy'
+        // ChangesResultAnswer returns the 'wrong' answer first and the 'right' after every
+        // invocation after that
+        customReplicator.sourceDb = spy(customReplicator.sourceDb);
+        doAnswer(new ChangesResultAnswer(changesResult1, changesResult2))
+                .when(customReplicator.sourceDb).
+                changes((Replication.Filter)
+                        anyObject(), anyObject(), anyInt());
+        // do the actual replication
+        System.out.println("running....");
```
ricellis: logging

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/BasicPullStrategyTest.java
```diff
@@ -379,4 +390,86 @@ public void pull_indexesUpdated()
...
+        customReplicator.run();
+        System.out.println("done....");
```
ricellis: log

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/BasicPullStrategyTest.java
```diff
@@ -379,4 +390,86 @@ public void pull_indexesUpdated()
...
+
+        // assert latest doc retrieved
+        Map m = datastore.getDocument(bar1.getId()).asMap();
+        System.out.println(m);
```
ricellis: log

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/BasicPullStrategyTest.java
```diff
@@ -379,4 +390,86 @@ public void pull_indexesUpdated()
...
+        Assert.assertEquals(41, m.get("age"));
+        Assert.assertEquals("Jerry", m.get("name"));
+        System.out.println("done asserts");
```
ricellis: log

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
```diff
@@ -179,6 +180,36 @@ public void run() {
...
     }
 
+    // Inner class used to fetch and process a batch of changes. See comment in replicate() for
+    // more details.
+    private class BatchProcessor implements Callable<BatchProcessor> {
+
+        ChangesResultWrapper changeFeeds;
+        int batchChangesProcessed;
+
+        public BatchProcessor call() throws ExecutionException, InterruptedException, DocumentException, DatastoreException {
```
ricellis: This should probably have an `@Override` annotation.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPullStrategy.java
ricellis: Without multi-catch I think the only alternative is catching `Exception` and doing `instanceof` checks which is not any nicer than this.

ricellis:
@tomblench is this still valid?

tomblench:
@ricellis I'm going to close this PR because I've never seen this bug occur "in the wild". If it crops up again we still have this diff as a starting point.

PR #215, opened 2015-11-19
Create GetAllDocumentIdsCallable from getAllDocumentIds method

mikerhodes:
This callable was standalone so can be pulled out with very
few changes -- literally just the IDEA refactor works for this
one, with a class-comment as to the class's purpose.

reviewer @rhyshort
reviewer @ricellis

PR #216, opened 2015-11-19
Create GetPossibleAncestorRevisionIdsCallable

mikerhodes:
This is from getPossibleAncestorRevisionIDs and also extracts the
getPossibleAncestorRevisionIDsInQueue method from BasicDatastore
which wasn't used aside from in getPossibleAncestorRevisionIDs.

This callable was standalone so can be pulled out with very
few changes -- literally just the IDEA refactor works for this
one, with a class-comment as to the class's purpose.

reviewer @ricellis
reviewer @rhyshort

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/callables/GetPossibleAncestorRevisionIdsCallable.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/callables/GetPossibleAncestorRevisionIdsCallable.java
@@ -1,0 +1,86 @@
+/**
+ * Copyright (c) 2015 IBM Cloudant. All rights reserved.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software distributed under the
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/callables/GetPossibleAncestorRevisionIdsCallable.java
```diff
@@ -0,0 +1,86 @@
...
+        this.revId = revId;
+        this.limit = limit;
+    }
+
+    @Override
+    public List<String> call(SQLDatabase db) throws Exception {
+        return getPossibleAncestorRevisionIDsInQueue(db, docId, revId, limit);
+    }
+
+    private List<String> getPossibleAncestorRevisionIDsInQueue(SQLDatabase db, final String docId,
```
ricellis: Any reason not to just put this method content into `call`? It seems a bit unnecessary having an extra method.

mikerhodes: It was basically because it made it simpler to review and because the `limit` variable is more obviously shadowed in the method.

ricellis: OK, I think it doesn't matter much either way, I was only considering it from the perspective of Android method limits.

ricellis:
+1

rhyshort:
+1

mikerhodes:
Rebased on master after merging #215

PR #217, opened 2015-11-23
54145 Make BasicDatastore handle preparing attachments more explicitly

mikerhodes:
# What

BasicDatastore and AttachmentManager are tightly coupled classes. In Objective-C the attachment code would probably be a category on the datastore class, as the split it more to keep logic together rather than encapsulate functionality.

This PR makes the preparing of attachments more explicit in the BasicDatastore code, while still leaving most of the logic in AttachmentManager. The problem was that prepareAttachments method did several things, which made it a bit harder to follow the code in BasicDatastore. The PR:

* Splits `prepareAttachments` into several explicit methods which make the logic in BasicDatastore slightly more complicated as it now splits out the unsaved from the already existing attachments. But it's clearer what's going on.
* Removes `setAttachments` in favour of explicit adding and copying operation from BasicDatastore, making it clearer that some attachments are copied and some are new. It's still a bit unclear what existing/copying means, given existing only actually means that it's a saved attachment and copying means only copying the database entries, so the method calls don't make it too clear what their cost is likely to be.
* The last couple of commits are cleanups -- they seemed small, but I can remove them in favour of a new PR is it'd help.

# Reviewers

reviewer @tomblench 
reviewer @rhyshort

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
@@ -227,11 +227,9 @@ class AttachmentManager {
     // * if attachment is not saved, prepare it, and add it to the prepared list
     // this way, the attachments are sifted through ready to be added in the database for a given revision
     protected PreparedAndSavedAttachments prepareAttachments(Collection<? extends Attachment> attachments) throws AttachmentException {
-        if (attachments == null || attachments.size() == 0) {
-            // nothing to do
-            return null;
-        }
-        // actually a list of prepared or saved attachments
+        // Ensure we have at least an empty attachments array
+        attachments = attachments != null ? attachments : new ArrayList<Attachment>();
+        // Actually a list of prepared or saved attachments
         PreparedAndSavedAttachments preparedAndSavedAttachments = new PreparedAndSavedAttachments();
         preparedAndSavedAttachments.preparedAttachments = this.prepareNewAttachments(this.findNewAttachments(attachments));
         preparedAndSavedAttachments.savedAttachments = this.findExistingAttachments(attachments);
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
@@ -227,24 +227,66 @@ class AttachmentManager {
     // * if attachment is not saved, prepare it, and add it to the prepared list
     // this way, the attachments are sifted through ready to be added in the database for a given revision
     protected PreparedAndSavedAttachments prepareAttachments(Collection<? extends Attachment> attachments) throws AttachmentException {
-        // actually a list of prepared or saved attachments
-        PreparedAndSavedAttachments preparedAndSavedAttachments = new PreparedAndSavedAttachments();
-
         if (attachments == null || attachments.size() == 0) {
             // nothing to do
             return null;
         }
+        // actually a list of prepared or saved attachments
+        PreparedAndSavedAttachments preparedAndSavedAttachments = new PreparedAndSavedAttachments();
+        preparedAndSavedAttachments.preparedAttachments = this.prepareNewAttachments(this.findNewAttachments(attachments));
+        preparedAndSavedAttachments.savedAttachments = this.findExistingAttachments(attachments);
+        return preparedAndSavedAttachments;
+    }
 
+    /**
+     * Return a list of the existing attachments in the list passed in.
+     *
+     * @param attachments Attachments to search.
+     * @return List of attachments which already exist in the attachment store, or an empty list if none.
+     */
+    private List<SavedAttachment> findExistingAttachments(Collection<? extends Attachment> attachments) {
+        ArrayList<SavedAttachment> list = new ArrayList<SavedAttachment>();
         for (Attachment a : attachments) {
-            if (!(a instanceof SavedAttachment)) {
-                PreparedAttachment pa = this.prepareAttachment(a);
-                preparedAndSavedAttachments.preparedAttachments.add(pa);
-            } else {
-                preparedAndSavedAttachments.savedAttachments.add((SavedAttachment)a);
+            if (a instanceof SavedAttachment) {
+                list.add((SavedAttachment)a);
             }
         }
+        return list;
+    }
 
-        return preparedAndSavedAttachments;
+    /**
+     * Return a list of the new attachments in the list passed in.
+     *
+     * @param attachments Attachments to search.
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -222,57 +220,55 @@ protected PreparedAttachment prepareAttachment(Attachment attachment, long lengt
...
-        if (attachments == null || attachments.size() == 0) {
-            // nothing to do
-            return null;
+    /**
+     * Return a list of the existing attachments in the list passed in.
+     *
+     * @param attachments Attachments to search.
+     * @return List of attachments which already exist in the attachment store, or an empty list if none.
+     */
+    protected List<SavedAttachment> findExistingAttachments(Collection<? extends Attachment> attachments) {
```
tomblench: Calls to `findExistingAttachments()` and `findNewAttachments()` are always paired, so why not combine them together to generate the two lists at the same time? It's a case of splitting a list into two sublists like we do for Interceptors. I understand the motivation of simplifying `prepareAttachments` but it seems a bit funny to have two almost identical methods.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -643,7 +643,7 @@ public LocalDocument call(SQLDatabase db) throws Exception {
         return null;
     }
 
-    private BasicDocumentRevision createDocument(SQLDatabase db,String docId, final DocumentBody body)
+    private BasicDocumentRevision createJsonDocument(SQLDatabase db, String docId, final DocumentBody body)
```
tomblench: What's the thinking behind the renaming, to make clear they are operating on the serialised json? Given that we use the term 'JSON' fairly loosely in methods and member names I wonder whether a suffix like 'Internal' or 'InDatabase' or similar might be more apt

doc/encryption.md
```diff
@@ -146,7 +146,7 @@ The exception has the form:
      Caused by: java.lang.RuntimeException: Unexpected type: 5
             at com.cloudant.sync.datastore.BasicDatastore.getFullRevisionFromCurrentCursor(BasicDatastore.java:1709)
             at com.cloudant.sync.datastore.BasicDatastore.getDocumentInQueue(BasicDatastore.java:284)
-            at com.cloudant.sync.datastore.BasicDatastore.createDocument(BasicDatastore.java:681)
+            at com.cloudant.sync.datastore.BasicDatastore.createJsonDocument(BasicDatastore.java:681)
```
tomblench: If we're being revisionist about stacktraces I guess you should change all the line numbers as well :)

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreExtended.java
```diff
@@ -128,7 +128,7 @@
      *                            Map of String[docId,revId] → list of attachments
      * @param pullAttachmentsInline If true, use {@code attachments} metadata and data directly
      *                              from received JSON to add new attachments for this revision.
-     *                              Else use {@code preparedAttachments} which were previously
+     *                              Else use {@code preparedNewAttachments} which were previously
```
tomblench: Changed argument name doesn't match signature here

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
rhyshort: I think combing them could leave to a messy return type, or a class we don't really need. I think having both methods is the way to go, for now at least.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
mikerhodes: Good point -- the idea was basically to say that this wasn't doing anything with the attachments only the JSON data itself. I'm not sure Internal or InDatabase get this across, though I'm not sure JSON does either -- any suggestions @rhyshort @tomblench ?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: The methods do two different things, the awkwardly named `PreparedAndSavedAttachments` class was the result of their previously being smooshed together. I couldn't think of a better name (@tomblench's comment "actually prepared or saved attachments" kinda gave away it needed changing) so split up the methods.

doc/encryption.md
mikerhodes: This would be the stacktrace from this commit on though, I think? I guess I could fix up the line numbers, but I guess you're not being super-serious?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreExtended.java
mikerhodes: Fixed in 92ad2c2.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: In addition, they need not be paired -- we could call the method to find Saved attachments within the transaction whereas the preparation does need to go outside the transaction.

doc/encryption.md
tomblench: Well I was being semi-serious. The stacktrace was correct at the time of writing. If your'e going to update the method name then you should ensure that the entire stacktrace is correct. I say you might as well leave it how it was.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
tomblench: Candidates I would suggest are `insertRevision` `insertRevisionBody` `insertRevisionJson` - something which emphasises the fact that it calls `insertRevision` and serialises the `DocumentBody` into JSON bytes before doing so.

mikerhodes: Changed in 659465a as agreed.

doc/encryption.md
mikerhodes: Reverted in 125aaa1.

tomblench:
+1

rhyshort:
+1

PR #218, opened 2015-11-26
Make AttachmentManager methods static

mikerhodes:
# What

This commit removes all state from the AttachmentManager class and makes
its methods into static ones. The two items of state, attachmentsDir and
attachmentStreamFactory are made into instance variables on
BasicDatastore instead.

# Why

This should enable us to make the AttachmentManager methods into their
own classes, which will help when we break out the callables from
BasicDatastore. In general, the overall aim is to have small classes
which encapsulate small pieces of functionality, whereas now we
have a couple of large tightly coupled classes which both know about
each others internals.

Hopefully we end up with a single class, BasicDatastore, which
encodes the knowledge about the policies for manipulating documents,
with a bunch of fairly dumb classes doing the underlying database
and file system changes.

# Reviewers

reviewer @tomblench 
reviewer @emlaver

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
@@ -47,7 +47,6 @@ import java.util.logging.Logger;
  */
 class AttachmentManager {
 
-    private static final String EXTENSION_NAME = "com.cloudant.attachments";
     private static final Logger logger = Logger.getLogger(AttachmentManager.class.getCanonicalName());
 
     private static final String SQL_ATTACHMENTS_SELECT = "SELECT sequence, " +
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -47,7 +47,6 @@
  */
```
tomblench: Comment `Attachments are stored on disk in an extension directory, {@code EXTENSION_NAME}.` probably needs to change.

tomblench:
Looks good to me. Not strictly in the remit of this PR but it might be worth re-visiting the method visibility on `AttachmentManager`. For instance, I think the `public` methods can be made `protected`.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
mikerhodes: How about f030ef8?

tomblench:
+1

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/AttachmentManager.java
```diff
@@ -190,21 +187,27 @@ public void addAttachment(SQLDatabase db, PreparedAttachment a, BasicDocumentRev
      * Creates a PreparedAttachment from {@code attachment}, preparing it for insertion into
      * the datastore.
      *
+     *
+     * @param attachmentStreamFactory
```
emlaver: I know the method is not public, but probably good to have a small comment for `attachmentStreamFactory` parameter?

emlaver:
Looks good +1

PR #219, opened 2015-11-26
Hoist insertRevision method into callable class

mikerhodes:
# What

This collects together into a single class the BasicDatastore's insertRevision method and InsertRevisionOptions class into a single class encapsulating inserting an entry into the revs table.

# Why

Part of the drive to encapsulate functionality. This is needed so that we don't end up with a "helper" class, as my first try at hoisting all the SQLCallables ended up with.

# Reviewers
reviewer @ricellis 
reviewer @emlaver


ricellis:
+1

emlaver:
LGTM +1

PR #223, opened 2016-01-04
Use Maven Central to publish Sync-Android packages

rhyshort:
## What

Move to using Maven Central to publish sync-android packages,

## How
- Remove GH Pages publishing gradle code
- Add maven central publishing code heavily inspired from [java-cloudant build.gradle](https://github.com/cloudant/java-cloudant/blob/82491d83f1dada6d7b0d2e542c182118a8af8352/build.gradle)

## Testing
Manually run upload task and snapshots will appear in https://oss.sonatype.org/content/repositories/snapshots/com/cloudant/ 

## Reviewers
reviewer @ricellis 
reviewer @tomblench 
reviewer @rhyshort

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -198,9 +198,7 @@ def getProjectDocSources(project) {
     // Get the projects from the dependencies
     def dependentProjects = projectDependencies*.dependencyProject
     // Recurse
-    if (dependentProjects.size > 0) {
     dependentProjects.each { dependentProjects += getProjectDocSources(it) }
-    }
     // Finally add the original project
     dependentProjects.add(project)
     // Return without duplicates
--- a/build.gradle
+++ b/build.gradle
@@ -3,6 +3,8 @@
 // ************ //
 
 apply plugin: 'java'
+apply plugin: 'maven'
+apply plugin: 'signing'
 
 //
 // General settings
@@ -14,6 +16,9 @@ allprojects {
     description = """cloudant-sync"""
 }
 
+
+//if the version says "snapshot" anywhere assume it is not a release
+ext.isReleaseVersion = !version.toUpperCase(Locale.ENGLISH).contains("SNAPSHOT")
 dependencies {
     compile project(':cloudant-sync-datastore-core')
     compile project(':cloudant-sync-datastore-javase')
@@ -52,18 +57,6 @@ subprojects {
     repositories {
         mavenLocal()
         mavenCentral()
-        maven { url "http://cloudant.github.io/cloudant-sync-eap/repository/" }
-    }
-
-    // local.gradle defines ghpages, so we want to publish somewhere specific for release
-    if (project.hasProperty('ghpages')) {
-        publishing {
-            repositories {
-                maven {
-                    url "$ghpages/repository"
-                }
-            }
-        }
     }
 
     //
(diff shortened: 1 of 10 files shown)


tomblench:
@rhyshort a few questions about this PR before I dive in:

- What files/env vars are needed for publishing? I can see some are used here https://github.com/cloudant/sync-android/pull/223/files#diff-c197962302397baf3a4cc36463dce5eaR116 
- Additionally, what error handling is there if these files/env vars aren't present?
- What's the procedure for publishing locally, both to a user's own `.m2` directory, and to a 'repo' dir, as we do for the packaging builds?
- How does the javadoc get published? We have a task to run javadoc over a union of all the sub-projects, does this still work properly?
- Do any of the user instructions need to change, do they need to adjust URLs, package names etc?

rhyshort:
@tomblench 

- The envs all come from java-cloudant (keeping them the same across the libs makes our lives easier), they are: 
  - ossrhUsername
  - ossrhPassword
  - signing.keyId
  - signing.password
  - signing.secretKeyRingFile

- If these properties are present, the build will either fail, for missing ossrh creds, or the deploy steps on OSSRH will fail, if the jars aren't signed.

- JavaDoc gets publish via a javadoc jar, I didn't know we had a union task so this probably doesn't work, the snapshots repo on OSSRH will need to be checked to confirm.

- There will need to be a slight adjustment in the instructions, the removal of the github.io urls to be replaced with `MavenCentral()` otherwise it will work exactly as before, although we should take the oppertunity to clarify which jars to include for the project type, eg for android you only need `cloudant-sync-datastore-android` which will pull in the core jar, or for encryption on android `cloudant-sync-datastore-android-encryption`

build.gradle
```diff
@@ -114,3 +109,25 @@ if (project.hasProperty('ghpages')) {
 task wrapper(type: Wrapper) {
     gradleVersion = '2.2.1'
 }
+
+//if the version says "snapshot" anywhere assume it is not a release
+ext.isReleaseVersion = !version.toUpperCase(Locale.ENGLISH).contains("SNAPSHOT")
```
ricellis: Duplicate of line 21?

ricellis:
Is it possible to common up the majority of the `uploadArchives` in the top level gradle file and only override the deltas in the lower level gradle files?
For example the URLs and credentials are the same as is a lot of the pom content.

rhyshort:
Possibly, I'm not sure, I was hoping you might know how to do that from your readings on the config

ricellis:
I made a lot of the upload configuration common in cab01ba

tomblench:
@rhyshort you'll definitely need to test how the javadoc publishing works before we can merge this. It would be a shame if there are separate javadocs per jar, but I suppose we can live with it. It would be nice if we could still get @see links etc to work across jars.

ricellis:
@rhyshort there is a bunch of `publishing` stuff in the subproject `build.gradle` files that I think is redundant now.

ricellis:
@rhyshort you probably ought to look over my commits as well

ricellis:
The docs need to be updated to point to the maven central repository too.

ricellis:
+1

build.gradle
```diff
@@ -74,43 +67,142 @@ subprojects {
...
+                        url 'https://java-cloudant@github.com/cloudant/sync-android.git'
+                    }
+                    properties {
+                        'project.build.sourceEncoding' 'UTF-8'
+                    }
+                    developers {
+                        developer {
+                            name 'IBM Cloudant'
+                            email 'support@cloudant.com'
+                            url 'https://cloudant.com'
```
tomblench: Could we have a more specific URL here?

rhyshort: I didn't think we could get more specific than cloudant.com for the developer name

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Jan 07 14:07:38 GMT 2015
+#Fri Dec 18 10:33:15 GMT 2015
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.2.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.2.1-all.zip
```
tomblench: Any reason why this changed? The `-all` distribution is likely to be bigger.

build.gradle
```diff
@@ -74,43 +67,142 @@ subprojects {
...
+}
+
+// Function to get project doc sources (i.e. the project and all its dependent projects)
+def getProjectDocSources(project) {
+    // Get the compile time dependencies
+    def projectDependencies = project.configurations.compile.getAllDependencies().withType(ProjectDependency)
+    // Get the projects from the dependencies
+    def dependentProjects = projectDependencies*.dependencyProject
+    // Recurse
+    if (dependentProjects.size > 0) {
```
tomblench: this `if` statement is redundant

gradle/wrapper/gradle-wrapper.properties
rhyshort: it adds code completion, I didn't realise this got checked in, I could remove if you think that is better?

build.gradle
rhyshort: fixed in 9a25c27

rhyshort: fixed in 9a25c27

tomblench:
+1

PR #224, opened 2016-01-08
Update build.gradle to use Maven Central

rhyshort:
Update build.gradle to use Maven Central to pull sync-android jars.

See: https://groups.google.com/forum/#!topic/cloudant-sync/pNgi2qtL5MQ


tomblench:
+1

PR #225, opened 2016-01-28
Avoid excessive database traffic when doing pull replications

tomblench:
Remove calls like `getAllRevisionsOfDocumentInQueue` and
`getDocumentInQueue` which incur an un-necessarily large overhead by
retrieving more data than is needed (especially document bodies) and
building temporary objects which contain large amounts of data which
are never used.

Instead we make direct SQL lookups on the data needed which is quicker
and also makes the code more compact.

reviewer @rhyshort

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
@@ -28,6 +28,7 @@ import com.cloudant.sync.datastore.migrations.MigrateDatabase6To100;
 import com.cloudant.sync.notifications.DatabaseClosed;
 import com.cloudant.sync.notifications.DocumentCreated;
 import com.cloudant.sync.notifications.DocumentDeleted;
+import com.cloudant.sync.notifications.DocumentModified;
 import com.cloudant.sync.notifications.DocumentUpdated;
 import com.cloudant.sync.sqlite.ContentValues;
 import com.cloudant.sync.sqlite.Cursor;
@@ -58,6 +59,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
@@ -72,16 +74,43 @@ class BasicDatastore implements Datastore, DatastoreExtended {
 
     private static final String FULL_DOCUMENT_COLS = "docs.docid, docs.doc_id, revid, sequence, json, current, deleted, parent";
 
+    private static final String GET_DOC_NUMERIC_ID =
+            "SELECT doc_id from docs WHERE docid=?";
+
+    // get all document columns for current ("winning") revision of a given doc id
     private static final String GET_DOCUMENT_CURRENT_REVISION =
             "SELECT " + FULL_DOCUMENT_COLS + " FROM revs, docs WHERE docs.docid=? AND revs.doc_id=docs.doc_id " +
                     "AND current=1 ORDER BY revid DESC LIMIT 1";
 
+    // get sequence number for current ("winning") revision of a given doc id
+    private static final String GET_SEQUENCE_CURRENT_REVISION =
+            "SELECT revs.sequence FROM revs, docs WHERE docs.docid=? AND revs.doc_id=docs.doc_id " +
+                    "AND current=1 ORDER BY revid DESC LIMIT 1";
+
+    // get all document columns for a given revision and doc id
     private static final String GET_DOCUMENT_GIVEN_REVISION =
             "SELECT " + FULL_DOCUMENT_COLS + " FROM revs, docs WHERE docs.docid=? AND revs.doc_id=docs.doc_id " +
                     "AND revid=? LIMIT 1";
 
+    // get sequence number for a given revision and doc id
+    private static final String GET_SEQUENCE_GIVEN_REVISION =
+            "SELECT revs.sequence FROM revs, docs WHERE docs.docid=? AND revs.doc_id=docs.doc_id " +
+                    "AND revid=? LIMIT 1";
+
     public static final String SQL_CHANGE_IDS_SINCE_LIMIT = "SELECT doc_id, max(sequence) FROM revs " +
             "WHERE sequence > ? AND sequence <= ? GROUP BY doc_id ";
+
+    // get all non-deleted leaf rev ids for a given doc id
+    public static final String GET_NON_DELETED_LEAFS = "SELECT revs.revid FROM revs " +
+            "WHERE revs.doc_id = ? " +
+            "AND revs.deleted = 0 AND revs.sequence NOT IN " +
+            "(SELECT DISTINCT parent FROM revs WHERE parent NOT NULL) ";
+
+    // get all leaf rev ids for a given doc id
+    public static final String GET_ALL_LEAFS = "SELECT revs.revid FROM revs " +
+            "WHERE revs.doc_id = ? " +
+            "AND revs.sequence NOT IN " +
+            "(SELECT DISTINCT parent FROM revs WHERE parent NOT NULL) ";
 
     // Limit of parameters (placeholders) one query can have.
     // SQLite has limit on the number of placeholders on a single query, default 999.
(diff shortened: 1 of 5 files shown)


rhyshort:
+1

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -934,138 +987,137 @@ public String call(SQLDatabase db) throws Exception {
...
-                        logger.log(Level.FINER, "Inserted revision: %s", rev);
-                        if (documentCreated != null) {
-                            return documentCreated;
-                        } else if (documentUpdated != null) {
-                            return documentUpdated;
+                        }
+                        if (ok) {
+                            logger.log(Level.FINER, "Inserted revision: %s", item.rev);
+                            if (documentCreated != null) {
+                                eventBus.post(documentCreated);
```
rhyshort: I don't think we should post event's here. Since the create/update hasn't been officially finished yet, and in theory the whole thing can be reverted if `eventBus.post` throws an runtime exception. I also wonder what the effect this will have on user's code, because they could potentially inadvertently slow replication if they do some kind of work on the thread that calls it, which would block the db queue.

tomblench: Fair point. I couldn't remember if event post/notification was async or not. We can just collect the events up in a list and post them at the end in a similar manner to how it used to be done.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1275,29 +1293,59 @@ private void pickWinnerOfConflicts(SQLDatabase db, BasicDocumentRevision previou
          */
 
         // first get all non-deleted leafs
-        List<BasicDocumentRevision> leafs = objectTree.leafRevisions(true);
+        List<String> leafs = new ArrayList<String>();
+        String sql = "SELECT revs.revid FROM revs " +
```
rhyshort: This should be a static constant as per the other SQL queries.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1275,29 +1293,59 @@ private void pickWinnerOfConflicts(SQLDatabase db, BasicDocumentRevision previou
...
+        } catch (SQLException sqe) {
+            throw new DatastoreException("Exception thrown whilst trying to fetch non-deleted leaf nodes in pickWinnerOfConflicts", sqe);
+        }
+
+        // this is a corner case - all leaf nodes are deleted
+        // re-get with the same query but without the revs.delete clause
         if (leafs.size() == 0) {
-            // all deleted, apply the normal rules to all the leafs
-            leafs = objectTree.leafRevisions();
+            sql = "SELECT revs.revid FROM revs " +
```
rhyshort: This should be a static constant like the other SQL statements

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/DatastoreWrapper.java
```diff
@@ -88,6 +90,25 @@ private String getCheckpointDocumentId(String replicatorIdentifier) {
...
+                itemsToInsert.add(new ForceInsertItem(doc, revisions, attachments, batch.attachments, pullAttachmentsInline));
+
+            }
+        }
+        dbCore.forceInsert(itemsToInsert);
+
+    }
+
+
     public void bulkInsert(DocumentRevsList documentRevsList, Map<String[],List<PreparedAttachment>> preparedAttachments, boolean pullAttachmentsInline) throws DocumentException  {
```
rhyshort: Has this method been replaced by `bulkInsert(List<BasicPullStrategy.BatchItem> batches, boolean pullAttachmentsInline) throws DocumentException`?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
rhyshort: I think that would be the best way to do it :)

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/DatastoreWrapper.java
tomblench: @rhyshort see https://github.com/cloudant/sync-android/pull/225/files#diff-170cf9606b2e074fdecdaff315d5c28bR996 - this was done for backwards compatibility but I don't think anyone will ever call it. Do you think I should just remove it?

rhyshort: I would yes, it's inefficient, we aren't tied to the API yet, the only thing that might be a problem is the usage of @snowch's p2p lib

rhyshort:
+1 :+1:

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1275,29 +1318,49 @@ private void pickWinnerOfConflicts(SQLDatabase db, BasicDocumentRevision previou
          */
 
         // first get all non-deleted leafs
-        List<BasicDocumentRevision> leafs = objectTree.leafRevisions(true);
+        List<String> leafs = new ArrayList<String>();
+        try {
+            Cursor cursor = db.rawQuery(GET_NON_DELETED_LEAFS, new String[]{Long.toString(docNumericId)});
```
brynh: This `Cursor` should be closed in a `finally` block.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -1275,29 +1318,49 @@ private void pickWinnerOfConflicts(SQLDatabase db, BasicDocumentRevision previou
...
+            throw new DatastoreException("Exception thrown whilst trying to fetch non-deleted leaf nodes in pickWinnerOfConflicts", sqe);
+        }
+
+        // this is a corner case - all leaf nodes are deleted
+        // re-get with the same query but without the revs.delete clause
         if (leafs.size() == 0) {
-            // all deleted, apply the normal rules to all the leafs
-            leafs = objectTree.leafRevisions();
+            try {
+                Cursor cursor = db.rawQuery(GET_ALL_LEAFS, new String[]{Long.toString(docNumericId)});
```
brynh: This `Cursor` should be closed in a `finally` block.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/ForceInsertItem.java
```diff
@@ -0,0 +1,25 @@
+package com.cloudant.sync.datastore;
```
brynh: This file should probably have a standard copyright header comment block.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/DatastoreWrapper.java
```diff
@@ -88,6 +90,25 @@ private String getCheckpointDocumentId(String replicatorIdentifier) {
         return "_local/" + replicatorIdentifier;
     }
 
+    public void bulkInsert(List<BasicPullStrategy.BatchItem> batches, boolean pullAttachmentsInline) throws DocumentException  {
+        List<ForceInsertItem> itemsToInsert = new ArrayList<ForceInsertItem>();
+        for (BasicPullStrategy.BatchItem batch : batches) {
```
brynh: Minor point as I doubt this `ArrayList` will ever be huge enough to make a massive performance difference, but as this is a performance PR I'll mention it anyway - traditional `for` loops are more efficient than enhanced `for` loops for `ArrayList` (but not for most other collections), so it _might_ be worth giving it a try to see what difference it makes.  See:
http://developer.android.com/training/articles/perf-tips.html#Loops

brynh:
+1

PR #226, opened 2016-02-10
Create directory for datastore management

rhyshort:
In order to make sync-android behave more like CDTDatastore,
create the directory to manage if it doesn't exist. This behaviour
should be the same as `mkdir`

reviewer @ricellis

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
@@ -98,6 +98,11 @@ public class DatastoreManager {
      */
     public DatastoreManager(File directoryPath) {
         logger.fine("Datastore path: " + directoryPath);
+
+        if(!directoryPath.exists()){
+            directoryPath.mkdir();
+        }
+
         if(!directoryPath.isDirectory() ) {
             throw new IllegalArgumentException("Input path is not a valid directory");
         } else if(!directoryPath.canWrite()) {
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
```diff
@@ -98,6 +98,11 @@ public DatastoreManager(String directoryPath) {
      */
     public DatastoreManager(File directoryPath) {
         logger.fine("Datastore path: " + directoryPath);
+
+        if(!directoryPath.exists()){
+            directoryPath.mkdir();
```
ricellis: Do we know for sure that we'd only need to create 1 part of this path? i.e. should it use `mkdirs()` instead of `mkdir()`?

Also is it worth checking the boolean returned and throwing an `Exception` if the directory couldn't be created (for example insufficient permissions, or if using `mkdir` not `mkdirs` then it is possible the parent directory didn't exist)?

rhyshort: We don't know for sure, the aim of this PR is to match CDTDatastore behaviour which does the equivalent of `mkdir $path`, so calling `mkdirs()` would then put CDTDatastore out of step with sync-android.

I don't think its worth checking the boolean returned, since we can just use the other checks to ensure that it was created successfully, or can write to it etc.

ricellis:
I understand this is an alignment exercise, but I do think we should consider an exception in future if the create has failed.

+1

rhyshort:
I agree, I'll add it to the fb ticket

PR #227, opened 2016-02-11
Revert "Avoid excessive database traffic when doing pull replications"

rhyshort:
Reverts cloudant/sync-android#225 due to a bug in the detection of previous revisions during a transaction. See slack history for full information.


tomblench:
+1

PR #228, opened 2016-02-11
51074 remove mutable document revision

tomblench:
Remove `MutableDocumentRevision`, `BasicDocumentRevision`, and clean up related code.


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryResult.java
```diff
@@ -195,7 +194,8 @@ private DocumentRevision projectFields(List<String> fields,
         revBuilder.setBody(DocumentBodyFactory.create(body));
         revBuilder.setDeleted(rev.isDeleted());
         revBuilder.setAttachments(new ArrayList<Attachment>(rev.getAttachments().values()));
-        revBuilder.setSequence(rev.getSequence());
+        // TODO sequence is internal anyway, why should we expose it out?
+        //revBuilder.setSequence(rev.getSequence());
```
rhyshort: We shouldn't expose it outside the datastore package, I think we should be safe to remove these two lines.

rhyshort:
I think you already know this but, a couple of things left to do.

1. Make sure `ProjectedDocumentRevision` can't be saved to the datastore
2. A few tests to make sure they can't be saved
3. Update documentation for `ProjectedDocumentRevision`

Then I think all we will need to do is the changes and API transition doc, which we can do in a later PR.

rhyshort:
+1

PR #229, opened 2016-02-12
2nd attempt at 'less database traffic' PR

tomblench:
Remove calls like getAllRevisionsOfDocumentInQueue and
getDocumentInQueue which incur an un-necessarily large overhead by
retrieving more data than is needed (especially document bodies) and
building temporary objects which contain large amounts of data which
are never used.

Instead we make direct SQL lookups on the data needed which is quicker
and also makes the code more compact.

reviewer @rhyshort


rhyshort:
Looks good +1

PR #233, opened 2016-02-22
Added task cancellation when replication stopped

### LINKED ISSUE #232: `java.lang.RuntimeException: Offer timed out` after stopping replications
Five minutes after stopping a replication a `java.lang.RuntimeException: Offer timed out` can be thrown.
This is caused by the replication thread no longer consuming `DocumentRevsList`, but responses from the server are waiting to append to the queue and eventually time out throwing the `RuntimeException`.

These tasks need to be cancelled and cleaned up when a replication is stopped.

ricellis:
*What*
Added task cancellation when replication stopped, fixes #232.

*How*
Added a list of tasks in `BasicPullStrategy`.
Added a cancel method on `GetRevisionTaskThreaded`.
Tracked submitted jobs in `GetRevisionTaskThreaded` and cancel them and
drain down the completed response queue when cancel is called.

*Testing*
Existing tests pass.

reviewer @tomblench 
reviewer @rhyshort

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
@@ -17,9 +17,12 @@ package com.cloudant.sync.replication;
 import com.cloudant.sync.datastore.DocumentRevsList;
 import com.google.common.base.Preconditions;
 
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorCompletionService;
@@ -80,6 +83,8 @@ class GetRevisionTaskThreaded implements Iterable<DocumentRevsList> {
     private final CouchDB sourceDb;
     private final List<BulkGetRequest> requests;
     private final boolean pullAttachmentsInline;
+    private final Set<Future<DocumentRevsList>> submittedJobs =
+            Collections.synchronizedSet(new HashSet<Future<DocumentRevsList>>());
 
     private boolean iteratorValid = true;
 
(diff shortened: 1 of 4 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
```diff
@@ -137,14 +142,36 @@ public boolean add(Future<DocumentRevsList> documentRevsListFuture) {
...
                             (request.id, request.revs, request.atts_since,
                                     GetRevisionTaskThreaded.this.pullAttachmentsInline));
                 }
-            });
+            }));
+        }
+    }
+
+    void cancel() {
+        synchronized (submittedJobs) {
```
tomblench: You don't need this `synchronized` block because `submittedJobs` was already constructed with `synchronizedSet`

ricellis: Iteration behaviour is non-deterministic for synchronized collections if the returned object is not synchronized, see https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedSet%28java.util.Set%29

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
```diff
@@ -137,14 +142,36 @@ public boolean add(Future<DocumentRevsList> documentRevsListFuture) {
...
+            for (Future<DocumentRevsList> f : submittedJobs) {
+                // Don't interrupt if already running, we'll get the running ones from the completed
+                // queue.
+                f.cancel(false);
+            }
+            // Cancelling will line a whole bunch of tasks up for addition to the queue, so we need
+            // to keep draining it
+            try {
+                while (!submittedJobs.isEmpty()) {
+                    Future<DocumentRevsList> r = completionService.take();
```
tomblench: As you mentioned, I would be a bit wary of a race condition creeping in between submittedJobs and  completionService - I wonder if there are any circumstances under which `take()` could block because there's nothing on the completionService? Maybe easier to `poll()` then `take()`?

ricellis: I've changed `take()` to `poll()` in 8223c28, but in that case if there is a race condition then the while loop might remain active polling the queue (in which case poll with timeout might actually be preferable).
I'm reasonably confident that the synchronized set should prevent any race condition, unless we've missed another code branch where `submittedJobs.remove()` needs to be called.
wdyt?

ricellis: @tomblench I've moved the `synchronized` to only surround the cancellation loop, which should allow for `submittedJobs.remove()` to be called from the iterator as well during the cancellation which I think eliminates the possibility of a job going untracked (i.e. I think there was a case before where the iterator poll could have pulled a job from the completion queue, but then ended up waiting behind the cancellation held lock to do the remove, which would prevent the cancellation completing because the jobs list would have one element left).

Additionally I changed it to poll with a timeout as per the previous comment.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/BasicPullStrategyTest.java
```diff
@@ -403,4 +408,82 @@ public void pull_indexesUpdated()
...
+        do {
+            TimeUnit.MILLISECONDS.sleep(100);
+        } while (replicator.getState() == Replicator.State.PENDING);
+
+        // Assert that the replicator is started
+        Assert.assertEquals("The replicator should be started", Replicator.State.STARTED,
+                replicator.getState());
+
+        // Let the replicator run for a short time
+        TimeUnit.MILLISECONDS.sleep(500);
```
tomblench: Is it possible to assert/wait until some jobs have been created? I'm not sure if this would make the test too complex but it would be reassuring to know that at least one job was created.

ricellis: Added an assertion in 294cfd5.
Also reduced the overall number of docs and wrote them in bulk batches to improve test performance (from ~22 s to ~ 4 s).

tomblench:
+1

rhyshort:
+1

PR #234, opened 2016-02-22
Wrapped test assertion in isBulkSupported check

ricellis:
*What*
Stopped test asserting for tasks when using `_bulk_get`

*How*
Wrapped test assertion in `!isBulkSupported()` check

*Testing*
Change to the `stopRunningReplication` test.

reviewer @tomblench


tomblench:
+1

PR #236, opened 2016-02-25
Misc test cleanup

ricellis:
*What*
Fix some test cleanup/ignore issues.

*How*
The `CouchClientWrapperTest.accessAndUpdateRemoteDbWithSlashInName()` test set a new value to the `remoteDb` field during the test. This meant the DB that test created was cleaned up by the `teardown()` method, but the DB created by the `setup` and initially assigned to the `remoteDb` field was never removed.

Update the `.gitignore` naming for the copied test fixtures so that they don't show up as changes. This rename was missed as part of #193.

*Testing*
No new tests needed.

reviewer @rhyshort

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ build
 *.iws
 out
 local.gradle
-sync-core/fixture
+cloudant-sync-datastore-core/fixture
 **/.idea/*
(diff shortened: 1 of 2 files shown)


rhyshort:
+1

PR #237, opened 2016-02-29
Address FB 61611

brynh:
_What?_
Add test to reproduce a scenario similar to the crash reported in FB 61535. Update the code to throw an unchecked exception so that  further diagnostic information may be obtained on the underlying cause of FB 61535.

_Why?_
FB 61535 caused a crash for MIA. The underlying cause of the crash is not clear and is likely masked by the fact `Datastore.getDocumentsWithIds()` returns `null` in case of failure, which then leads to an NPE elsewhere.

How?
A test has been added to cause a failure in `Datastore.getDocumentsWithIds()` such that the existing code would have returned `null`. `Datastore.getDocumentsWithIds()` has been updated to throw a `DocumentException` retaining the details of the exception that caused it. Because `QueryResult.iterator()` implements the `Iterable` interface, we can't pass a `DocumentException` through that interface, so we encapsulate the `DocumentException` in an unchecked exception `UncheckedDocumentException` so that it can be thrown by `QueryResult.iterator()`. For further discussion of this, see FB 61611.

reviewer @mikerhodes
reviewer @rhyshort

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
@@ -615,7 +615,8 @@ class BasicDatastore implements Datastore, DatastoreExtended {
     }
 
     @Override
-    public List<DocumentRevision> getDocumentsWithIds(final List<String> docIds) {
+    public List<DocumentRevision> getDocumentsWithIds(final List<String> docIds)  throws
+            DocumentException {
         Preconditions.checkState(this.isOpen(), "Database is closed");
         Preconditions.checkNotNull(docIds, "Input document id list cannot be null");
         try {
@@ -632,13 +633,12 @@ class BasicDatastore implements Datastore, DatastoreExtended {
                 }
             }).get();
         } catch (InterruptedException e) {
-            logger.log(Level.SEVERE,"Failed to get documents with ids",e);
+            logger.log(Level.SEVERE, "Failed to get documents with ids", e);
+            throw new DocumentException("Failed to get documents with ids", e);
         } catch (ExecutionException e) {
             logger.log(Level.SEVERE, "Failed to get documents with ids", e);
+            throw new DocumentException("Failed to get documents with ids", e);
         }
-
-        return null;
-
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryResult.java
```diff
@@ -34,6 +35,18 @@
...
+     * This unchecked exception is used to wrap another exception so
+     * we can pass meaningful exception traces over the {@link #iterator()}
+     * boundary as we can't change the signature of {@link #iterator} to
+     * add a checked exception.
+     */
+    public class UncheckedDocumentException extends RuntimeException {
+        public UncheckedDocumentException(Exception causedBy){
+            super(causedBy);
+        }
+    }
```
mikerhodes: This is a bad choice of name. A class enclosed within QueryResult shouldn't be called something as generic as `*DocumentException`.

Given we need to wrap the exception to maintain compatibility with the existing interface, we should probably go for something that's more generic across the querying code. I'd go for something like `com.cloudant.sync.query.QueryException extends RuntimeException`. That name also implies putting it in it's own class.

We can add a TODO to make this exception checked in 2.0 if we want.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
```diff
@@ -633,12 +634,11 @@ public int compare(BasicDocumentRevision documentRevision, BasicDocumentRevision
             }).get();
         } catch (InterruptedException e) {
             logger.log(Level.SEVERE,"Failed to get documents with ids",e);
```
mikerhodes: Might as well add the spaces after the commas while we're in this code.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryResult.java
rhyshort: Making this exception checked would require alot of work due to the `Iterable` interface, so I think we should raise a ticket for that, and maybe try and get to it before 1.0

rhyshort:
Other than the name change mike mentioned +1

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryResult.java
rhyshort: renamed exception in 86abd42

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/BasicDatastore.java
rhyshort: done in dfe90e9

tomblench:
+1

mikerhodes:
+1

PR #239, opened 2016-03-04
Move to using HTTP classes from Java-Cloudant

rhyshort:
Move to using HTTP classes from Java-cloudant, this replaces our current
implementation of *all* classes in the HTTP package with classes provided
from java-cloudant.

* Set version number to 1.0.0-SNAPSHOT to better reflect what we are working towards
* Delete all com.cloudant.http.* classes
* New Version class for providing the User-Agent to the http layer, it is loaded at runtime.
* Removed header test, this is covered by java-cloudant tests

Note: for now we are using the a SNAPSHOT for cloudant-http 2.4, this will become stable with the
release of java-cloudant


ricellis:
+1

ricellis:
We should track the change from snapshot to non-snapshot though so we don't forget to update it and remove the snapshot repo.

rhyshort:
Yup, there is a ticket for that FB 62233

cloudant-sync-datastore-core/src/main/java/com/cloudant/library/LibraryVersion.java
```diff
@@ -0,0 +1,60 @@
+package com.cloudant.library;
```
mikerhodes: licence?

cloudant-sync-datastore-core/src/main/java/com/cloudant/library/LibraryVersion.java
```diff
@@ -0,0 +1,60 @@
...
+
+import com.cloudant.sync.util.Misc;
+import com.google.common.io.Resources;
+
+import java.net.URL;
+import java.util.Properties;
+
+/**
+ * Created by Rhys Short on 04/03/2016.
```
mikerhodes: proper docs

cloudant-sync-datastore-core/src/main/java/com/cloudant/library/LibraryVersion.java
rhyshort: fixed in 8d8401f

cloudant-sync-datastore-core/src/main/java/com/cloudant/library/LibraryVersion.java
rhyshort: fixed in 8d8401f

cloudant-sync-datastore-core/src/main/java/com/cloudant/library/LibraryVersion.java
```diff
@@ -0,0 +1,74 @@
...
+package com.cloudant.library;
+
+import com.cloudant.sync.util.Misc;
+import com.google.common.io.Resources;
+
+import java.net.URL;
+import java.util.Properties;
+
+/**
+ * A Class to provide the user-agent string to the HTTP layer.
```
mikerhodes: Be specific to _this_ classes role -- you're describing the interface here. "Provides a HTTP User-Agent string for CloudantSync"

build.gradle
```diff
@@ -12,7 +12,7 @@ apply plugin: 'signing'
 
 allprojects {
     group = 'com.cloudant'
-    version = '0.15.6-SNAPSHOT'
+    version = '1.0.0-SNAPSHOT'
```
mikerhodes: Does the version change belong here? At least a separate commit, as it's not directly related to the HTTP layer change.

cloudant-sync-datastore-core/src/main/java/com/cloudant/library/LibraryVersion.java
rhyshort: done in 8fed310

build.gradle
rhyshort: I'll split into a separate commit before merging, so I only need to make one force push

mikerhodes:
+1

PR #247, opened 2016-03-21
Add cookie auth on all replications

tomblench:
FB 51076

If username and password are specified in URL for ReplicatorBuilder, then add a CookieInterceptor and remove these credentials from the URL.

reviewer @brynh 
reviewer @alfinkel

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
@@ -14,16 +14,16 @@
 
 package com.cloudant.sync.replication;
 
+import com.cloudant.http.interceptors.CookieInterceptor;
 import com.cloudant.http.HttpConnectionRequestInterceptor;
 import com.cloudant.http.HttpConnectionResponseInterceptor;
 import com.cloudant.sync.datastore.Datastore;
 
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 
 /**
  * A Builder to create a {@link Replicator Object}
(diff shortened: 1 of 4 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java
```diff
@@ -90,12 +90,21 @@
 
     public PushAttachmentsInline pushAttachmentsInline = PushAttachmentsInline.Small;
 
+    // request and response interceptors are only held here for convenience - they are eventually
+    // used by the underlying CouchClient instance
+
+    protected List<HttpConnectionRequestInterceptor> requestInterceptors;
+
+    protected List<HttpConnectionResponseInterceptor> responseInterceptors;
+
```
rhyshort: Why are we holding the inceptors here? We don't use them anywhere in the class.

tomblench: Right now it's only for the benefit of the test cases. I wasn't entirely happy with this. Some alternatives are: widening the access to the fields on the CouchClient, using reflection in the test cases, not doing the asserts on the interceptors.

rhyshort: Since we want to eventually turn on findbugs for sync-android, I think using reflection might be the way to do it, but that seems a bit wrong just to avoid a findbugs warning of dead store.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
```diff
@@ -239,12 +257,23 @@ public void replicatorBuilderAddsCookieInterceptorDefaultPort() throws Exception
...
                         toString()
         );
-        // we know there will only be one interceptor and it will be first in the list
-        Assert.assertEquals(CookieInterceptor.class,
-                ((BasicPullStrategy)r.strategy).requestInterceptors.get(0).getClass());
-        Assert.assertEquals(CookieInterceptor.class,
-                ((BasicPullStrategy)r.strategy).responseInterceptors.get(0).getClass());
-        // TODO how to assert user/pass in cookie interceptor is correct
+        // peek inside these private fields to see that interceptors have been set
+        Field reqIField = ReplicatorBuilder.class.getDeclaredField("requestInterceptors");
```
rhyshort: I think we should refactor this code to be a helper method, so we can use it in `replicatorBuilderAddsCookieInterceptorCustomPort` too

rhyshort: It also appears the same code is present in PushReplicatorTest

tomblench: I'm generally less concerned about copy/paste in test code as the maintenance issues are different. Also where should this shared method go? There's no obvious class for it right now.

rhyshort: It should go in `com.cloudant.sync.util.TestUtils`or `com.cloudant.sync.replication.ReplicationTestBase`, it's more likely to be a better fit in the latter however.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
```diff
@@ -39,6 +39,46 @@
...
 
+    private URI addCookieInterceptorIfRequired(URI uri) {
+        String urlProtocol = uri.getScheme();
+        String urlHost = uri.getHost();
+        String urlPath = uri.getRawPath();
+
+        //Check if port exists
+        int urlPort = uri.getPort();
+        if (urlPort < 0) {
+            if (urlProtocol.equals("http")) {
```
brynh: This would be better written as `if("http".equals(urlProtocol)) {` so it can't NPE.
Similarly for the `else if` two lines later.

rhyshort: I don't think it can NPE, because the URL object should always have a protocol associated with it, the java doc also doesn't indicate that it can return null.

brynh: I'm pretty certain this can NPE. `uri.getScheme()` can return null if the `URI` has no scheme (see the [URI javadoc](http://developer.android.com/reference/java/net/URI.html#getScheme())), and since the URI comes from the library user's code rather than being something we're in control of, it could therefore have a null scheme.

rhyshort: Oh I see, I looked at the doc for `URL` not `URI` dam variable names, I think I'll change them while i'm there

rhyshort: fixed in 141e81e

brynh:
+1

alfinkel:
+1

Issue #246, opened 2016-03-16
Query: use enum for `indexType` in public API


tomblench:
Using an untyped string for `indexType` in `IndexManager` is confusing and error-prone.

- Deprecate methods which take a `String` for `indexType`
- Add enum for `IndexType` with values `json` and `text`
- Replace deprecated methods with new ones taking enum

See also #242 - this has caused confusion for at least one user.

PR #248, opened 2016-03-21
Use enum for `indexType` parameters in Public API

rhyshort:
## What
Use a enum instead of strings for index type parameters

## Why
Users have been confused with the index type parameter and have been using their own, restricting this to an enum will make the API easier to understand.

## How
- Created `IndexType` enum
- Change the parameter of type of `indexType` from `String` to `IndexType`

## Tests

No additional tests have been added, however some tests were removed as they are no longer needed.

## Reviewers 

reviewer @tomblench 
reviewer @alfinkel 

## Issues

Fixes: #246

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
@@ -72,7 +72,7 @@ class IndexCreator {
             return null;
         }
 
-        if (index.indexType.equalsIgnoreCase("text")) {
+        if (index.indexType == IndexType.TEXT) {
             if (!IndexManager.ftsAvailable(queue, database)) {
                 logger.log(Level.SEVERE, "Text search not supported.  To add support for text " +
                                          "search, enable FTS compile options in SQLite.");
@@ -120,7 +120,7 @@ class IndexCreator {
             if (existingIndexes != null && existingIndexes.get(index.indexName) != null) {
                 Map<String, Object> existingIndex =
                         (Map<String, Object>) existingIndexes.get(index.indexName);
-                String existingType = (String) existingIndex.get("type");
+                IndexType existingType = (IndexType) existingIndex.get("type");
                 String existingSettings = (String) existingIndex.get("settings");
                 List<String> existingFieldsList = (List<String>) existingIndex.get("fields");
                 Set<String> existingFields = new HashSet<String>(existingFieldsList);
@@ -153,7 +153,7 @@ class IndexCreator {
                 for (String fieldName: fieldNamesList) {
                     ContentValues parameters = new ContentValues();
                     parameters.put("index_name", index.indexName);
-                    parameters.put("index_type", index.indexType);
+                    parameters.put("index_type", index.indexType.toString());
                     parameters.put("index_settings", index.settingsAsJSON());
                     parameters.put("field_name", fieldName);
                     parameters.put("last_sequence", 0);
@@ -174,7 +174,7 @@ class IndexCreator {
                 }
 
                 List<String> statements = new ArrayList<String>();
-                if (index.indexType.equalsIgnoreCase(Index.TEXT_TYPE)) {
+                if (index.indexType == IndexType.TEXT) {
                     List<String> settingsList = new ArrayList<String>();
                     // Add text settings
                     for (String key : index.indexSettings.keySet()) {
@@ -284,11 +284,11 @@ class IndexCreator {
      */
     @SuppressWarnings("unchecked")
     protected static boolean indexLimitReached(Index index, Map<String, Object> existingIndexes) {
-        if (index.indexType.equalsIgnoreCase(Index.TEXT_TYPE)) {
+        if (index.indexType == IndexType.TEXT) {
             for (String name : existingIndexes.keySet()) {
                 Map<String, Object> existingIndex = (Map<String, Object>) existingIndexes.get(name);
-                String type = (String) existingIndex.get("type");
-                if (type.equalsIgnoreCase(Index.TEXT_TYPE) &&
+                IndexType type = (IndexType) existingIndex.get("type");
+                if (type == IndexType.TEXT &&
                     !name.equalsIgnoreCase(index.indexName)) {
                     logger.log(Level.SEVERE,
                             String.format("The text index %s already exists.  ", name) +
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
@@ -188,7 +188,7 @@ public class IndexManager {
             indexes = new HashMap<String, Object>();
             while (cursor.moveToNext()) {
                 String rowIndex = cursor.getString(0);
-                String rowType = cursor.getString(1);
+                IndexType rowType = IndexType.enumValue(cursor.getString(1));
                 String rowField = cursor.getString(2);
                 String rowSettings = cursor.getString(3);
                 if (!indexes.containsKey(rowIndex)) {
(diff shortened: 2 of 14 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexType.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Denotes the type for an Query Index.
+ */
+public enum IndexType {
+
+    /**
+     * JSON Index
+     */
+    JSON,
```
tomblench: How about calling these "json" and "text" - then their enum names would match the strings and you don't have to over-ride `toString` or provide `enumValue` (which appears to be doing a similar thing to the built-in `Enum.valueOf`)

The only downside is that it may break coding standards but in situations like these where we have to maintain compatibility with some existing strings I think it's worth it.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -255,7 +255,7 @@ public String ensureIndexed(List<Object> fieldNames, String indexName) {
      *  @param indexType The type of index (json or text currently supported)
      *  @return name of created index
      */
-    public String ensureIndexed(List<Object> fieldNames, String indexName, String indexType) {
```
tomblench: I had planned to deprecate the old String versions of the methods and leave them for compatibility. Any thoughts on that?

rhyshort: Since we are prior to 1.0 I thought it made sense to make the change breaking, and not carry the legacy through, otherwise we'd be stuck with it until 2.0, and based on the level of confusion caused I think its a safer bet to remove & replace for 1.0.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexType.java
rhyshort: I generally don't think we should break coding standards to maintain the compatibility. It would be expected that the enum values would be capitalised, and doing the thing of least surprise is probably a better option. However we should probably get some more input on this before deciding if we should break convention or not before merging.

alfinkel: FWIW, I agree with @rhyshort here and [here](https://github.com/cloudant/sync-android/pull/248/files/c253e6309a713c1c8c7f0cd207dec650917d785c#r56856390).

doc/query.md
```diff
@@ -74,28 +74,28 @@ Basic querying of fields benefits but does _not require_ a JSON index. For examp
...
 ```
 
 Use either of the following methods to create a TEXT index:
 
 ```
-ensureIndexed(List<Object> fieldNames, 
-              String indexName, 
+ensureIndexed(List<Object> fieldNames,
+              String indexName,
               String indexType)
```
alfinkel: Change `String` to `IndexType`.

doc/query.md
```diff
@@ -74,28 +74,28 @@ Basic querying of fields benefits but does _not require_ a JSON index. For examp
...
 
-// or 
+// or
 
-ensureIndexed(List<Object> fieldNames, 
-              String indexName, 
-              String indexType, 
+ensureIndexed(List<Object> fieldNames,
+              String indexName,
+              String indexType,
```
alfinkel: Change `String` to `IndexType`.

rhyshort: fixed in ab92940

doc/query.md
rhyshort: fixed in ab92940

doc/query.md
```diff
@@ -74,28 +74,28 @@ Basic querying of fields benefits but does _not require_ a JSON index. For examp
...
+The first argument, `fieldNames` is a list of fields to put into the index. The second argument, `indexName` is a name for the index. This is used to delete indexes at a later stage and appears when you list the indexes in the database.  The third argument, `indexType` defines what type of index to create. If not provided, the index type defaults to `json`.  The fourth argument, `indexSettings` is comprised of index parameters and their values.  Currently the only valid index setting is `tokenize` and it can only apply to a TEXT index.  If index settings are not provided for a TEXT index, the `tokenize` parameter defaults to the value `simple`.
```
alfinkel: In describing the third argument, I think we should specify that the index type defaults to JSON instead `json`.  We use the word TEXT to describe an index of type TEXT in the same paragraph, it only makes sense that we use JSON to describe the default for the index type.

rhyshort: done in 694f1a7

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexType.java
tomblench: OK it's a pain we can't override any of the appropriate `Enum.valueOf` methods, but what you're really doing here is (de)serialising between types. So we should have a standard way of doing that rather than each enum having its own (differently named) methods for converting to and from strings.

So I would like to suggest that this should this be registered as a proper jackson type. This probably seems like overkill for a 2-value enum, but I don't think we can afford to allow these inconsistent approaches in our code.

rhyshort: I don't think jackson is the right tool here, we only need to single string value for database insertion, if we used it for JSON I would agree, but from my reading of the code we don't, @alfinkel is probably in a better position than me to comment on it

tomblench: Right, how about do it the way it was supposed to be then? Upgrade the schema and store the int value from the enum.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -284,11 +284,11 @@ protected static boolean validFieldName(String fieldName) {
      */
     @SuppressWarnings("unchecked")
     protected static boolean indexLimitReached(Index index, Map<String, Object> existingIndexes) {
-        if (index.indexType.equalsIgnoreCase(Index.TEXT_TYPE)) {
+        if (index.indexType == IndexType.TEXT) {
             for (String name : existingIndexes.keySet()) {
                 Map<String, Object> existingIndex = (Map<String, Object>) existingIndexes.get(name);
                 String type = (String) existingIndex.get("type");
-                if (type.equalsIgnoreCase(Index.TEXT_TYPE) &&
+                if (IndexType.enumValue(type) == IndexType.TEXT &&
```
tomblench: Continuing the discussion from my previous comment, is this a case where we serialise the index type value which isn't related to database access?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexType.java
rhyshort: We can do that, it is a major amount of work however, since SQLite doesn't support dropping columns, so we need to rename and re add all the data. Do you think its worth the engineering effort to do?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
rhyshort: Nope, it's the value we got back from the DB, so we are just create the object for comparison, we could go the other way and turn the enum into the db string to compare.

tomblench: I was confused because the `existingIndexes` structure looks like some sort of deserialised JSON! Well I think that `listIndexesInDatabase` and similar methods should return more self-describing data types, and this should include the use of the Enum.

This way you can limit the conversion between enums and strings to the database specific code in `IndexManager`.

rhyshort: Done in 3c5e647, I think I got all the references, so when ever you interact with indexes you should get `IndexType` instead of `String`

tomblench:
+1

alfinkel:
+1

PR #249, opened 2016-03-22
Findbugs high

ricellis:
*What*
Enabled findbugs reporting at high level

*How*
Applied findbugs plugin to gradle build for main source sets.
Fixed existing high issues:
* Avoided unnecessary unboxing of long
* Ensured the charset encoding is specified
* Fixed broken `equals` method and added `hashCode` methods
* Added excludes for class names shadowing superclasses (another option would be to rename these classes).
* Fixed error message formatting

*Testing*
Findbugs passes for local `gradlew clean build`.
Existing unit and integration tests pass, no new tests added.

reviewer @tomblench 
reviewer @rhyshort

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 script: "./gradlew check integrationTest"
 
 jdk:
-  - openjdk6
+  - openjdk7
 
 services:
   - couchdb
--- a/cloudant-sync-datastore-core/findbugs_excludes.xml
+++ b/cloudant-sync-datastore-core/findbugs_excludes.xml
@@ -1,0 +1,31 @@
+<!--
+  ~ Copyright (c) 2016 IBM Corp. All rights reserved.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+  ~ except in compliance with the License. You may obtain a copy of the License at
+  ~
+  ~ http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software distributed under the
+  ~ License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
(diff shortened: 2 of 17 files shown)


rhyshort:
+1

cloudant-sync-datastore-core/findbugs_excludes.xml
```diff
@@ -0,0 +1,29 @@
...
+  -->
+<!--
+A list of known findbugs issues that we cannot fix at this time.
+-->
+<FindBugsFilter>
+
+    <!-- A couple of classes shadow the names of their supertypes.
+    -->
+    <Match>
+        <Bug code="Nm" pattern="NM_SAME_SIMPLE_NAME_AS_SUPERCLASS" />
```
tomblench: Should these be marked as a TODO or tracked in another ticket? I mean we're only excluding them to make findbugs pass, but we should fix the underlying problem eventually.

.travis.yml
```diff
@@ -3,7 +3,7 @@ sudo: false
 script: "./gradlew check integrationTest"
 
 jdk:
-  - openjdk6
+  - openjdk7
```
tomblench: So openjdk 7 is required for findbugs? I guess this is a reasonable upgrade, do we think there are any risks of compiling/running under a newer jdk?

rhyshort: The only risk is using APIs and language features not in 6, the android builds on jenkins should pick that up though

ricellis: Yes, findbugs requires 1.7 to run. I guess the risk is that we accidentally include some Java 7 API, but using the correct Java level in the IDE project settings and the Android testing should help protect us from that.

cloudant-sync-datastore-core/findbugs_excludes.xml
ricellis: Raised #252 and added a reference to it in the excludes in 45cc5f2.

tomblench:
+1

PR #250, opened 2016-03-23
60490 policies remote config

brynh:
*What*
Make Android replication policies easier to use when getting credentials for replication from a remote service.

*How*
The previous implementation required that the `getReplicators()` method was overridden in the concrete implementation to retrieve `Replicator` objects configured to perform the desired replications. It was always assumed this could be called from the `ReplicationService.onCreate()` method, which causes issues when an asynchronous call is required to get credentials in order to setup the `Replicator` objects. The new implementation removes the abstract `getReplicators()` method from the interface and instead requires the user to call `setReplicators(Replicator[])` before the service will respond to commands sent to it via the `Intent` mechanism. Any commands sent to the `ReplicationService` prior to `setReplicators(Replicator[])` being called will be queued and only processed once `setReplicators(Replicator[])` has been called. If consecutive identical commands are sent to the `ReplicationService` before `setReplicators(Replicator[])` has been called, the duplicates will be ignored.

*Testing*
New tests have been added to ensure that commands sent to the `ReplicationService` are not processed before `setReplicators(Replicator[])` has been called and that the queuing mechanism works as described above.

reviewer @ricellis
reviewer @alfinkel

### DIFF
--- a/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
+++ b/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
@@ -13,7 +13,9 @@ import android.os.Looper;
 import android.os.Message;
 import android.support.v4.content.WakefulBroadcastReceiver;
 
+import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 /**
@@ -32,6 +34,8 @@ public abstract class ReplicationService extends Service
 
     private Handler mServiceHandler;
     private ReplicationPolicyManager mReplicationPolicyManager;
+    private boolean mReplicatorsInitialised;
+    private List<Message> mCommandQueue = new ArrayList<Message>();
 
     /**
      * Stores the set of {@link ReplicationCompleteListener}s listening for replication complete
@@ -56,7 +60,7 @@ public abstract class ReplicationService extends Service
 
     public interface ReplicationCompleteListener {
         /**
-         * Callback to indicate that all replications specified by {@link #getReplicators(Context)}
+         * Callback to indicate that all replications passed to {@link #setReplicators(Replicator[])}
          * are complete.
          */
         void allReplicationsComplete();
@@ -153,18 +157,38 @@ public abstract class ReplicationService extends Service
         // Get the HandlerThread's Looper and use it for our Handler.
         Looper serviceLooper = thread.getLooper();
         mServiceHandler = getHandler(serviceLooper);
+    }
 
-        Replicator[] replicators = getReplicators(getApplicationContext());
-
+    /**
+     * Set the {@link Replicator} objects configured to perform the required replications.
+     * The {@code ReplicationService} will not begin replications until this method has
+     * been called. This operation should only be called once.
+     *
+     * @param replicators An array of the configured {@code Replicator} objects.
+     * @throws IllegalArgumentException if {@code replicators} is null or empty.
+     * @throws IllegalStateException if called again after a previous call to this method
+     *         with a valid array of {@code Replicator} objects.
+     */
+    public void setReplicators(Replicator[] replicators) {
+        if (mReplicatorsInitialised) {
+            throw new IllegalStateException("Replicators already set");
+        }
         if (replicators != null && replicators.length > 0) {
             if (mReplicationPolicyManager == null) {
                 mReplicationPolicyManager = new ReplicationPolicyManager();
             }
             mReplicationPolicyManager.addReplicators(replicators);
             mReplicationPolicyManager.setReplicationsCompletedListener(this);
+            synchronized (mCommandQueue) {
+                mReplicatorsInitialised = true;
+                for (Message msg : mCommandQueue) {
+                    mServiceHandler.sendMessage(msg);
+                }
+                mCommandQueue.clear();
+            }
         } else {
-            throw new RuntimeException(
-                    "No replications setup. Please return Replicators from getReplicators()");
+            throw new IllegalArgumentException(
+                    "No replications setup. Please pass Replicators to setReplicators(Replicator[])");
         }
     }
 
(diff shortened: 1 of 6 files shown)


cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
```diff
@@ -179,7 +190,19 @@ public int onStartCommand(Intent intent, int flags, int startId) {
             Bundle bundle = new Bundle();
             bundle.putParcelable(EXTRA_INTENT, intent);
             msg.setData(bundle);
-            mServiceHandler.sendMessage(msg);
+            synchronized (mCommandQueue) {
+                if (mReplicatorsInitialised) {
+                    mServiceHandler.sendMessage(msg);
+                } else {
+                    // Add the message to the command queue if it's different to the one at the
+                    // head of the queue. These messages will then be
```
ricellis: Isn't the code actually checking the message at the tail of the queue?

cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
```diff
@@ -459,4 +470,127 @@ else if (operationId == PeriodicReplicationService.COMMAND_STOP_REPLICATION && l
...
+            }
+        });
+
+        service.onStartCommand(intent, 0, 0);
+        try {
+            // Wait a bit to check the OperationStartedListener isn't called.
+            latch.await(1000, TimeUnit.MILLISECONDS);
+            assertEquals("CountDownLatch should not be decremented", latch.getCount(), 1);
+            service.setReplicators(mMockReplicators);
+            latch.await();
```
ricellis: I think this (and subsequent occurrences) would be better as something like:
 `assertTrue("The countdown should reach zero", latch.await(1, TimeUnit.MINUTES));`
That way at least if we end up on a failure path we aren't waiting indefinitely or relying on a global test timeout that could abort a build.

cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
```diff
@@ -459,4 +470,127 @@ else if (operationId == PeriodicReplicationService.COMMAND_STOP_REPLICATION && l
...
+
+        service.onCreate();
+        final CountDownLatch latch = new CountDownLatch(2);
+
+        service.setOperationStartedListener(new PeriodicReplicationService
+            .OperationStartedListener() {
+            @Override
+            public void operationStarted(int operationId) {
+                if (operationId == PeriodicReplicationService.COMMAND_START_REPLICATION) {
+                    latch.countDown();
```
ricellis: Is there any value in asserting that these commands come in the right order? Although the implementation is a `List` I guess it protects against changes to duplicate checking or something doing an insertion into the wrong place in future.
Maybe something like:
`if (latch.getCount() == 2){ latch.countDown() } else { fail("Wrong order")}` for start and similarly with a 1 count for the stop event. WDYT?

cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
```diff
@@ -459,4 +470,127 @@ else if (operationId == PeriodicReplicationService.COMMAND_STOP_REPLICATION && l
...
+        service.onCreate();
+        final CountDownLatch latch = new CountDownLatch(2);
+
+        service.setOperationStartedListener(new PeriodicReplicationService
+            .OperationStartedListener() {
+            @Override
+            public void operationStarted(int operationId) {
+                if (latch.getCount() == 2 && operationId == PeriodicReplicationService
+                    .COMMAND_START_REPLICATION) {
+                    latch.countDown();
```
ricellis: As per the previous case, maybe it is worth checking the ordering here too.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
```diff
@@ -153,18 +157,25 @@ public void onCreate() {
         // Get the HandlerThread's Looper and use it for our Handler.
         Looper serviceLooper = thread.getLooper();
         mServiceHandler = getHandler(serviceLooper);
+    }
 
-        Replicator[] replicators = getReplicators(getApplicationContext());
-
+    public void setReplicators(Replicator[] replicators) {
```
ricellis: What is the expected behaviour if this method is called twice? I'm a bit concerned that it appears to be a setter, but at present it will add additional replicators. I'm not sure what happened in the previous implementation if the `Replicator[]` returned from `getReplicators()` was modified.
I think we need to do one of these three things:

1. Don't allow the list to be changed and throw an `IllegalStateException` or similar if `setReplicators` is called twice.
2.  Have setter behaviour by creating a new `ReplicationPolicyManager` (or clear the existing replicators before calling `mReplicationPolicyManager.addReplicators` although it doesn't look like there is a way to do this at present).
3. Rename the method to `addReplicators` and allow replicators to be added.

I think for cases 2 or 3 some considerations may need to be made for how the command queue is processed if commands effectively apply to different groups of replicators.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
brynh: Fixed in 6d82437

cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
brynh: Good suggestion - fixed in abde67d.

cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
brynh: Addressed in 	24c9a47. Also updated some of the existing tests so they check that they only receive the commands they expect to rather than ignoring ones they weren't expecting and only responding to the ones they were expecting.

cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
brynh: This one does already check the ordering.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
brynh: Fixed in 7f971d3. Also added some additional tests and some javadoc to the `setReplicators()` and changed the existing `RuntimeException` to a more specific `IllegalArgumentException`.

cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
ricellis: Sorry, code blindness!

ricellis:
I'm +1 on this now, whether or not `@Test(expected = ...)` format gets used for the new tests.

brynh:
The `@Test(expected = IllegalArgumentException.class)` style annotation sadly still does not work in Android tests - I actually forgot that and wrote them that way initially but then had them fail when I ran them. The tests missing `@Test` are indeed getting picked up by virtue of being prefixed `test`.

alfinkel:
+1

PR #253, opened 2016-03-29
Fix FindBugs Class Shadowing issues

### LINKED ISSUE #252: Investigate classes that shadow simple name
`com.cloudant.sync.sqlite.ContentValues` and `com.cloudant.sync.util.Document` both have the same simple names as their super classes. This is a bad practice as it can be very confusing. We should decide if these classes are really necessary and remove them if they are not needed, or consider renaming them if they are.
When done the excludes can be removed from `﻿﻿cloudant-sync-datastore-core/findbugs_excludes.xml`.

rhyshort:
## What
Fix FindBugs class shadowing issues.

## How

Removed the following classes: 
- com.cloudant.sync.sqlite.ContentValues
- com.cloudant.sync.uti.Document

##  Testing

No additional testing

## Reviewers

reviewer @alfinkel 

Fixes #252


alfinkel:
+1

PR #255, opened 2016-03-30
Fix FindBugs issues 

rhyshort:
## What

Fix FindBugs issues that are raised when the reporting level is set to `medium`.

## Testing 

Run the following gradle command from root of the checkout, firstly at the head of master before
checking out the branch and then again, after the branch has been checked out. The number of warnings have been reduced,

`$ ./gradlew findBugsMain -Dfindbugs.report.level=medium`

## Reviewers
reviewer @alfinkel 
reviewer @mikerhodes

### DIFF
--- a/cloudant-sync-datastore-core/findbugs_excludes.xml
+++ b/cloudant-sync-datastore-core/findbugs_excludes.xml
@@ -19,8 +19,17 @@ A list of known findbugs issues that we cannot fix at this time.
 
     <Match>
         <Bug code="RV" pattern="RV_RETURN_VALUE_IGNORED_BAD_PRACTICE" />
-        <Class name="com.cloudant.sync.datastore.AttachmentManager"/>
-        <Method name="addAttachment" />
+        <Or >
+            <And>
+                <Class name="com.cloudant.sync.datastore.AttachmentManager"/>
+                <Method name="addAttachment" />
+            </And>
+            <And>
+                <Class name="com.cloudant.sync.datastore.DatastoreManager"/>
+                <Method name="&lt;init&gt;" />
+            </And>
+        </Or>
+
     </Match>
     <Match>
         <Bug code="RV" pattern="RV_NEGATING_RESULT_OF_COMPARETO" />
--- a/cloudant-sync-datastore-core/findbugs_excludes.xml
+++ b/cloudant-sync-datastore-core/findbugs_excludes.xml
@@ -1,0 +1,25 @@
+
+<!--
+  ~ Copyright (c) 2016 IBM Corp. All rights reserved.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+  ~ except in compliance with the License. You may obtain a copy of the License at
+  ~
+  ~ http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software distributed under the
+  ~ License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+  ~ either express or implied. See the License for the specific language governing permissions
+  ~ and limitations under the License.
+  -->
+<!--
+A list of known findbugs issues that we cannot fix at this time.
+-->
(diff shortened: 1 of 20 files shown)


alfinkel:
+1

cloudant-sync-datastore-core/findbugs_excludes.xml
```diff
@@ -28,6 +28,10 @@ A list of known findbugs issues that we cannot fix at this time.
                 <Class name="com.cloudant.sync.datastore.DatastoreManager"/>
                 <Method name="&lt;init&gt;" />
             </And>
+            <And>
+                <Class name="com.cloudant.sync.sqlite.SQLDatabaseQueue"/>
+                <Method name="updateSchema"/>
+            </And>
```
mikerhodes: Change `updateSchema` to use `execute` instead?

```java
queue.execute(new Runnable() {
    @Override
    public void run() {
        try {
            SQLDatabaseFactory.updateSchema(db, migration, version);
        } catch (SQLException e) {
            e.printStackTrace();
        }
    }
});
```

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/util/JSONUtils.java
```diff
@@ -34,7 +34,9 @@
 
     private static final String LOG_TAG = "JSONUtils";
 
-    public static final byte[] EMPTY_JSON = "{}".getBytes(Charset.forName("UTF-8"));
+    public static byte[] emptyJSONAsBytes(){
+        return "{}".getBytes(Charset.forName("UTF-8"));
+    }
```
mikerhodes: This isn't empty JSON, it's an empty JSON object (see suggested name in snippet below).

Can you be a bit more efficient here (and is FindBugs smart enough to accept), something like:

```java
private static final byte[] EMPTY_JSON = "{}".getBytes(Charset.forName("UTF-8"));

public static byte[] emptyJSONObjectAsBytes() {
    byte[] copy = new byte[EMPTY_JSON.length];
    System.arraycopy(EMPTY_JSON, 0, copy, 0, EMPTY_JSON.length);
    return copy;
}
```

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentRevisionBuilder.java
```diff
@@ -237,7 +237,7 @@ public static DocumentRevision buildRevisionFromMap(URI documentURI, Map<String,
 
         String docId = (String) map.get("_id");
         String revId = (String) map.get("_rev");
-        Boolean deleted = map.get("_deleted") == null ? false : (Boolean) map.get("_deleted");
+        Boolean deleted = map.get("_deleted") == null ? Boolean.FALSE : (Boolean) map.get("_deleted");
```
mikerhodes: It'd be good to see the failure message the FindBugs is spewing out for each of these fixes.

mikerhodes:
I'd consider collapsing the "Make X static" into a single commit, as it's the same small change to a bunch of things.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentRevisionBuilder.java
rhyshort: Added to commit 0b6db86

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/util/JSONUtils.java
rhyshort: Done in dbec83e however I opted to use `Arrays.copyOf` because I like using less lines :)

cloudant-sync-datastore-core/findbugs_excludes.xml
rhyshort: done in 80e73d1

rhyshort:
Squashed "Make X static" commits into 9674984

mikerhodes:
+1

PR #257, opened 2016-03-30
Implement ensureIndexed with index name generation

### LINKED ISSUE #245: Query: implement `ensureIndexed(List<Object> fieldNames)`
Most users don't care about index names and it should be straightforward enough to generate unique ones on the fly.

This is also relevant because it's easy enough for users to specify index names which will cause an error: see #244

rhyshort:
## What
Implement ensureIndex with index name generation.

## How
Generate a random index name when `ensureIndex(List<Object>)` is called.
The index name is generated with code inspired in the AttachmentManager
class which generates the attachment names. In the case of running out
of iterations before generating an unique attachment name, `null` is
returned. This causes ensureIndexed to behave as if it was passed a
`null` index name.

I'm unsure if asking the `IndexManager` for the current list of indexes for each iteration is a good 
idea some input on this specific point would be appreciated.

## Testing
A single test has been renamed and reworked to expect that the index name returned is not `null`

## Reviewers

reviewer @mikerhodes

## Issues

Fixes #245

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/Index.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/Index.java
@@ -61,7 +61,7 @@ class Index {
      * This method sets the index type to the default setting of "json"
      *
      * @param fieldNames the field names in the index
-     * @param indexName the index name
+     * @param indexName the index name or null
      * @return the Index object or null if arguments passed in were invalid.
      */
     public static Index getInstance(List<Object> fieldNames, String indexName) {
@@ -77,7 +77,7 @@ class Index {
      * Index object is valid.
      *
      * @param fieldNames the field names in the index
-     * @param indexName the index name
+     * @param indexName the index name or null
      * @param indexType the index type (json or text)
      * @param indexSettings the optional settings used to configure the index.
      *                      Only supported parameter is 'tokenize' for text indexes only.
@@ -92,8 +92,7 @@ class Index {
             return null;
         }
 
-        if (indexName == null || indexName.isEmpty()) {
-            logger.log(Level.SEVERE, "No index name was provided.");
+        if(indexName != null && indexName.isEmpty()){
             return null;
         }
 
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
@@ -17,11 +17,15 @@ import com.cloudant.sync.datastore.Datastore;
 import com.cloudant.sync.sqlite.SQLDatabase;
 import com.google.common.base.Joiner;
 
+import org.apache.commons.codec.binary.Hex;
+
+import java.nio.charset.Charset;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
@@ -37,6 +41,7 @@ class IndexCreator {
 
     private final SQLDatabase database;
     private final Datastore datastore;
+    private static Random indexNameRandom = new Random();
 
     private final ExecutorService queue;
 
@@ -63,16 +68,16 @@ class IndexCreator {
      *
      *  This function generates a name for the new index.
      *
-     *  @param index The object that defines an index.  Includes field list, name, type and options.
+     *  @param proposedIndex The object that defines an index.  Includes field list, name, type and options.
      *  @return name of created index
      */
     @SuppressWarnings("unchecked")
-    private String ensureIndexed(final Index index) {
-        if (index == null) {
+    private String ensureIndexed(Index proposedIndex) {
+        if (proposedIndex == null) {
             return null;
         }
 
-        if (index.indexType == IndexType.TEXT) {
+        if (proposedIndex.indexType == IndexType.TEXT) {
             if (!IndexManager.ftsAvailable(queue, database)) {
                 logger.log(Level.SEVERE, "Text search not supported.  To add support for text " +
                                          "search, enable FTS compile options in SQLite.");
@@ -80,7 +85,7 @@ class IndexCreator {
             }
         }
 
-        final List<String> fieldNamesList = removeDirectionsFromFields(index.fieldNames);
+        final List<String> fieldNamesList = removeDirectionsFromFields(proposedIndex.fieldNames);
 
         for (String fieldName: fieldNamesList) {
             if (!validFieldName(fieldName)) {
@@ -93,7 +98,7 @@ class IndexCreator {
         Set<String> uniqueNames = new HashSet<String>(fieldNamesList);
         if (uniqueNames.size() != fieldNamesList.size()) {
             String msg = String.format("Cannot create index with duplicated field names %s"
-                                       , index.fieldNames);
+                                       , proposedIndex.fieldNames);
             logger.log(Level.SEVERE, msg);
         }
 
@@ -110,29 +115,46 @@ class IndexCreator {
         // Then check whether the index already exists; return success if it does and is same,
         // else fail.
         try {
+
             Map<String, Object> existingIndexes = listIndexesInDatabaseQueue();
-            if (indexLimitReached(index, existingIndexes)) {
+
+            if(proposedIndex.indexName == null){
+                // generate a name for the index.
+                String indexName = IndexCreator.generateIndexName(existingIndexes.keySet());
+                if(indexName == null){
+                    logger.warning("Failed to generate unique index name");
+                    return null;
+                }
+
+                proposedIndex = Index.getInstance(proposedIndex.fieldNames,
+                                          indexName,
+                        proposedIndex.indexType,
+                        proposedIndex.indexSettings);
+            }
+
+
+            if (indexLimitReached(proposedIndex, existingIndexes)) {
                 String msg = String.format("Index limit reached.  Cannot create index %s.",
-                                           index.indexName);
+                                           proposedIndex.indexName);
                 logger.log(Level.SEVERE, msg);
                 return null;
             }
-            if (existingIndexes != null && existingIndexes.get(index.indexName) != null) {
+            if (existingIndexes != null && existingIndexes.get(proposedIndex.indexName) != null) {
                 Map<String, Object> existingIndex =
-                        (Map<String, Object>) existingIndexes.get(index.indexName);
+                        (Map<String, Object>) existingIndexes.get(proposedIndex.indexName);
                 IndexType existingType = (IndexType) existingIndex.get("type");
                 String existingSettings = (String) existingIndex.get("settings");
                 List<String> existingFieldsList = (List<String>) existingIndex.get("fields");
                 Set<String> existingFields = new HashSet<String>(existingFieldsList);
                 Set<String> newFields = new HashSet<String>(fieldNamesList);
                 if (existingFields.equals(newFields) &&
-                    index.compareIndexTypeTo(existingType, existingSettings)) {
-                    boolean success = IndexUpdater.updateIndex(index.indexName,
+                        proposedIndex.compareIndexTypeTo(existingType, existingSettings)) {
+                    boolean success = IndexUpdater.updateIndex(proposedIndex.indexName,
                                                                fieldNamesList,
                                                                database,
                                                                datastore,
                                                                queue);
-                    return success ? index.indexName : null;
+                    return success ? proposedIndex.indexName : null;
                 }
             }
         } catch (ExecutionException e) {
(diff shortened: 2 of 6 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -442,10 +448,53 @@ public Boolean call() {
...
+
+        String indexName = null;
+
+        int tries = 0;
+        while (tries < 200 && indexName == null) {
+            byte[] randomBytes = new byte[20];
+            indexNameRandom.nextBytes(randomBytes);
+            String candidate = new String(new Hex().encode(randomBytes), Charset.forName("UTF-8"));
+
+            Set<String> indexNames = listIndexes().keySet();
```
mikerhodes: Do this outside the loop.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -442,10 +448,53 @@ public Boolean call() {
...
+     * @return The generated index name or {@code null} if it failed.
+     *
+     */
+    private String generateIndexName() {
+
+        String indexName = null;
+
+        int tries = 0;
+        while (tries < 200 && indexName == null) {
+            byte[] randomBytes = new byte[20];
```
mikerhodes: No need to create a new array.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -442,10 +448,53 @@ public Boolean call() {
...
+            indexNameRandom.nextBytes(randomBytes);
+            String candidate = new String(new Hex().encode(randomBytes), Charset.forName("UTF-8"));
```
mikerhodes: Create the Hex object outside the loop?

rhyshort: done in b35d928

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
rhyshort: done in b35d928

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
rhyshort: done in b35d928

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -462,8 +462,7 @@ public boolean isTextSearchEnabled() {
      * of index names returned by {@link #listIndexes()} method.
      * This is because we avoid knowing about how indexes are stored in SQLite, however
      * this means that it is not thread safe, it is possible for a new index with the same
-     * name to be created after a copy of the indexes has been taken from the database. It must also
-     * be noted that this will list the indexes present for <strong>every</strong> iteration.
+     * name to be created after a copy of the indexes has been taken from the database.
```
mikerhodes: Hmm -- should this be done in a transaction?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -442,10 +448,53 @@ public Boolean call() {
...
+     *
+     * We allow up to 200 random name generations, which should give us many millions
+     * of indexes before a name fails to be generated and makes sure this method doesn't
+     * loop forever.
+     *
+     * @return The generated index name or {@code null} if it failed.
+     *
+     */
+    private String generateIndexName() {
```
mikerhodes: For me, this method belongs in the index creator class.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -110,6 +115,16 @@ private String ensureIndexed(final Index index) {
...
         // else fail.
         try {
+            if(index.indexName == null){
+                // generate a name for the index.
+                String indexName = this.generateIndexName();
+                if(indexName == null){
+                    return null;
+                }
+
+                index = Index.getInstance(index.fieldNames,indexName,index.indexType,index.indexSettings);
```
mikerhodes: spacing

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -143,6 +158,7 @@ private String ensureIndexed(final Index index) {
             return null;
         }
 
+        final Index indexToCreate = index;
```
mikerhodes: I'd be tempted to instead change the argument to `proposedIndex` and this variable to `index`, I think that's clearer.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -355,4 +371,47 @@ private String createVirtualTableStatementForIndex(String indexName,
...
+     * name to be created after a copy of the indexes has been taken from the database.
+     *
+     * We allow up to 200 random name generations, which should give us many millions
+     * of indexes before a name fails to be generated and makes sure this method doesn't
+     * loop forever.
+     *
+     * @return The generated index name or {@code null} if it failed.
+     *
+     */
+    private String generateIndexName() throws ExecutionException, InterruptedException {
```
mikerhodes: Can this be `private static string generateIndexName(Set existingNames)`? For me, it'd be nice if this method didn't do any database access -- make it a pure function with fewer side-effects.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -355,4 +371,47 @@ private String createVirtualTableStatementForIndex(String indexName,
...
+            if(!indexNames.contains(candidate)){
+                indexName = candidate;
+            }
+            tries++;
+        }
+
+        if (indexName != null) {
+            return indexName;
+        } else {
+            logger.warning("Failed to generate unique index name");
```
mikerhodes: I'd be tempted to log this on line 122, which adds a bit of self-documentation to why we're failing out there.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -158,7 +162,7 @@ private String ensureIndexed(Index index) {
             return null;
         }
 
-        final Index indexToCreate = index;
+        final Index proposedIndex = index;
```
mikerhodes: Ah you got this the wrong way around -- sorry for not being clear. I meant _this_ variable would be called `index` and the argument to `ensureIndexed` would be called `proposedIndex`.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -117,12 +117,16 @@ private String ensureIndexed(Index index) {
         try {
             if(index.indexName == null){
                 // generate a name for the index.
-                String indexName = this.generateIndexName();
+                String indexName = this.generateIndexName(listIndexesInDatabaseQueue());
```
mikerhodes: Pass in the `keySet()` here to be specific with the method's required input -- all `generateIndexName()` needs is the names of the indexes.

Also, for calling static methods, use `IndexCreator.generateIndexName` to make it clear. At least in IDEA, this is a warning too.

mikerhodes: Also, this is a duplicate database call -- there's another call to `listIndexesInDatabaseQueue()` just below.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -117,12 +117,16 @@ private String ensureIndexed(Index index) {
...
                 if(indexName == null){
+                    logger.warning("Failed to generate unique index name");
                     return null;
                 }
 
-                index = Index.getInstance(index.fieldNames,indexName,index.indexType,index.indexSettings);
+                index = Index.getInstance(index.fieldNames,
+                                          indexName,
+                                          index.indexType,
+                                          index.indexSettings);
```
mikerhodes: Apparently this could return `null` which is causing IDEA to get angry.

rhyshort: hmmm I don't get the warning in studio, are you using the latest release of IDEA?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
rhyshort: fixed in 9caa9c7

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
rhyshort: done in 9caa9c7

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
mikerhodes: 15.0.1 of IDEA. Could be a JavaSE compiler warning which the Android compiler doesn't kick out.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -355,4 +378,47 @@ private String createVirtualTableStatementForIndex(String indexName,
...
+     * We allow up to 200 random name generations, which should give us many millions
+     * of indexes before a name fails to be generated and makes sure this method doesn't
+     * loop forever.
+     *
+     * @param indexNames The names of the indexes that exist in the database.
+     *
+     * @return The generated index name or {@code null} if it failed.
+     *
+     */
+    private static String generateIndexName(Set<String> indexNames) throws ExecutionException, InterruptedException {
```
mikerhodes: I think `indexNames` should be `existingIndexNames`. Things like `if (!existingIndexNames.contains(candidate))` read more clearly to my mind.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -276,12 +274,12 @@ public String ensureIndexed(List<Object> fieldNames,
                                 IndexType indexType,
```
mikerhodes: Should we note in the JavaDoc which of the arguments can be null do you think?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
rhyshort: done in 1991fc2

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
rhyshort: done in 79496ee

mikerhodes:
+1

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/Index.java
```diff
@@ -61,7 +61,7 @@ private Index(List<Object> fieldNames,
      * This method sets the index type to the default setting of "json"
      *
      * @param fieldNames the field names in the index
-     * @param indexName the index name
+     * @param indexName the index name or null
```
ricellis: I'd like to see this description (and on line 80) extended to explain what happens in the `null` case, something like `or null to generate a unique index name`

rhyshort: I don't think its right to to say what happens in the `null` case when creating an index, as saying it here would make the user think that the indexname is generated when the Index object is created, which it isn't, so it would be misleading to the user if they were trying to create their own `Index` object.

ricellis: When I reached the end of the diff I did see that the behaviour is documented on the `ensureIndexed`, which is probably sufficient.

ricellis:
+1

Issue #214, opened 2015-11-09
Add new feature: Filters for Push Replication


kazbeel:
I propose the addition of the same feature I have found when using Couchbase Lite: Filter for Push Replications.  As you can see in its documentation this feature in pretty straightforward (http://goo.gl/JezE2X).

rhyshort:
Thanks for reporting this, we'l hopefully get this in for 1.0.0, so we have feature parity with CDTDatastore

PR #258, opened 2016-04-01
Enable Filtering of Push Replication

rhyshort:
## What
Make it possible to filter push replications.

## Why
This feature was requested by a user and brings sync-android closer to feature parity with CDTDatastore.

## How
- Added a new option to `ReplicatorBuilder`, to add filters to push replications.
- Added a new interface to be the integration point for user filter functions.

The replicator will write a checkpoint in the case where the changes feed had at least 1 change (even if that change was filtered out and therefore not pushed). This means that replications will not rewind in the event that a batch of changes are completely filtered out. If there were no changes at all then no checkpoint will be written.
The replication will complete when there are no more unfiltered changes.

## Reviewers
reviewer @ricellis
reviewer @tomblench 
reviewer @rhyshort - can you look over my modifications to your original

## Issues
- #214

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PushFilter.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PushFilter.java
@@ -1,0 +1,57 @@
+/*
+ *  Copyright (c) 2016 IBM Corp. All rights reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ *  except in compliance with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software distributed under the
+ *   License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ *  either express or implied. See the License for the specific language governing permissions
+ *  and limitations under the License.
+ */
+
+package com.cloudant.sync.replication;
+
+import com.cloudant.sync.datastore.DocumentRevision;
+/*
+ * Created by Rhys Short on 31/03/2016.
+ */
+
+/**
+ * <P>
+ * Interface for filtering documents during a push replication.
+ * </P>
+ * <P>
+ * Example usage - a filter for replicating only documents with an ID beginning with the letter a:
+ * </P>
+ * <pre>
+ * <code>
+ * replicatorBuilder.filter(new PushFilter(){
+ *             {@literal @}Override
+ *              public boolean shouldReplicateDocument(DocumentRevision revision) {
+ *                  if (revision.getId().startsWith("a") {
+ *                      return true;
+ *                  } else {
+ *                      return false;
+ *                  }
+ *              }
+ *     });
+ * </code>
+ * </pre>
+ *
+ * @api_public
+ * @see com.cloudant.sync.replication.ReplicatorBuilder.Push#filter(PushFilter)
+ */
+public interface PushFilter {
+
+    /**
+     * Determines if a DocumentRevision should be replicated to the remote database
+     *
+     * @param revision the DocumentRevision under consideration for replication
+     * @return true if the DocumentRevision should be replicated.
+     */
+    boolean shouldReplicateDocument(DocumentRevision revision);
+
+}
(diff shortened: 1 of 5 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/BasicPushStrategy.java
```diff
@@ -256,7 +256,36 @@ private void replicate()
...
+
+        List<DocumentRevision> allowedChanges = new ArrayList<DocumentRevision>(changes.getResults().size());
+
+        for (DocumentRevision revision : changes.getResults()){
+            if(this.filter.shouldReplicateDocument(revision)) {
+                allowedChanges.add(revision);
+            }
+        }
+
+        if(allowedChanges.size() == 0){
```
ricellis: This if/else block is redundant, both branches set changes to the same thing.

ricellis:
@rhyshort I was having a think about the checkpoint and the limitations you described above.
I'd like to propose these changes:
* Revert the `getNextBatch()` method to its previous content, such that it always gets the next batch of changes
* Move the filtering and checkpointing code into the `replicate()` method
* Although using the `last_sequence` instead of the size of changes is a possibility as we discussed yesterday, it might be easiest to use a second `Changes` object for the changes to process (i.e. filtered if necessary) as well as the original `Changes` object then the size check can still be valid. I did an example in 05aca56885061ff3e0a6b896dfc66ee4693e40d6

Probably will need a test for that last case.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PushFilter.java
```diff
@@ -0,0 +1,31 @@
...
+ *  either express or implied. See the License for the specific language governing permissions
+ *  and limitations under the License.
+ */
+
+package com.cloudant.sync.replication;
+
+import com.cloudant.sync.datastore.DocumentRevision;
+
+/**
+ * Created by Rhys Short on 31/03/2016.
```
tomblench: This can be marked @api_public

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/PushReplicatorTest.java
```diff
@@ -183,8 +185,135 @@ public void replicatorBuilderAddsCookieInterceptorDefaultPort() throws Exception
...
+     * Then it does another replication <b>without a filter</b>. We have made no more local changes
+     * so if the replicator starts from the correct checkpoint then there will be no changes and no
+     * more documents will be copied.
+     * The test will fail if the replicator starts from the wrong checkpoint because the 2 original
+     * changes would be visible after the second start() call.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testPushReplicationFilterContinuesFromCorrectPlace() throws Exception {
```
tomblench: Can you do asserts on the last saved sequence after each replication for these tests? Might be useful.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PushFilter.java
ricellis: Fixed in 8e2cd25. Also added a description of the interface and an example usage.

rhyshort:
+1

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/PushReplicatorTest.java
ricellis: Done in a14978a.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PushFilter.java
```diff
@@ -0,0 +1,57 @@
...
+ * <P>
+ * Interface for filtering documents during a push replication.
+ * </P>
+ * <P>
+ * Example usage - a filter for replicating only documents with an ID beginning with the letter a:
+ * </P>
+ * <pre>
+ * {@code
+ * replicatorBuilder.filter(new PushFilter(){
+ *              @Override
```
tomblench: tag should be `@override` (all lowercase) - this causes a javadoc build error.

tomblench: javadoc chokes on this - you need to use
```
{@literal @}Override
```

ricellis: Eventually wrangled it in 08e36bc.

tomblench:
+1

PR #265, opened 2016-04-18
EventBus API

ricellis:
#*What*

Remove Guava `EventBus` from the API to allow potential future removal of the (large) Guava dependency.

*How*

Added new `EventBus` class and `@Subscribe` annotation. Scan for subscriber methods and register a proxy on the Guava bus to invoke the subscribers.
Replace existing `EventBus` uses with new API, including some visibility changes since the new API invokes only visible methods (using `setAccessible` is not desirable

*Testing*

The notifications framework is covered in many existing tests notably:
`com/cloudant/sync/datastore/DatabaseNotificationsMoreTest.java`
`com/cloudant/sync/datastore/DatabaseNotificationsTest.java`
`com/cloudant/sync/datastore/DocumentNotificationsTest.java`
and many of the replication tests.
Whilst the new delegating `EventBus` could potentially use some unit tests, I expect us to remove it in the medium future and I don't think any unit tests would expose any problems not already found by the existing integration tests.

*Reviewers*

Thoughts on whether the new `EventBus` and `@Subscribe` should be in the existing `com.cloudant.sync.notifications` package - I put them in their own package since everything in notificaitons was a type you could subscribe to, but I'm still not sure it is worth isolating them.

reviewer @tomblench 
reviewer @rhyshort

### DIFF
--- a/doc/events.md
+++ b/doc/events.md
@@ -1,9 +1,12 @@
 # Events
 
 _This functionality is available in versions 0.3.0 and up._
+Note that prior to version 1.0.0 the Google Guava EventBus (`com.google.common.eventbus` package)
+API was used, but now there is an API provided in the `com.cloudant.sync.event` package. To migrate
+your code you will need to adjust your imports accordingly.
 
-Cloudant Sync uses [EventBus](https://code.google.com/p/guava-libraries/wiki/EventBusExplained) from
-the Google Guava library to post events to interested parties in a publish and subscribe fashion.
+Cloudant Sync uses an `EventBus` to post events to interested parties in a publish and subscribe
+fashion.
 
 The `Datastore` class posts events about Documents:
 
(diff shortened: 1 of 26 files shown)


rhyshort:
+1

doc/events.md
```diff
@@ -1,9 +1,11 @@
 # Events
 
 _This functionality is available in versions 0.3.0 and up._
+Note that prior to version 1.0.0 the Google Guava EventBus (`com.google.common.eventbus` package)
+API was used, but now there is an API provided in the `com.cloudant.sync.event` package.
```
tomblench: something like "to migrate your code you will need to adjust your imports occasionally"

ricellis: done in 18a1a8c replacing occasionally with accordingly

tomblench: oh haha, that's what i meant

tomblench:
+1

Issue #264, opened 2016-04-18
API cleanup (pre-1.0.0)


tomblench:
Clean up public APIs so they can be frozen for 1.0.0.

This includes removal of unused methods and moving internal methods out of public classes; this makes the public/internal boundary clearer and is less likely to cause code breakages for API consumers in future.

tomblench:
Done in pull request #266

PR #266, opened 2016-04-19
API cleanup

tomblench:
_What_
API cleanup as per issue #264

_How_
This is one of our last opportunities to ensure that the public facing API is correct before going to 1.0.0. After this point we would like to avoid breaking changes as this could imply a major version increment as per the semantic versioning rules. This PR addresses the following API issues:
* Naming: changing from the `Basicxxx` pattern to `xxxImpl` which is more java-friendly.
* Interfaces and inheritance: removal of un-needed interfaces (`DatastoreExtended`) and down-casting directly to concrete type (`DatatoreImpl`) when needed. Note that this down-casting is not ideal but is no worse than the previous situation (which would downcast to the `DatastoreExtended` interface).
* Appropriate visibility and access of methods: certain methods which are not 'public-facing' (internal) have been moved or had their accessibility modifiers changed to discourage or prohibit their use by library consumers. (See commits for details).

_Testing_
Existing tests fixed to use new class names where necessary. Also some tests renamed to reflect new class names.

reviewer: @ricellis 
reviewer: @rhyshort


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Changes.java
```diff
@@ -24,7 +24,7 @@
  * <p>{@code Changes} objects describe a list of changes to the datastore.</p>
  *
  * <p>The object contains a list of the changes between some sequence number
- * (passed to the {@link DatastoreExtended#changes(long, int)} method) and
+ * (passed to the {@link DatastoreImpl#changes(long, int)} method) and
```
rhyshort: Shouldn't this be pointing to `Datastore`?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Datastore.java
```diff
@@ -30,7 +29,7 @@
  * <p>The Datastore can be viewed as a pool of heterogeneous JSON documents. One
  * datastore can hold many different types of document, unlike tables within a
  * relational model. The datastore provides hooks, particularly in the
- * {@link DatastoreExtended} interface, which allow for various querying models
+ * {@link DatastoreImpl} class, which allow for various querying models
```
rhyshort: We shouldn't mention DatastoreImpl here

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Datastore.java
```diff
@@ -49,7 +48,7 @@
  * propagated, thereby resolving the conflicted document across the set of
  * peers.</p>
  *
- * @see DatastoreExtended
+ * @see DatastoreImpl
```
rhyshort: We shouldn't mention DatastoreImpl here

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Datastore.java
tomblench: Since DatastoreExtended has been retired, the 'extra' functionality not available through the Datastore interface now has to be accessed through DatastoreImpl.

That's what this clumsy sentence (which I didn't originally write) is trying to say - should I re-write it?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Datastore.java
tomblench: I don't think we can pretend it doesn't exist. Note that this documentation is also aimed at library developers as well as library consumers. Or do you think DatastoreImpl should have a stronger warning to tell library consumers not to use it directly?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Datastore.java
rhyshort: Yeah I think so.

ricellis:
As discussed I think we need some way to identify which classes are part of the API to help users understand what they can use and to help developers know what is semver major changes going forwards.

ricellis:
Reminder that following #265 this will need to look at `/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java` and probably mark the inner class as non-API.

ricellis:
Is it worth having a statement in the `README` or similar to make readers aware of the public/private tags? Although they are pretty self-explanatory once in the javadoc I wonder if it is worth calling them out up front.

This will also need a `CHANGES.md` entry.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
```diff
@@ -28,6 +28,9 @@
         void replicationErrored(int id);
     }
 
+    /**
```
ricellis: After talking to @brynh I think we should probably add a comment here pointing people at the `ReplicationsCompletedListener` so that if they think this is the class they want they can find what they really want :)

ricellis:
Nothing is jumping out at me as being incorrectly classified with the current list of public/private.

tomblench:
@ricellis that's good to know!

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
tomblench: @ricellis @brynh how about ccd93a1cc1bb3553d17026d83ce7546b6346049c ?

README.md
```diff
@@ -142,42 +142,60 @@ Once the libraries are added to a project, the basics of adding and reading
...
+reference for the library. Each jar contains the full javadoc for
+itself and the other jars - this is for convenience at the slight
+expense of duplication.
+
+Note that each class has an "API Status" declaration in the
+javadoc. This status is either "Public" or "Private". The general
+guidance is that API consumers (who are building software which uses
+this library) are discouraged from using "Private" API classes. This
+is because they may expose implementation details of the library which
+may be subject to change without notice.
```
ricellis: Do you think it is worth mentioning semver again here? Maybe something like
> subject to change without notice or without change to the major version number.

Although I guess it is covered by the changes content.

rhyshort: As the first thing that people see, it should be in README, there is a chance some people won't read the change log.

README.md
```diff
@@ -142,42 +142,60 @@ Once the libraries are added to a project, the basics of adding and reading
 a document are:
 
 ```java
-import com.cloudant.sync.datastore.DatastoreManager;
-import com.cloudant.sync.datastore.Datastore;
-import com.cloudant.sync.datastore.DatastoreNotCreatedException;
-import com.cloudant.sync.datastore.DocumentBodyFactory;
-import com.cloudant.sync.datastore.DocumentException;
-import com.cloudant.sync.datastore.DocumentRevision;
-import com.cloudant.sync.datastore.UnsavedFileAttachment;
```
rhyshort: I don't think we should remove the import statements, it makes it easier to for users to see where the classes (outside of the JDK) are coming from.

README.md
```diff
@@ -189,9 +207,6 @@ device the the remote database.
 Replication is simple to get started in the common cases:
 
 ```java
-import com.cloudant.sync.replication.ReplicatorBuilder;
-import com.cloudant.sync.replication.Replicator;
```
rhyshort: Again I don't think we should remove the import statements.

tomblench: With the import statements it's not a valid snippet of java. If we include the import statements we should also have the class/method definition.

rhyshort:
I don't think any of `com.cloudant.mazha` classes should be marked as API public exceptions or not, since we shouldn't be leaking those exceptions directly to user since they are an internal detail to how the mazha deals with connections to cloudant.

README.md
tomblench: see b4d642e

tomblench:
@rhyshort after auditing we realised that these don't leak out into the public API although the user could see them via runtime introspection. So I've made them API private in bd53eda

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Datastore.java
tomblench: see 	dda1d77

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Datastore.java
tomblench: see 	dda1d77

ricellis:
+1

rhyshort:
+1

PR #267, opened 2016-05-03
Fix sample app for 1.0.0

tomblench:
_What_

* proguard fixes
* upgrade build tools and gradle
* add android support library

_Why_

Sample app is currently broken, fix it before 1.0.0 release

_Testing_

Testing has been confirmed as follows - please follow this recipe to reproduce:

```shell
# in todo-sync
# build release version so proguard runs
./gradlew assembleRelease
# sign with debug key
jarsigner -verbose -keystore ~/.android/debug.keystore -storepass android -keypass android  ./build/outputs/apk/todo-sync-release-unsigned.apk androiddebugkey
# install onto currently running emulator or device
$ANDROID_HOME/platform-tools/adb install -r ./build/outputs/apk/todo-sync-release-unsigned.apk
# now run the sample on emulator or device and confirm that push and pull replication work OK
```

reviewer @ricellis
reviewer @rhyshort


ricellis:
Wow that is a lot simpler :)

Any thoughts on using `latest.release` for the Android dependencies? I am tempted from the perspective of not having to update the sample as much, although I guess it exposes the sample of risk to breaking when a new version is released.

tomblench:
@ricellis gradle, gradle support tooling, and android libraries often have to have matching (but different) versions. So I think we'll have to leave this for now but periodically ensure that all the numbers are right.

ricellis:
As I suspected, no freebies!

+1

rhyshort:
+1

PR #269, opened 2016-06-08
DatastoreManager uniqueness

ricellis:
*What*

Made it easier to obtain a singleton `DatastoreManager` for a storage path.

*Why*

The `DatastoreManager` expects to manage a storage location without anything else interfering with the files in that location. Exposing the constructor allows multiple `DatastoreManager` instances to be created for the same storage location. It is preferable to return a singleton instance for any given storage location (at least within the scope of the `DatastoreManager` class). This also has the advantage of allowing the same `DatastoreManager` instance to be retrieved in multiple places when the object cannot be passed around and only the path (string) is available (e.g. as used by the sync-android-p2p project).

*How*

Simplify locking for `Datastore` instance opening/creation/closing/deletion, removing double-lock pattern, which may have returned a `null` `Datastore` instance if `openDatastore` was called in parallel with a deletion.
Deprecate the public `DatastoreManager` constructors (leave them in place until the next major release for compatibility with existing code).
Added new static `getInstance` methods on `DatastoreManager` that create or return an existing `DatastoreManager` instance per storage path.

*Testing*

Added new tests:
* multithreadedDatastoreCreation
* datastoreInstanceNotReusedAfterClose
* assertFactoryReturnsSameInstanceString
* assertFactoryReturnsSameInstanceFile

Refactored existing test code to use `DatastoreManager.getInstance`.

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
@@ -34,7 +34,6 @@ import java.io.File;
 import java.io.IOException;
 import java.sql.SQLException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -62,7 +61,8 @@ public class DatastoreManager {
 
     private final String path;
 
-    private final Map<String, Datastore> openedDatastores = Collections.synchronizedMap(new HashMap<String, Datastore>());
+    /* This map should only be accessed inside synchronized(openDatastores) blocks */
+    private final Map<String, Datastore> openedDatastores = new HashMap<String, Datastore>();
 
     /**
      * The regex used to validate a datastore name, {@value}.
@@ -183,21 +183,21 @@ public class DatastoreManager {
      *
      * @see DatastoreManager#getEventBus()
      */
-    public Datastore openDatastore(String dbName, KeyProvider provider) throws DatastoreNotCreatedException {
+    public Datastore openDatastore(String dbName, KeyProvider provider) throws
+            DatastoreNotCreatedException {
         Preconditions.checkArgument(dbName.matches(LEGAL_CHARACTERS),
                 "A database must be named with all lowercase letters (a-z), digits (0-9),"
                         + " or any of the _$()+-/ characters. The name has to start with a"
                         + " lowercase letter (a-z).");
-        if (!openedDatastores.containsKey(dbName)) {
-            synchronized (openedDatastores) {
-                if (!openedDatastores.containsKey(dbName)) {
-                    Datastore ds = createDatastore(dbName, provider);
-                    ds.getEventBus().register(this);
-                    openedDatastores.put(dbName, ds);
-                }
+        synchronized (openedDatastores) {
+            Datastore ds = openedDatastores.get(dbName);
+            if (ds == null) {
+                ds = createDatastore(dbName, provider);
+                ds.getEventBus().register(this);
+                openedDatastores.put(dbName, ds);
             }
+            return ds;
         }
-        return openedDatastores.get(dbName);
     }
 
     /**
@@ -227,9 +227,9 @@ public class DatastoreManager {
         Preconditions.checkNotNull(dbName, "Datastore name must not be null");
 
         synchronized (openedDatastores) {
-            if (openedDatastores.containsKey(dbName)) {
-                openedDatastores.get(dbName).close();
-                openedDatastores.remove(dbName);
+            Datastore ds = openedDatastores.remove(dbName);
+            if (ds != null) {
+                ds.close();
             }
             String dbDirectory = getDatastoreDirectory(dbName);
             File dir = new File(dbDirectory);
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
@@ -37,6 +37,8 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.logging.Logger;
 
 /**
@@ -56,6 +58,10 @@ import java.util.logging.Logger;
  */
 public class DatastoreManager {
 
+    // Static map to manage DatastoreManager instances (one per directory path)
+    private static final ConcurrentMap<File, DatastoreManager> datastoreManagers = new
+            ConcurrentHashMap<File, DatastoreManager>();
+
     private final static String LOG_TAG = "DatastoreManager";
     private final static Logger logger = Logger.getLogger(DatastoreManager.class.getCanonicalName());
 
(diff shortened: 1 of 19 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
```diff
@@ -57,12 +58,17 @@
...
+    private static final ConcurrentMap<File, DatastoreManager> datastoreManagers = new
+            ConcurrentHashMap<File, DatastoreManager>();
+
     private final static String LOG_TAG = "DatastoreManager";
     private final static Logger logger = Logger.getLogger(DatastoreManager.class.getCanonicalName());
 
     private final String path;
 
-    private final Map<String, Datastore> openedDatastores = Collections.synchronizedMap(new HashMap<String, Datastore>());
+    /* This map should only be accessed inside synchronized(openDatastores) blocks */
```
tomblench: What's the tradeoff between 'manually' synchronising and using `Collections.synchronizedMap`? I gather the code previously did both which is obviously wrong.

ricellis: A `Collections.synchronizedMap` will synchronize all access to the map via the `Map` interface methods. This only guarantees that the elements are not modified during the operation so multiple callers see a consistent picture of the `Map`.
Caller 1: openDatastore("ds")
Caller 2: deleteDatastore("ds")
Caller 1 would get create the `Datastore` instance, and caller 2 would remove it from the Map and proceed to delete the files whilst caller 1 was also creating them. To avoid that problem we need to lock around the entire on disk action.
Once we have a `synchronized` block that extends around the on-disk work we can perform the `Map` access in the same block so there is no value in having a `Collections.synchronizedMap`. The locks are re-entrant so I don't think there is an issue with leaving it as a synchonizedMap if you prefer.
The real issue was the window presented by the double lock pattern, which could have released the lock for long enough (between the `contains` check and the `get`) for a `remove` operation to happen (as part of a delete) and result in the method returning `null`.

ricellis:
@tomblench also would appreciate any thoughts on the fact that we never remove instances from the map. My thought was that since it was one per storage location it was unlikely that an application would maintain more instances than could be safely stored and that the memory usage wouldn't grow out of control.
I guess the other option is to add a `removeInstance` or similar that closes all the `Datastore` instances opened by the instance before it is removed from the map, thus at least ensuring that anyone re-creating a `DatastoreManager` instance for the same storage location wouldn't be dealing with two lists of active `Datastore` objects.

tomblench:
@ricellis I don't see the storage overhead being a huge deal - as you say it's unlikely that users will want datastores in 1000s of locations in the lifetime of one application.

One thing you could maybe do is remove the `datastoreManagers` entry when the 'last' `Datastore` is closed - that is at the point when the manager is longer managing anything.

ricellis:
Removing the empty instance seems like a nice idea, but I think the synchonization becomes significantly more complex.
Consider a case with a `DatastoreManager` with a single opened `Datastore ds`.
Thread 1: is processing ds.close()
Thread 2 does: `DatastoreManager.getInstance()`
Thread 1: finishes closing `ds` and removes the `DatastoreManager` instance from the static map.
The next call to `DatastoreManager.getInstance()` will return a different (new) `DatastoreManager` instance than the one that was returned to Thread 2 and we end up with two instances in play, only one of which is tracked in the map.
To avoid that scenario I think we would have to combine the `openedDatastores` synchronization of the `DatastoreManager` instance with that of the global `datastoreManagers` map. I'll give it some more thought, but it seems like it could get pretty ugly fast WDYT?

ricellis:
@tomblench as we discussed and having thought about it a bit more I think it is better to just allow the map to grow. We could add a removal method, but I'm not sure it is necessary as I just don't think there will be that many storage locations being used within the same application.
Removing an instance starts to allow for the possibility that multiple `DatastoreManager` instances for the same location get created even if we get the locking right around deletion because references may be held elsewhere. Of course all of this is relatively weak if people use the constructor instead of `getInstance`, but we can't really eliminate that behaviour until the next major version.
I've added another test in 800ee63 to validate that the same `DatastoreManager` instance is returned for the same storage location from many threads.

ricellis:
@tomblench had another thought about protecting against use of the deprecated constructors. We could effectively delegate the method calls to the preferred instance obtained from `getInstance` using the newly constructed object as just a wrapper.  Whilst they wouldn't be the same instance the function would all go through the same `DatastoreManager` instance for the same storage location.
See 19b3d41acd520295bbe4b23076e40e1c4a4db668 - do you think this is worth adding?

tomblench:
@ricellis we still have a few grep hits for the deprecated constructor:

```
./cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java:157:        DatastoreManager manager = new DatastoreManager(directoryPath);
./cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/PullRepTest.java:26:            DatastoreManager dm = new DatastoreManager(path.getAbsolutePath());
./sample/todo-sync/src/com/cloudant/todo/Test.java:19:        DatastoreManager manager = new DatastoreManager(path.getAbsolutePath());
```

This also raises the question of reducing the warning count so that we have more visibility of deprecated usages in the future.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreManagerTest.java
```diff
@@ -213,4 +219,73 @@ public void listEmptyDatastore() throws Exception {
...
+                                return manager.openDatastore("sameDatastoreName");
+                            }
+                        };
+            }
+        }.run();
+    }
+
+    @Test
+    public void datastoreInstanceNotReusedAfterClose() throws Exception {
+        Datastore ds1 = manager.openDatastore("ds1");
```
tomblench: Is there any use having a test for the opposite - instance reused without closing?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/util/MultiThreadedTestHelper.java
```diff
@@ -0,0 +1,109 @@
...
+        this.results = Collections.synchronizedList(new ArrayList<T>(numberOfThreads));
+        this.exceptions = Collections.synchronizedList(new ArrayList<Throwable>(numberOfThreads));
+        for (int i = 0; i < numberOfThreads; i++) {
+            final Callable<T> callable = getCallable();
+            threads.add(new Thread(new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        latch.countDown();
+                        if (latch.await(20, TimeUnit.SECONDS)) {
```
tomblench: This might benefit from a comment like "wait for all theads to start" just to be clear that we're waiting on that and not the actual callable - I know you explained this to me before.

ricellis:
`./cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java:157:        DatastoreManager manager = new DatastoreManager(directoryPath);`
Fixed by adding a private constructor with a dummy parameter in e9711e7.

`./cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/PullRepTest.java:26:            DatastoreManager dm = new DatastoreManager(path.getAbsolutePath());`
`./sample/todo-sync/src/com/cloudant/todo/Test.java:19:        DatastoreManager manager = new DatastoreManager(path.getAbsolutePath());`
I don't have either of these classes on my branch, are they new?

Raised issue #274 for cleaning up all the build warnings that are unrelated to this.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreManagerTest.java
ricellis: The `multithreadedDatastoreCreation` test asserts that all the instances are the same, which covers the case of it being reused when it hasn't been closed.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/util/MultiThreadedTestHelper.java
ricellis: Added in 182aa42.

tomblench:
@ricellis sorry you're right those are untracked files I had in my directory from ages ago

tomblench:
+1

PR #270, opened 2016-06-17
Update README dependency examples

ricellis:
*What*
Provide separate maven/gradle examples for the different targets.

*How*
Make it more clear that the required dependencies will be pulled in automatically by clarifying the documentation and having separate examples for each target case.

*Testing*
Doc only change.


rhyshort:
+1

PR #300, opened 2016-06-30
Allow Queries which include booleans to execute

### LINKED ISSUE #273: Boolean query in IndexManager.find method issue
When searching over a Boolean field in documents matching results are not being returned when using the IndexManager.find method.

rhyshort:
## What
Allow Queries which include booleans to execute, for example:
`{ "married" : { "$eq" : true } }` previously these would be marked
incorrectly as invalid during the normalization and validation phase.

## How
- Update validation logic to include a boolean as an acceptable value.
- Correctly transform a boolean value into `1` or `0` to be used with SQLite queries.

## Testing
- New test added to check validation
- New test to make sure `QueryValidator` correctly normalises the  an `$eq` query with a boolean value.

## Issues

Fixes #273

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
@@ -484,7 +484,17 @@ class QuerySqlTranslator {
         // [ { "fieldName":  "mike"}, ...]
 
         List<String> whereClauses = new ArrayList<String>();
-        List<Object> sqlParameters = new ArrayList<Object>();
+        List<Object> sqlParameters = new ArrayList<Object>() {
+            @Override
+            public boolean add(Object o) {
+                if (o instanceof Boolean){
+                    Boolean bool = (Boolean) o;
+                    return super.add(bool ? "1" : "0");
+                } else {
+                    return super.add(String.valueOf(o));
+                }
+            }
+        };
 
         Map<String, String> operatorMap = new HashMap<String, String>();
         operatorMap.put(EQ, "=");
@@ -558,11 +568,14 @@ class QuerySqlTranslator {
                         sqlParameters.add(modulus.get(1));
                     } else {
                         // The predicate map value must be either a
-                        // String or a non-Float Number here.
+                        // String, a non-Float Number or Boolean here.
                         // This was validated during normalization.
+                        // Boolean values need to be converted into 1 or 0
+                        // to match SQLite expected values.
                         predicateValue = negatedPredicate.get(operator);
                         placeholder = "?";
-                        sqlParameters.add(String.valueOf(predicateValue));
+                        sqlParameters.add(predicateValue);
+
                     }
 
                     whereClause = whereClauseForNot(fieldName, sqlOperator, tableName, placeholder);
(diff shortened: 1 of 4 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -562,7 +562,12 @@ protected static SqlParts whereSqlForAndClause(List<Object> clause, String index
                         // This was validated during normalization.
```
ricellis: You need to update the comment above this, as it is not limited to String or non-Float now.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
```diff
@@ -562,7 +562,12 @@ protected static SqlParts whereSqlForAndClause(List<Object> clause, String index
...
                         predicateValue = negatedPredicate.get(operator);
                         placeholder = "?";
-                        sqlParameters.add(String.valueOf(predicateValue));
+                        if (predicateValue instanceof Boolean){
```
ricellis: Instead of writing this block twice, why not just subclass `ArrayList` and override `add` to do the appropriate translation for booleans? It would make it easier to add any other types or changes later.

ricellis:
Have you validated that there are query tests for all the types we index?

rhyshort:
There are tests that cover the types we index, and can use in selectors. But it is limited to `int` for the `Number` rather than going through each boxed type that is a subclass of number.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
rhyshort: done in 75a0785

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QuerySqlTranslator.java
rhyshort: It's an instance of `List` so its not something we should do, since it will run the risk of later using `ArrayList` only APIs which would make it harder to migrate to a different list implementation in the future.

rhyshort: We also currently do the entire set of if else statements again, which we should replace, so I think we could do something like that as part of a bigger refactoring.

ricellis: I meant (for example) like this, so the API wouldn't be an issue:
```java
        List<Object> sqlParameters = new ArrayList<Object>() {
            @Override
            public boolean add(Object o) {
                         if (o instanceof Boolean){
                            Boolean bool = (Boolean) o;
                            return super.add(bool ? "1" : "0");
                        } else {
                            return super.add(String.valueOf(o));
                        }
            }
        };
```
It would avoid the need to `add(String.valueOf(predicateValue))` in so many places as well, but we'd need to take care with the `add(modulus.get(0))` since they are `Integer`.

rhyshort: If we are going to do that, and remove the `String.valueOf` calls, we should push it as a single task for a refactor rather than doing it here, since we'd then be calling `String.valueOf(o)` multiple times.I'm not sure why we'd need to be mindful of the `Integer` types since wouldn't they drop into the else and use `String.valueof` anyway?

ricellis: My point was we don't call `String.valueOf()` on the `Integer` types at the moment, but we would be after the change, which I don't think matters because I think they end up in the SQL statement as strings anyway, but I wasn't 100% sure if they actually needed to be `Integer` in the parameter list.

rhyshort: done in 0154a94, we will still need to file a ticket to reduce duplication in the class though.

ricellis:
+1

Issue #286, opened 2016-06-22
Refactor IndexManager to use SQLDatabaseQueue instead of SQLDatabase


ricellis:
`IndexManager` should be refactored to use the `SQLDatabaseQueue` rather than establishing its own queue and using a SQLDatabase object to perform database operations.  The functionality is the same, but we need to ensure the `IndexManager` operations are routed through the same `SQLDatabaseQueue` thread/queue as other database operations throughout the code.

This should also fix the issue with index tests using multiple connections for SQLDatabase:
As is referenced in this comment: https://github.com/cloudant/sync-android/pull/89#discussion-diff-24152565 it should be correct to use the call:
`SQLDatabaseTestUtils.assertTablesExist(db, metadataTableList);`
instead of
`SQLDatabaseTestUtils.assertTablesExist(TestUtils.getDatabaseConnectionToExistingDb(db), metadataTableList);`
since db is actually set to `TestUtils.getDatabaseConnectionToExistingDb(db)` a few lines prior to the call.

To replicate the problem change line 52 in `AbstractIndexTestBase` to `SQLDatabaseTestUtils.assertTablesExist(db, metadataTableList);`

Then execute the `updateOneFieldIndex` test as a Junit test.

The problem is related to accessing the database from mutliple threads with SQLite4Java, we should make sure that whatever the tests are doing that the access goes through the correct SQLDatabase queue which protects against access from mutliple threads. It is probably beneficial to remove the `TestUtils.getDatabaseConnectionToExistingDb(db)` method to make sure that this is resolved for all tests.

Issue #284, opened 2016-06-22
Move IndexManager public API to Datastore


ricellis:
The IndexManager lifecycle is tied to that of a Datastore because it cannot exist without it. We should cement that relationship by moving the public methods from IndexManager to Datastore. This will also make the query experience on Android and iOS more consistent.

~~1.  Mark IndexManager as not API (this could be a major version change if for example making IndexManager a default visibility class, so will need some careful thought around deprecation or using the not API tags and backwards compatibility)
2.  Add a DatastoreIndexManager interface containing the public methods in the IndexManager
3.  DatastoreImpl to implement Datastore and DatastoreIndexManager
4. Close IndexManager when the datastore is closed~~

(See edit below)

tomblench:
Feature branch: `feature-284-refactor-datastore-indexmanager`

Inspired by the ideas above, the current thinking we've arrived as is:

- Changes will introduce breaking changes and necessitate major version increment to 2.
- Introduce a new class - `CloudantSync`, with overall lifecycle responsibility for CRUD (`Database`) and `IndexManager` objects. These names are still _definitely_ up for debate. Specifically `CloudantSync` could be confusing because it doesn't suggest that the data is 'local'. Also `Database` and `IndexManager` are inconsistent in that one has "Manager" in the name and one doesn't.
- Close `IndexManager` when `Database` is closed. Achieved by adding `close()` to `CloudantSync` class and removing public close methods for the classes it manages (so you can't close one without closing the other).
- Enable "purge" feature by `IndexManager` subscribing to `DocumentPurged` events from `Datastore`. Would allow implementation of #160 
- Use breaking API change as an opportunity to address indexing code shortcomings including #305 #366. This would tidy up method signatures and add `throws` clauses to allow checked exceptions to be thrown.

ricellis:
@tomblench can this close now?

tomblench:
@ricellis yes I think we've done everything now - this issue covers what became the `DocumentStore` class which manages `Database` and `Query` lifecycle in tandem.

ricellis:
Fixed by the merge of v2 stuff #501

Issue #305, opened 2016-07-04
Refactor indexing code


ricellis:
Refactoring the `Index` class code as well as code in the `IndexManager`, `IndexCreator`, and the `IndexUpdater` to better utilize the `Index` class.  The `Index` class needs to be decoupled from the other classes and be able to stand on its own.  We will look to:
1.  Implement as `Comparable` and utilize the `compareTo` method interface.
2.  ~~Possibly implement `IndexType` as an enumeration.~~ Done
3.  Refactor code blocks in the `IndexManager`, `IndexCreator`, and `IndexUpdater` that currently handle indexes as `Map<String, Object>` and instead solely deal with an index as an `Index` object.

@rhyshort FYI

rhyshort:
4. Make sure any newly public class are marked as API private in the doc annotations.

tomblench:
Change API methods to be more logical and consistent. Instead of all the different `ensureIndexed` we should have:

```java
ensureIndexed(List<FieldSort> fieldNames, String indexName, IndexType indexType, String tokenize)
```

- If the user wants an auto-generated name, set `indexName` `null`
- If the user is creating a JSON index, `tokenize` should be `null`
- If the user is creating a TEXT index, they can set `tokenize` as desired or `null` for the default tokenizer.

rhyshort:
I've got a couple of other options that take away the need to pass `null` into ensured index for `indexName` and `tokenize`.

```java
        // a builder like pattern
        ds.query().indexer().ensureIndex().named().fields().tokenize().isIndexed();
        
        IndexDefintion def = IndexDefiniton.json().fields().name("index1");
        def = IndexDefiniton.text().fields().name("index1").tokenize("example");
        Index created = ds.query().ensureIndexed(def);  // throws QueryException
```

Using `IndexDefinition` constructor(s) directly would simply shift the problem of `null` appearing 
in the call from `ensureIndexed` to `IndexDefinition`.   Another solution would be to have `Index` double for the definition and the representation of stored index.

I think switching to a builder pattern is the best way to go, it makes it clearer what the method is
doing at the call site, and a user mataining the code doesn't need to wonder "What are those `null`s for".

ricellis:
I think the builder pattern as shown is pretty confusing because it isn't immediately clear what each of these three methods are for:
`.indexer().ensureIndex().isIndexed()`

[...]

Issue #291, opened 2016-06-22
Finish refactoring of DatastoreImpl functionality into Callables


ricellis:
From @mikerhodes:

> So the first attempt, https://github.com/cloudant/sync-android/compare/54145-hoist-sqlqueuecallables, gives you a good idea of the aims of this task.
> 
> BasicDatastore is a huge class, which contains virtually all the logic to manage the MVCC semantics of the database. AttachmentManager breaks out some attachment logic, but really it's all still driven from BasicDatastore. What I'd like to do is to break up BasicDatastore into smaller components, which hopefully can have simpler tasks and can each be understood as a single unit doing a single task. Latterly, we might even be able to start testing things individually.
> 
> I think of this a bit like the operations in objective-cloudant -- each SQLCallable is a logical MVCC operation or perhaps lower level database operation. I'm not sure yet whether we want to more formally layer the MVCC semantics vs the database SQL semantics (which would allow us to more easily swap out storage layers later).
> 
> For example, https://github.com/cloudant/sync-android/pull/219 pulls out the work of inserting a single revision into the `revs` table. This might be too small a thing, but the logic is required by several of the higher level operations -- creating, updating and deleting -- so it makes sense to have separately.
> 
> Further from that, extracting the database interaction from the BasicDatastore "orchestration" logic would be a good thing. In the first instance, that won't really be possible. The first piece of work will be pulling out all the higher-level things like "create a document" into their own classes, which will do things like checking for MVCC validity (e.g., my update has the right revId of a current leaf node). It'd be nice long term if the database operations could be separated from the business logic.
> 
> For example, #219 is a storage level operation, almost. It might actually be slightly _too_ low-level, perhaps it should have some more logic of its own, but we'll see. At least we'll hopefully get to a stage where it's easier to conceptually see what's going on when we have separate classes. If not, then at least we learned that just making classes, however well encapsulated, doesn't help.

Note that `BasicDatastore` is now `DatastoreImpl`.

rhyshort:
#393 Introduces a few static inner classes which should be hoisted into their own files and consolidate as appropriate.

rhyshort:
PRs #460 #461 #463 #464  extract callables from the Query code.

rhyshort:
I think it's all done now so closing.

PR #307, opened 2016-07-05
Initial Query refactor

rhyshort:
## What

Initial refactor of Query code to bring the function from `IndexManager` into the `Datastore`

## How

- deprecate the `IndexManager`
- add the `IndexManager` public methods to the datastore.
- Use the SQLDatabaseQueue rather than manage the database access directly.
  - Remove direct use of transaction methods, and use the `submitTransaction` method on the queue.
- Move function from `IndexManager` to `QueryableDatastore` 
- Add a Temporary checked exception `CheckedQueryException` to be a place holder for better Query error handling.
- Index Manager now just proxies calls to the `Datastore` catching and logging any exceptions and returning `null`.

## Testing

All tests pass, they were also updated to use SQLDatabaseQueue rather than an executor service and SQLDatabase directly.

## Notes

There is still a lot to do, hence aiming to merge this with a feature branch before continuing.
There are several `FIXME` items related to Checked Exception These will be looked at as part of improvements to Query Error handling in a later PR, before the feature branch is merged to master.

## Issues

* #286 
* #284


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Datastore.java
```diff
@@ -338,5 +342,95 @@ void resolveConflictsForDocument(String docId, ConflictResolver resolver)
      */
     void compact();
 
+    /**
```
ricellis: As per #284 I think these methods should be part of another interface.

rhyshort: Just to some up from our conversation, we're going to put the query methods on the Datastore interface, in order to make it more inline with CDTDatastore.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/QueryableDatastore.java
```diff
@@ -0,0 +1,282 @@
+package com.cloudant.sync.datastore;
```
ricellis: Copyright statement needed.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
```diff
@@ -333,7 +333,7 @@ private Datastore createDatastore(String dbName, KeyProvider provider) throws Da
             // if it does not exist
 
             //Pass database directory, database name, and SQLCipher key provider
-            DatastoreImpl ds = new DatastoreImpl(dbDirectory, dbName, provider);
+            QueryableDatastore ds = new QueryableDatastore(dbDirectory, dbName, provider);
```
ricellis: I don't like this because of the upfront work of the `QueryableDatastore` constructor. We shouldn't be creating the index tables etc in case they are never used.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/QueryableDatastore.java
```diff
@@ -0,0 +1,282 @@
...
+
+
+    private static final String EXTENSION_NAME = "com.cloudant.sync.query";
+    private static final String INDEX_FIELD_NAME_PATTERN = "^[a-zA-Z][a-zA-Z0-9_]*$";
+
+    private static final Logger logger = Logger.getLogger(QueryableDatastore.class.getName());
+
+    private final Pattern validFieldName;
+
+    private final SQLDatabaseQueue dbQueue;
```
ricellis: I think this should be renamed to make it clear that this `dbQueue` is for the index databases, not the datastore that this extends.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/CheckedQueryException.java
```diff
@@ -0,0 +1,8 @@
+package com.cloudant.sync.query;
+
+/**
+ * Created by rhys on 04/07/2016.
+ */
+//FIXME: Give this a propername
```
ricellis: I guess you are planning to do this FIXME as part of the exception handling merge to the branch? I think something like `DatastoreQueryException` would be a good superclass for all of these query based exceptions. Probably better than just `QueryException` as that could be confused as SQL or even a server side query issue.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -38,27 +41,24 @@
...
     private final Datastore datastore;
     private static Random indexNameRandom = new Random();
 
-    private final ExecutorService queue;
+    private final SQLDatabaseQueue queue;
 
     private static final Logger logger = Logger.getLogger(IndexCreator.class.getName());
 
-    public IndexCreator(SQLDatabase database, Datastore datastore, ExecutorService queue) {
+    public IndexCreator(Datastore datastore, SQLDatabaseQueue queue) {
```
ricellis: Given that we would now rely on the `Datastore` being a `QueryableDatastore` and that this class is not API I suggest that the constructor is changed to accept a `QueryableDatastore`.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
```diff
@@ -79,7 +79,7 @@ private String ensureIndexed(Index proposedIndex) {
         }
 
         if (proposedIndex.indexType == IndexType.TEXT) {
-            if (!IndexManager.ftsAvailable(queue, database)) {
+            if (!((QueryableDatastore)datastore).ftsAvailable(queue)) {
```
ricellis: This cast can be removed if the constructor and local `datastore` field are updated.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/QueryableDatastore.java
```diff
@@ -0,0 +1,282 @@
...
+            logger.log(Level.SEVERE,"Failed to list indexes",e);
+            throw new RuntimeException(e);
+        } catch (ExecutionException e) {
+            logger.log(Level.SEVERE,"Failed to list indexes",e);
+            throw new RuntimeException(e);
+        }
+
+    }
+
+    public static Map<String, Object> listIndexesInDatabase(SQLDatabase db) {
```
ricellis: I know this is mostly a copy/paste, but I don't see the point of this static method I think a `ListIndexesCallable` would be better. Although I guess that probably falls into the remit of #291.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/QueryableDatastore.java
```diff
@@ -0,0 +1,282 @@
...
+                indexType,
+                indexSettings),
+                this,
+                dbQueue);
+    }
+
+    @Override
+    public boolean deleteIndexNamed(final String indexName) throws CheckedQueryException {
+        if (indexName == null || indexName.isEmpty()) {
+            logger.log(Level.WARNING, "TO delete an index, index name should be provided.");
```
ricellis: Fix capitalization in `TO`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/QueryableDatastore.java
```diff
@@ -0,0 +1,282 @@
...
+    @Override
+    public QueryResult find(Map<String, Object> query) throws CheckedQueryException {
+        return find(query, 0, 0, null, null);
+    }
+
+    @Override
+    public QueryResult find(Map<String, Object> query, long skip, long limit, List<String>
+            fields, List<Map<String, String>> sortDocument) throws CheckedQueryException {
+        if (query == null) {
+            logger.log(Level.SEVERE, "-find called with null selector; bailing.");
```
ricellis: Any idea why this has a `-` prefix?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
rhyshort: I don't think the overhead of the index tables is really going to be high enough for us to worry about, do you have a better way of going about it?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/CheckedQueryException.java
rhyshort: Yup, as per the notes in the description

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
rhyshort: I don't think that is the best approach, it only needs the queue to perform the operations, it doesn't require any of the code from Datastore, so its introducing coupling that isn't required.

rhyshort: Actually scratch that.

rhyshort: I don't think we should couple it to an implementation, I not keen on have to cast later on, but I think that can be fixed with some refactoring of the indexing code, by doing the work during construction to figure out if we can use FTS when we create the database rather than running the check each them, so when calling the FTS method to see if it's there we would know ahead of time. Or perhaps moving said check to the Database Queue, since its a property more todo with the database than the datastore itself.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/QueryableDatastore.java
rhyshort: Yeah it would either be #291 or #305

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/QueryableDatastore.java
rhyshort: Probably a copy from objc, `-` indicates an instance method.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -71,80 +74,24 @@
...
-                }
-            }).get();
-        } catch (InterruptedException e) {
-            logger.log(Level.SEVERE,"Failed to close db",e);
-            throw new RuntimeException(e);
-        } catch (ExecutionException e) {
-            logger.log(Level.SEVERE, "Failed to close db", e);
-        }
-        queue.shutdown();
+        // do nothing.
```
ricellis: I think it would help anyone debugging into this method to understand _why_ it does nothing. Please add some additional explanation here - something along the lines of "The index lifecycle is now combined with the datastore lifecycle so this deprecated method need not close anything as the index will be shutdown when the datastore is closed."

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/QueryableDatastore.java
rhyshort: done in d8a9709

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/QueryableDatastore.java
rhyshort: done in 88b9c1f

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/QueryableDatastore.java
rhyshort: done in aa2b930

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
rhyshort: done in 7f65a0e

ricellis:
I think the query package classes that have been made public need to be tagged with the API private stuff so as to be marked appropriately in the docs.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
ricellis: If you have a cast it is coupled to the implementation so you may as well be up front about it rather than allowing someone to accidentally provide some other Datastore impl that throws a CCE later. Obviously if the FTS check is moved then it doesn't apply.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
ricellis: I guess there are two aspects to this, i) using up extra space by creating things that may never be used and ii) the time of execution to do so.
Regarding i) when we talked about this @tomblench mentioned it could be potentially confusing in debugging if this stuff was automatically created even if indexing wasn't being used. I'm sure we can deal with that, and if we're talking a few bytes vs a few MB then it probably isn't worth the hassle of deferring.
Regarding ii) it could change the startup profile for apps that may currently intentionally defer the `IndexManager` creation to some later stage. I guess it depends enormously on how long it actually takes to do this stuff.

You could add an init and boolean check, but it would be annoying to remember to add the init check for every query method call unless we integrated it into a superclass of all the necessary callables to keep it in one place for example.

Another option would be if DatastoreImpl was reverted to non-abstract then it could have a private method for getQueryableDatastore that could do the init check and initiatilization and then the delegated method calls could be `getQueryableDatastore().find` etc.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/AbstractIndexTestBase.java
```diff
@@ -29,38 +31,37 @@
...
-        db = TestUtils.getDatabaseConnectionToExistingDb(im.getDatabase());
-        assertThat(db, is(notNullValue()));
-        assertThat(im.getQueue(), is(notNullValue()));
-        String[] metadataTableList = new String[] { IndexManager.INDEX_METADATA_TABLE_NAME };
-        SQLDatabaseTestUtils.assertTablesExist(TestUtils.getDatabaseConnectionToExistingDb(db),
-                                               metadataTableList);
+        fd = new ForwardingDatastore(ds);
+        dbq = ds.getQueryQueue();
+        assertThat(fd, is(notNullValue()));
+//        db = TestUtils.getDatabaseConnectionToExistingDb(im.getDatabase());
```
ricellis: If the code is dead delete it.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/AbstractIndexTestBase.java
```diff
@@ -29,38 +31,37 @@
...
     }
 
     @After
     public void tearDown() {
-        im.close();
-        assertThat(im.getQueue().isShutdown(), is(true));
-        TestUtils.deleteDatabaseQuietly(db);
         ds.close();
+//        assertThat(im.getQueue().isShutdown(), is(true));
+//        TestUtils.deleteDatabaseQuietly(db); // FIXME make this work
```
ricellis: Does this actually need fixing?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/ForwardingDatastore.java
```diff
@@ -0,0 +1,181 @@
...
+import com.cloudant.sync.event.EventBus;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Created by rhys on 04/07/2016.
+ */
+public class ForwardingDatastore implements Datastore {
```
ricellis: What is the point of this class, I can't see any method interception going on here only pure delegation?
If we need something that exposes the `Datastore` interface only but not the impl, why not just do
`Datastore fd = ds;`?
Also if we do need this it would be much simpler to implement with a `java.lang.reflect.Proxy` and an `InvocationHandler`.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/AbstractQueryTestBase.java
```diff
@@ -56,20 +58,19 @@ public void setUp() throws Exception {
...
+        dbq = ds.getQueryQueue();
         assertThat(ds, is(notNullValue()));
     }
 
     @After
     public void tearDown() {
-        im.close();
-        assertThat(im.getQueue().isShutdown(), is(true));
-        TestUtils.deleteDatabaseQuietly(db);
+//        fd.close();
```
ricellis: dead code?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
```diff
@@ -70,7 +71,7 @@ public void tearDown() {
     @Test
     public void updateIndexNoIndexName() throws Exception {
         createIndex("basic", Arrays.<Object>asList("name"));
-        assertThat(IndexUpdater.updateIndex(null, fields, db, ds, im.getQueue()), is(false));
+//        assertThat(IndexUpdater.updateIndex(null, fields, db, ds, im.getQueue()), is(false));
```
ricellis: Do we need some new assertion here?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
```diff
@@ -246,52 +268,59 @@ public void updateTwoFieldIndex() throws Exception {
...
+            public Void call(SQLDatabase db) throws Exception {
+                Cursor cursor = null;
+                try {
+                    cursor = db.rawQuery(sql, new String[]{});
+                    assertThat(cursor.getCount(), is(0));
+                } finally {
+                    DatabaseUtils.closeCursorQuietly(cursor);
+                }
+
+                    return null;
```
ricellis: too far indented?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/ForwardingDatastore.java
ricellis: I found the override usage in the MockMatcher now. A reflect proxy and invocation handler would still be a lot less code.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexCreator.java
rhyshort: Ideally I like it to be moved, since its a database property rather a datastore one

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/ForwardingDatastore.java
rhyshort: Done in d555987

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/AbstractIndexTestBase.java
rhyshort: Not sure, since the directory is deleted, it'll probably make more sense to just make sure it's closed and then delete.

rhyshort:
@ricellis w.r.t. to tagging public classes, as api private I'd like to defer that until later, so it can be determined if they can be lowered in scope again as part of the refactor.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
rhyshort: Having the `DatastoreImpl` delegate to `Queryable` would cause abstraction problems. Since we'd have to decouple them and `Queryable` relies on Datastore APIs, which is why I selected a subclass. 

w.r.t. start up time, the update schema methods are async, so shouldn't be an issue. FTS is sync, however but if that goes down onto the `SQLDatabaseQueue` we could push the check to when we access the property for the first time. Otherwise there is no real work occurring in the constructor, e.g. we don't update indexes or anything.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/AbstractQueryTestBase.java
```diff
@@ -63,6 +67,12 @@ public void setUp() throws Exception {
         assertThat(ds, is(notNullValue()));
     }
 
+    protected Datastore proxy(InvocationHandler handler) throws Exception {
+        Class proxy = Proxy.getProxyClass(Datastore.class.getClassLoader(), Datastore.class);
+        Constructor constructor = proxy.getConstructor(InvocationHandler.class);
+        return (Datastore) constructor.newInstance(handler);
```
ricellis: The shorter pattern for this is:
```java
(Datastore) Proxy.newProxyInstance(Datastore.class.getClassLoader(), new Class<?>[]{Datastore.class}, handler);
```

ricellis:
@rhyshort I can live with that but please add it to whichever github issue you intend to do it under so it is not missed before the query refactoring branch merges back to master.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
rhyshort: fixed in 9da5dfd

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/AbstractQueryTestBase.java
rhyshort: done in 006243b

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
rhyshort: done in 6beab50

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/AbstractQueryTestBase.java
rhyshort: removed in c778cc3

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/AbstractIndexTestBase.java
rhyshort: removed in ba3d602,I'm letting the directory delete remove the database

rhyshort:
api private tags to be done as part of #305

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreManager.java
tomblench: I'm afraid I agree with @ricellis on his delegation point. I don't see how it makes sense for `DatastoreImpl` to be abstract (leaving naming conventions to one side). The problem of using inheritance is that the java model of single inheritance isn't the right tool for the job. It means that your class hierarchy is inverted - each sub-class is increasingly specialised whereas what we want is to add functionality with each layer (there is a subtle difference).

Consider that we might have other 'managers' with functionality analogous to Index/Query. We might have a geo-spatial manager which adds extra functionality, and we might also have (hypothetically) a large attachment manager which switches storage technologies so that attachments are stored somewhere other than Cloudant on the server side.

Depending on what capabilities the user wants, we now have:

```
DatastoreImpl
QueryableDatastore
GeospatialDatastore
LargeAttachmentDatastore
```

The user has to choose the one they want. But wait, we're not done, what if they want to combine the features?

```
QueryableGeospatialDatastore
QueryableLargeAttachmentDatastore
GeospatialLargeAttachmentDatastore
QueryableGeospatialLargeAttachmentDatastore
```

Well I think you can see where this is going. I honestly think that inheritance isn't the right tool for the job. Some sort of class composition as suggested by Rich would be better. There are a few options here, which maybe include the [Facade Pattern](https://en.wikipedia.org/wiki/Facade_pattern) (this is better explained in the _Design Patterns_ book).

The downside with the Facade Pattern is that we have to write a fair bit of boilerplate proxying calls to the other components ("sub-system classes", like `IndexManager`), but on the upside it has high usability for the API consumer. The alternative delegation method suggested like `getQueryableDatastore().find` requires a bit more typing by the API consumer but requires less boilerplate on our part.

I also think it might be worth prototyping some of these approaches before generating any larger diffs.

Comments welcome!

ricellis: As you know I am in favour of the composition of the implementations. I think when we were in discussion the `getQueryableDatastore().find()` method was seen as adding too much complexity for users (although it is still simpler than the current `IndexManager`.

The facade pattern seems like a reasonable compromise, keeping the interface simple for the user as `Datastore` with all available methods, but the complexity of implementation broken down into different internal types.
I guess this would look like `DatabaseImpl` constructing the various instances required (e.g. `QueryableDatastore` and delegating the interface calls. If we wanted to reduce the boiler plate we could use the `java.lang.reflect.Proxy` to direct to the necessary objects, but obviously that would introduce complexity of its own, for example, in identifying which method needed to be invoked where.

rhyshort: I think you missed the point of using the inheritance in this case. The user would never see the back ends you have listed. The simple idea for using the inheritance model was to cut the line count down from the datastore, so we could match `CDTDatastore`'s interface to query e.g. 

```java
Datastore ds = manager.getDatastore("example");
ds.ensureIndexed(....);
ds.find(....);
```

So just to clear up, this doesn't propose you have a bunch of classes which you'd expose depending on the capabilities, you'd get a single interface `Datastore` in this case, which would have all those methods, so a user would only have a single instance which could be `LargeAttachmentDatastore` (being the last set of function we added), but they wouldn't know they'd just have the new `LargeAttachmentDatastore` methods appear.

`Façade Pattern` will give us what we want in the single unifying interface to make it appear more like `CDTDatastore` (in fact it will work more like `CDTDatastore` than it currently does). And in long term I think it if we ever add Geo capabilities façade will give us the best option. So I think that is the option to go for.

tomblench: But how does the manager know which concrete type to return when `getDatastore()` is invoked? It can't tell in advance whether you want to call Query/Index methods, Attachment, or something else. You have:
```
QueryableDatastore ds = new QueryableDatastore(dbDirectory, dbName, provider);
```

tomblench: I'm also troubled by this:

```LargeAttachmentDatastore (being the last set of function we added), but they wouldn't know they'd just have the new LargeAttachmentDatastore methods appear```

Because the name of each class because an increasing blatant lie (it's actually a Large Attachment _and_ Queryable Datastore)

rhyshort: Because it would *only* use the one at the bottom of the tree. The others would be abstract, like `DatastoreImpl` but anyway, like I said, the façade is the way to go, i'm looking at prototyping it right now.

rhyshort: I also admit the naming wasn't great. It was all I had at the time.

tomblench: @rhyshort cool glad we're in agreement. I do also concede your point that there will be some ugly coupling between (for example) the basic CRUD datastore and the Index/Query 'manager', but that's no worse than what we have at the moment except we can at least hide the complexity behind a façade.

rhyshort:
See the https://github.com/cloudant/sync-android/tree/facade-experiment for what I currently have for changing to using a facade.

rhyshort:
Closing until we can decide the best way forward for API compatibility

PR #309, opened 2016-07-12
Added gradle tasks for counting methods

### LINKED ISSUE #271: Generate method counts
We get asked about the method count of the library which is important with the 65k method limit before multidex.
A lot of this goes away in later versions of Android with the newer tool chain, however it would be helpful to have the method count.

ricellis:
*What*
Added gradle tasks for counting methods

*Why*
Older versions of Android have a 65k method limit and it is good to know how much the library might take up.

*How*
Added tasks to copy all dependencies and produce a single dex file.
Extracted the `method id` statistic from the dex log.
Added task to delete the copied dependencies and generated file.

*Test*
No new tests, the task is designed to be run manually when a new method count is needed and I have tested it locally.
Runs with e.g.:
`./gradlew methodCount -Ddx.exec=~/Library/Android/sdk/build-tools/24.0.0/dx`
with the `dx.exec` property pointing to the Android build tools dx executable.

Fixes #271

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,8 +27,6 @@ allprojects {
     }
 }
 
-
-
 dependencies {
     compile project(':cloudant-sync-datastore-core')
     compile project(':cloudant-sync-datastore-javase')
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -224,6 +224,36 @@ subprojects {
...
+    task dx(type: Exec, dependsOn: copyLib) {
+        workingDir new File(buildDir, "dex")
+        commandLine 'bash', '-e', '-c', """
+            ls -1 . > dexInputList.txt
+            dx.exec --dex --statistics --num-threads=4 --core-library --output=./file.dex --input-list=dexInputList.txt >dex.log 2>&1
+            grep 'method id:.*' dex.log
+        """.replace("dx.exec",(System.getProperty("dx.exec") == null) ? "dx" : System.getProperty("dx.exec"))
+    }
+
+    task deleteDex(type: Delete, dependsOn: dx) {
```
tomblench: I think you drop the `deleteDex` class and rely on the `clean` targets doing the right thing.

build.gradle
```diff
@@ -224,6 +224,36 @@ subprojects {
             }
         }
     }
+
+
+    task copyDependencies(type: Copy) {
+        from configurations.runtime
+        into new File(buildDir, "dex")
```
tomblench: `File(buildDir, "dex-all")` should be in some shared config and re-used by these tasks. I suggest something like `dex-all` just to distinguish from other invocations of `dex`

build.gradle
```diff
@@ -224,6 +224,36 @@ subprojects {
...
+    }
+
+    task copyLib(type: Copy, dependsOn: copyDependencies) {
+        from new File(buildDir, "libs")
```
tomblench: I wonder if there is a config variable for the `"libs"` dir, but we already use this string elsewhere so it's already established practice.

build.gradle
```diff
@@ -224,6 +224,36 @@ subprojects {
...
+        into new File(buildDir, "dex")
+    }
+
+    task dx(type: Exec, dependsOn: copyLib) {
```
tomblench: maybe call this task `dexAll`?

build.gradle
```diff
@@ -224,6 +224,36 @@ subprojects {
...
+        """.replace("dx.exec",(System.getProperty("dx.exec") == null) ? "dx" : System.getProperty("dx.exec"))
+    }
+
+    task deleteDex(type: Delete, dependsOn: dx) {
+        delete new File(buildDir, "dex")
+    }
+
+    task methodCount(dependsOn: [build, copyDependencies, copyLib, dx, deleteDex]) {
+        // Run this task to get the Android method counts.
+        // Relies on the dx.exec system property being set to point to the dx executable
```
tomblench: or on `dx` being in your `PATH`

ricellis:
@tomblench addressed comments in 0bf7f29:

Added constant `dexAllDir` for `dex-all` directory.
Used `jar.destinationDir` instead of buildDir + “libs”.
Renamed `dx` task to `dexAll`.
Added clarification that `dx` can be on the path for gradle.
Removed `deleteDex` task, `clean` will remove it anyway.

build.gradle
```diff
@@ -224,6 +224,33 @@ subprojects {
...
+    task dexAll(type: Exec, dependsOn: [copyDependencies, copyLib]) {
+        workingDir dexAllDir
+        commandLine 'bash', '-e', '-c', """
+            ls -1 . > dexInputList.txt
+            dx.exec --dex --statistics --num-threads=4 --core-library --output=./file.dex --input-list=dexInputList.txt >dex.log 2>&1
+            grep 'method id:.*' dex.log
+        """.replace("dx.exec",(System.getProperty("dx.exec") == null) ? "dx" : System.getProperty("dx.exec"))
+    }
+
+    task methodCount(dependsOn: [build, copyDependencies, copyLib, dexAll]) {
```
tomblench: `copyDependencies, copyLib` dependencies are implied don't strictly need to be there.

ricellis: @tomblench fixed the dependencies and made these other improvements in 42dbe4b:
* Created `AndroidExec` task type to add android tools to the path reading
`sdk.home` or `ANDROID_HOME`.
* Removed `dx.exec` property and used `AndroidExec` type instead.
* Fixed dependencies including using `assemble` instead of `build` to avoid
running tests if just doing `methodcount`.
* Added input/output directories to `dexAll` task to avoid running twice
unnecessarily.

Unfortunately sharing the `AndroidExec` type with the AndroidTest/app project is proving problematic because they don't share a root build. I think I'd like to merge what I have here and consider the sharing later (possibly with the log collection as that is going to need android commands as well).

tomblench:
+1

Issue #276, opened 2016-06-21
Small attachments unnecessarily base64 encoded


ricellis:
In com.cloudant.sync.datastore.RevisionHistoryHelper#addAttachments the attachments are checked 1 at a time and inline attachments are always base64 encoded. However, in a mixture of large and small attachments where the attachments will be sent as part of a MIME multipart/related message it does not make sense to do this base64 encoding.

We should perform an initial check of the attachments to identify if any attachment revisions that need to be sent qualify for multipart and if so all attachments should be done as multipart.

tomblench:
Done in #311

PR #311, opened 2016-07-15
Small attachments unnecessarily base64 encoded

tomblench:
_What_: If we decide to send attachments via multipart/related, then all attachments should be sent as additional body parts even if they were previous determined as being too small to send as body parts. Put another way, either all attachments are sent as inline base64 in the json body, or they are sent as body parts.

_Why_: The previous behaviour was slightly inefficient. Once we switched to the multipart/related method, it's wasteful to base64 encode some attachments.

_Testing_: Additional tests pass, added new test `pushBigAndSmallAttachmentsTest` and verified correct behaviour via wireshark.

GitHub issue: #276

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/RevisionHistoryHelper.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/RevisionHistoryHelper.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013 Cloudant, Inc. All rights reserved.
+ * Copyright (c) 2013, 2016 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/RevisionHistoryHelper.java
```diff
@@ -164,14 +164,52 @@ private static boolean checkRevisionStart(DocumentRevs.Revisions revisions) {
...
+     *
+     * @return a boolean flag determining whether all attachments should be sent inline or not.
+     */
+    public static boolean shouldInline(List<? extends Attachment> attachments,
+                                PushAttachmentsInline inlinePreference,
+                                int minRevPos) {
+        boolean shouldInline = true;
+        // first figure out if any attachments don't want to be inlined
+        for (Attachment att : attachments) {
+            SavedAttachment savedAtt = (SavedAttachment)att;
```
ricellis: missing space between cast and field.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PushAttachmentsInline.java
```diff
@@ -20,10 +20,10 @@
  * </p>
  *
  * <ul>
- *     <li>False: Always push attachments separately as an HTTP binary PUT</li>
+ *     <li>False: Always push attachments separately from JSON body, via multipart/related</li>
  *     <li>Small: Push small attachments inline, and large attachments separately.
```
ricellis: Do you think we need to add a clarification here, since the behaviour is slightly different now?
I mean if I specify `Small` and have 1 large attachment my small attachments will also be sent separately now right?

ricellis:
As a general comment the copyright dates should probably be updated.

tomblench:
@ricellis thanks for the feedback, these should all be addressed in a0b3b484d873eabf41ae1e4623da9702dc4d864a

ricellis:
+1

PR #312, opened 2016-07-15
Collect logcat from device runs

ricellis:
*What*

Collect logcat from device runs.

*Why*

Additional debug info is useful if tests fail on emulator or device.

*How*

Added `clearDeviceLog` and `pullDeviceLog` tasks to run either
side of tasks of type `AndroidTestTask`. It is annoying to reference that internal type, but I couldn't find a better way to get the dependency on the correct task.

Enabled the new tasks and `uploadFixtures` to use `AndroidExec` task type by:

* Moved `AndroidExec` task into a shareable build script.
* Made `AndroidExec` search paths for matching Android tools to execute as
environment variables required calling bash which wasn’t platform
independent (env only applies to the executed process not the search
for that executable).
* Imported shared script into build.gradle and AndroidTest/build.gradle.
* Refactored commandLine properties to suit AndroidExec without needing bash

*Test*

Running the `connectedCheck` produced output in `AndroidTest/app/build/reports/logcat.log`

### DIFF
--- a/AndroidExec.gradle
+++ b/AndroidExec.gradle
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2016 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+// Define AndroidExec as an external property for includers to be able to use it
+ext.AndroidExec = AndroidExec
+
+// A gradle Exec task that searches the Android tools locations for matching executables.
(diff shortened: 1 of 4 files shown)


AndroidExec.gradle
```diff
@@ -0,0 +1,72 @@
...
+                    @Override
+                    boolean accept(File dir, String name) {
+                        return name.equals(getExecutable());
+                    }
+                })
+                if (matchingExec.length == 1) {
+                    setExecutable(new File(toolsPath, matchingExec[0]))
+                    break pathLoop
+                } else {
+                    continue
```
tomblench: else/continue un-necessary because nothing happens after this point

tomblench:
+1 although `AndroidExec.gradle` isn't ideal it looks like a useful rationalisation.

AndroidExec.gradle
ricellis: Removed the else and unnecessary loop label in bdb7192

ricellis:
I thought `AndroidExec` was preferable to either the code duplication, building a separate jar to add to the classpath, or a custom gradle plugin. The only other way I can see to avoid the `AndroidExec` file is if the AndroidTest project became a sub-project of the sync-android build, but I'm not sure what impact having sync-android as a rootProject of AndroidTest might have on other parts of the build, so I thought "here be dragons" and left that approach alone. Perhaps it is something we could revisit if AndroidExec is in some way cumbersome or inconvenient.

PR #319, opened 2016-07-21
QueryResultIterator NoSuchElementException

### LINKED ISSUE #318: QueryResultIterator doesn't account for Post SQL matching
`QueryResultIterator` doesn't count for post sql query matching, resulting in a `NoSuchElementException` being thrown when we attempt to provide a document to `next()` call via the list iterator. This occurs when a batch is emptied by the Post SQL matcher.

rhyshort:
## What
Correctly calculate if there are more documents to be iterated over.

## How.

Calculate if there a remaing documents to iterator over correctly, by loading
the next batch of documents before returning the last document of the previous
batch, and only if the limit on the number of results hasn't been reached.

## Tests

New tests added to `QueryWithoutCoveringIndexesTest` which populate the datastore with documents which match using the part of the query which can be performed in SQL, but doesn't match the remaining parts of the query using post hoc matcher. 

## Issues

Fixes #318

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryResult.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryResult.java
@@ -122,10 +122,21 @@ public class QueryResult implements Iterable<DocumentRevision> {
             if (!this.hasNext()) {
                 throw new NoSuchElementException();
             }
-            if (!documentBlock.hasNext()) {
+
+            // it is always safe to call documentBlock.next() without calling hasNext because
+            // the this.hasNext() will return false if there are no more documents to read this
+            // is because we load the next block of documents (if there are any), before
+            // returning the last document from the last batch, ensuring that documentBlock.next()
+            // will always have a document to return.
+            DocumentRevision doc =  documentBlock.next();
+
+            // Only load the next batch if the limit hasn't been reached, if it has
+            // it will cause an off by one error, eg instead of 60, you'd get 61 results.
+            if (!documentBlock.hasNext() && !limitReached) {
                 documentBlock = populateDocumentBlock();
             }
-            return documentBlock.next();
+
+            return doc;
         }
 
         @Override
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/QueryWithoutCoveringIndexesTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/QueryWithoutCoveringIndexesTest.java
@@ -21,6 +21,8 @@ import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
 import static org.junit.runners.Parameterized.Parameters;
 
+import com.cloudant.sync.datastore.DocumentBodyFactory;
+import com.cloudant.sync.datastore.DocumentRevision;
 import com.cloudant.sync.util.SQLDatabaseTestUtils;
 import com.cloudant.sync.util.TestUtils;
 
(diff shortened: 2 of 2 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryResult.java
```diff
@@ -122,10 +122,16 @@ public DocumentRevision next() {
             if (!this.hasNext()) {
                 throw new NoSuchElementException();
             }
-            if (!documentBlock.hasNext()) {
+
+            DocumentRevision doc =  documentBlock.next();
```
ricellis: I think this could do with a comment explaining why it is safe to call this without an explicit proceeding `documentBlock.hasNext()` just so it is easier for the next person who wanders into this block.

`this.hasNext()` calls `documentBlock.hasNext()`, but it ORs that result with another check so it is at first glance possible to reach this line without a document being available in the `documentBlock`. I think it would be helpful to clarify that in the case of the initial or final block being empty the range check in `hasNext` protects this call.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryResult.java
```diff
@@ -122,10 +122,16 @@ public DocumentRevision next() {
...
+
+            // Only load the next batch if the limit hasn't been reached, if it has
+            // it will cause an off by on error, eg instead of 60, you'd get 61 results.
```
ricellis: off by on**e**

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryResult.java
rhyshort: thoughts @ricellis 
```java
            // it is always safe to call documentBlock.next() without calling hasNext because
            // the this.hasNext() will return false if there are no more documents to read this
            // is because we load the next block of documents (if there are any), before
            // returning the last document from the last batch, ensuring that documentBlock.next()
            // will always have a document to return.
            DocumentRevision doc =  documentBlock.next();

```

ricellis: I wonder if there is a record for comment per line :)
I think that is helpful though.

rhyshort: Cool, Guinness world records will know if there is one.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryResult.java
rhyshort: fixed in a5f0033

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryResult.java
rhyshort: added in a5f0033

ricellis:
+1 but I'd like another reviewer to check this too as the iteration has subtleties and I think another thought pattern (further removed from the original debugging) would be good.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/QueryWithoutCoveringIndexesTest.java
```diff
@@ -109,6 +111,72 @@ public void canQueryWithoutIndexSingleClause() throws Exception {
...
+            Map<String, Object> bodyMap = new HashMap<String, Object>();
+            bodyMap.put("name", "rhys");
+            bodyMap.put("age", 12);
+            bodyMap.put("pet", "cat");
+            bodyMap.put("town", "cardiff");
+            rev.setBody(DocumentBodyFactory.create(bodyMap));
+            ds.createDocumentFromRevision(rev);
+        }
+
+
```
brynh: I'd prefer only one blank line here.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/QueryWithoutCoveringIndexesTest.java
```diff
@@ -109,6 +111,72 @@ public void canQueryWithoutIndexSingleClause() throws Exception {
...
+            } else {
+                bodyMap.put("town", "bristol"); //+2
+            }
+            rev.setBody(DocumentBodyFactory.create(bodyMap));
+            ds.createDocumentFromRevision(rev);
+        }
+
+
+
+
```
brynh: I'd prefer only one blank line here.

brynh:
+1 once extra blank lines are removed.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/QueryWithoutCoveringIndexesTest.java
rhyshort: done in 494aaf3

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/QueryWithoutCoveringIndexesTest.java
rhyshort: done in 494aaf3

PR #321, opened 2016-07-22
Throw Exception instead of returning null

### LINKED ISSUE #320: SQLDatabaseFactory can return null
The `SQLDatabaseFactory` can return `null` when it fails to open a database. The `IndexManager` and `SQLDatabaseQueue` currently assume as per the documentation that the class **will** return a SQLDatabase instance, or throw an exception. However this is not the case, it will only throw an exception if it cannot create the file, and then returns null if it fails to create open the database for use.

rhyshort:
## What

Do not return null when the database can not be opened

## How

Throw SQLDatabaseException instead of returning null when the database
cannot be opened.

## Tests

No additional tests, all current tests pass.

## Issues

Fixes #320


brynh:
I note that `SQLDatabaseFactory.openSqlDatabase()` can also return `null` in 2 places and that this behaviour isn't documented. I think for consistency, this should also be changed to throw a `SQLException` in these cases.

rhyshort:
@brynh Done in d0f71ba

brynh:
+1

PR #322, opened 2016-07-27
Use SQLDatabaseQueue instead of an executor service directly

### LINKED ISSUE #286: Refactor IndexManager to use SQLDatabaseQueue instead of SQLDatabase

rhyshort:
## What

Use `SQLDatabaseQueue` instead of an `ExecutorService` and `SQLDatabse` directly.

## How

Swap out the `SQLDatabase` and `ExecutorService` with `SQLDatabaseQueue`

## Testing

All tests pass, however some have been reworked slightly so they run on the queue rather than on the main test thread.

## Issues

Resolves #286

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
@@ -43,9 +43,12 @@ import com.cloudant.sync.datastore.migrations.SchemaOnlyMigration;
 import com.cloudant.sync.sqlite.Cursor;
 import com.cloudant.sync.sqlite.SQLDatabase;
 import com.cloudant.sync.sqlite.SQLDatabaseFactory;
+import com.cloudant.sync.sqlite.SQLDatabaseQueue;
+import com.cloudant.sync.sqlite.SQLQueueCallable;
 import com.cloudant.sync.util.DatabaseUtils;
 
 import java.io.File;
+import java.io.IOException;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -84,9 +87,9 @@ public class IndexManager {
     private static final Logger logger = Logger.getLogger(IndexManager.class.getName());
 
     private final Datastore datastore;
-    private final SQLDatabase database;
     private final Pattern validFieldName;
-    private final ExecutorService queue;
+
+    private final SQLDatabaseQueue dbQueue;
 
     private boolean textSearchEnabled;
 
@@ -97,54 +100,29 @@ public class IndexManager {
     public IndexManager(Datastore datastore) {
         this.datastore = datastore;
         validFieldName = Pattern.compile(INDEX_FIELD_NAME_PATTERN);
-        queue = Executors.newSingleThreadExecutor();
 
         final String filename = ((DatastoreImpl)datastore).extensionDataFolder(EXTENSION_NAME) + File.separator
                                                                               + "indexes.sqlite";
         final KeyProvider keyProvider = ((DatastoreImpl)datastore).getKeyProvider();
 
-        SQLDatabase sqlDatabase = null;
+        SQLDatabaseQueue queue = null;
+
         try {
-            sqlDatabase = queue.submit(new Callable<SQLDatabase>() {
-                @Override
-                public SQLDatabase call() throws Exception {
-                    SQLDatabase db = SQLDatabaseFactory.openSqlDatabase(filename, keyProvider);
-
-                    if (db != null) {
-                        SQLDatabaseFactory.updateSchema(db,
-                                new SchemaOnlyMigration(QueryConstants.getSchemaVersion1()), 1);
-                        SQLDatabaseFactory.updateSchema(db,
-                                new SchemaOnlyMigration(QueryConstants.getSchemaVersion2()), 2);
-                    }
-
-                    return db;
-                }
-            }).get();
-        } catch (InterruptedException e) {
-            logger.log(Level.SEVERE, "Problem opening or creating database.", e);
-        } catch (ExecutionException e) {
-            logger.log(Level.SEVERE, "Problem opening or creating database.", e);
+            queue = new SQLDatabaseQueue(filename, keyProvider);
+            queue.updateSchema(new SchemaOnlyMigration(QueryConstants.getSchemaVersion1()), 1);
+            queue.updateSchema(new SchemaOnlyMigration(QueryConstants.getSchemaVersion2()), 2);
+            textSearchEnabled = ftsAvailable(queue);
+        } catch (Exception e) {
+            logger.log(Level.SEVERE, "Failed to open database", e);
+            queue = null;
         }
-        database = sqlDatabase;
-        textSearchEnabled = ftsAvailable(queue, database);
+
+        dbQueue = queue;
+
     }
 
     public void close() {
-        try {
-            queue.submit(new Runnable() {
-                @Override
-                public void run() {
-                    database.close();
-
-                }
-            }).get();
-        } catch (InterruptedException e) {
-            logger.log(Level.SEVERE,"Failed to close db",e);
-            throw new RuntimeException(e);
-        } catch (ExecutionException e) {
-            logger.log(Level.SEVERE, "Failed to close db", e);
-        }
-        queue.shutdown();
+        dbQueue.shutdown();
     }
 
     /**
@@ -161,9 +139,9 @@ public class IndexManager {
      */
     public Map<String, Object> listIndexes() {
         try {
-            return queue.submit(new Callable<Map<String, Object>>() {
+            return dbQueue.submit(new SQLQueueCallable<Map<String, Object> >() {
                 @Override
-                public Map<String, Object> call() throws Exception {
+                public Map<String, Object> call(SQLDatabase database) throws Exception {
                      return IndexManager.listIndexesInDatabase(database);
                 }
             }).get();
@@ -241,9 +219,8 @@ public class IndexManager {
      */
     public String ensureIndexed(List<Object> fieldNames, String indexName) {
         return IndexCreator.ensureIndexed(Index.getInstance(fieldNames, indexName),
-                database,
                 datastore,
-                queue);
+                dbQueue);
     }
 
     /**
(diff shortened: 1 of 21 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
```diff
@@ -139,9 +135,9 @@ private boolean updateIndex(final String indexName,
             return false;
         }
 
-        Future<Boolean> result = queue.submit( new Callable<Boolean>() {
+        Future<Boolean> result = queue.submitTransaction(new SQLQueueCallable<Boolean>() {
```
ricellis: Now that this is using a `SQLQueueCallable` and `submitTransaction` it is strange that it also calls begin transaction. I think we should either just use `submit` and let it handle the transaction or preferably remove the extra transaction calls and let `submitTransaction` control the transaction (I think we just need to throw where currently we set `transactionSuccess=false`). At the moment we have two sets of calls trying to do the same thing.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -385,23 +361,27 @@ protected Datastore getDatastore() {
     }
 
     protected ExecutorService getQueue() {
-        return queue;
+        return null;
```
ricellis: I think we'd be better off deleting this rather than returning `null` and equally for `getDatabase`. I think the chances of anyone calling/using them in a subclass are remote.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
```diff
@@ -385,23 +361,27 @@ protected Datastore getDatastore() {
...
+    }
+
+    protected SQLDatabaseQueue getDatabaseQueue() {
```
ricellis: I don't like adding the method just for the tests, that's probably how we ended up with `getQueue()` and `getDatabase()` here in the first place.
Admittedly I'm not a huge fan of reflectively obtaining it in the base class of the tests, but I think that's preferable to adding a method.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/AbstractIndexTestBase.java
```diff
@@ -43,18 +43,16 @@ public void setUp() throws Exception {
         assertThat(ds, is(notNullValue()));
         im = new IndexManager(ds);
         assertThat(im, is(notNullValue()));
-        db = TestUtils.getDatabaseConnectionToExistingDb(im.getDatabase());
-        assertThat(db, is(notNullValue()));
-        assertThat(im.getQueue(), is(notNullValue()));
+        assertThat(im.getDatabaseQueue(), is(notNullValue()));
```
ricellis: The `SqlDatabase db` field can be removed now that the value write for it has been deleted.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
```diff
@@ -563,36 +633,42 @@ public void indexSingleArrayFieldWhenIndexingArraysInSubDoc() throws Exception {
...
+                        assertThat(cursor.getString(1), is(saved.getRevision()));
+                        assertThat(cursor.columnName(2), is("name"));
+                        assertThat(cursor.getString(2), is("mike"));
+                        assertThat(cursor.columnName(3), is("pet.species"));
+                        assertThat(cursor.getString(3), is(notNullValue()));
+                        petList.add(cursor.getString(3));
+                    }
+                    assertThat(petList, containsInAnyOrder("cat", "dog"));
+                }catch (SQLException e) {
+                    Assert.fail(String.format("SQLException occurred executing %s: %s", sql, e));
```
ricellis: Any reason you chose to keep this catch block, but not the others (i.e. in previous tests)? The `SQLException` being thrown would end up making the test error via the `ExecutionException` from the call to `get()` anyway, but I guess it is vaguely nicer to have the fail rather than the error.
Anyway I think we should keep it the same in the tests either catch and fail, or just allow to bubble out and error.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
```diff
@@ -635,37 +715,43 @@ public void rejectsDocsWithMultipleArrays() throws Exception {
...
+                        assertThat(cursor.getString(2), is("mike"));
+                        assertThat(cursor.columnName(3), is("pet"));
+                        assertThat(cursor.getString(3), is(notNullValue()));
+                        petList.add(cursor.getString(3));
+                        assertThat(cursor.columnName(4), is("pet2"));
+                        assertThat(cursor.getString(4), is(nullValue()));
+                    }
+                    assertThat(petList, containsInAnyOrder("cat", "dog", "parrot"));
+                }catch (SQLException e) {
+                    Assert.fail(String.format("SQLException occurred executing %s: %s", sql, e));
```
ricellis: Another one

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
rhyshort: Ah no, it should have been removed, blinded by  the amount of code i needed to change

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
rhyshort: removed in 329ba24

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java
rhyshort: changed it to using `submit`  in 329ba24

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManager.java
rhyshort: done in 36056b1

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/AbstractIndexTestBase.java
rhyshort: done in baf177a

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
rhyshort: done in 15dfb37

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
rhyshort: done in 15dfb37

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/AbstractIndexTestBase.java
```diff
@@ -43,16 +43,16 @@ public void setUp() throws Exception {
         assertThat(ds, is(notNullValue()));
         im = new IndexManager(ds);
         assertThat(im, is(notNullValue()));
-        assertThat(im.getDatabaseQueue(), is(notNullValue()));
+        assertThat(TestUtils.getDBQueue(im), is(notNullValue()));
```
ricellis: It may be preferable to declare
`protected SQLDatabaseQueue queue;`
and set the value in the `setUp` using:
`queue = TestUtils.getDBQueue(im);`
That way the reflection only needs to be performed once per test.

emlaver: I agree with @ricellis, no need to have `TestUtils.getDBQueue(im)` run multiple times each test.

rhyshort: done in 3921a46

emlaver:
LGTM +1

ricellis:
@rhyshort the Android test failures appear to be due to `IndexManager.close()` not being called in the various tests in the `EndToEndEncryptionTest` class.

rhyshort:
@ricellis commit 323a707 should fix it.

ricellis:
@rhyshort there were other tests using IndexManager, should also be closed in finally block, done in 6c8af67.

rhyshort:
cool, and the builds green again :)

ricellis:
+1 - I saw some more db leaks in the logcat, but I don't think they are related to the IndexManager so we'll get them under another issue.

PR #327, opened 2016-08-10
326 pick winner

### LINKED ISSUE #326: Pick winner fails with two roots with identical rev IDs
Consider the sequence:
`1: 1-x 2: 1-x`
Initially 1: 1-x is the winner.
In the case where a deletion of `1:1-x` is force inserted by a replication we end up with a new deleted revision `3: 2-x` and `1: 1-x` is marked as not current because it is superseded by `3: 2-x` on its branch.
Pick winner then attempts to determine the new winner and identifies the non-deleted `2: 1-x` as the winning non-deleted leaf. It then asks (using the rev id only) for the sequence number for that revision and `1` is returned because `1: 1-x` shares the revision ID. The old winner was `1` and the new winner is thought to be `1` (though it is in fact `2`) so no new winner is selected.

ricellis:
*What*

Corrected selection of winner when multiple root nodes have identical revisions - fixes #326 

*How*

Added sequence to information returned from `GET_NON_DELETED_LEAFS` and  `GET_ALL_LEAFS` SQL queries and stored query results in a `SortedMap`.

*Test*
Added test `forceInsert_pickWinnerOfConflicts_TwoOfSameRoot_OneDeleted`.

### DIFF
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreImplForceInsertTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreImplForceInsertTest.java
@@ -18,7 +18,9 @@ import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.hasItems;
 import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
 
+import com.cloudant.android.ContentValues;
 import com.cloudant.sync.sqlite.SQLDatabase;
+import com.cloudant.sync.sqlite.SQLQueueCallable;
 import com.cloudant.sync.util.CouchUtils;
 import com.cloudant.sync.util.JSONUtils;
 import com.cloudant.sync.util.TestUtils;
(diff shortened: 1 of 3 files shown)


rhyshort:
+1

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreImplForceInsertTest.java
```diff
@@ -505,6 +505,30 @@ public void forceInsert_pickWinnerOfConflicts_Simple_2y_last() throws Exception
...
+        // Now force insert the deleted
+        datastore.forceInsert(rev2, "1-x", "2-x");
+
+        // Get the document
+        DocumentRevision doc = datastore.getDocument(OBJECT_ID);
+
+        // We favour non-deleted nodes so the winner should be a 1-x
+        Assert.assertEquals("The document winner should be the expected revision", "1-x", doc
+                .getRevision());
+        Assert.assertEquals("The document winner should have the expected body", bodyTwo.asMap(),
```
tomblench: There is no `order by` clause in the SQL used for `pickWinnerOfConflicts` so by my reading of the code it is arbitrary which sequence is last put into the map for a given rev id. Possibly the underlying SQLite structure is sorted in the order in which rows were inserted but we can't rely on that.

Not that it really matters, if we have got into the situation where two documents have exactly the same revision ID then we have no way of distinguishing which one is "right" anyway. You could be consistent and pick the lowest/highest sequence number on the basis that first/last write wins.

rhyshort: SQLite's Query planner appears to guarantee ordering of results without the ordered by statement see: http://www.sqlite.org/queryplanner.html

tomblench: As long as we don't have 9223372036854775807 entries in the table :) https://www.sqlite.org/autoinc.html

ricellis: As per our discussion the ordering is not guaranteed by our query, but in practice appears to always work. I have done the following to make this more robust:
* changed the query to order by sequence so even if there are duplicates the behaviour of where things get grafted will be guaranteed in future: ac570b0
* 70483d1 added a lot more explanation to the test and changed the duplicate insert to not use `forceInsert` so that when `forceInsert` is updated to protect against duplicates the test won't fail. If we update the schema to prevent duplicates then we probably just have to get rid of the test later.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreImplForceInsertTest.java
```diff
@@ -504,6 +506,63 @@ public void forceInsert_pickWinnerOfConflicts_Simple_2y_last() throws Exception
...
+    /**
+     * <p>
+     * This test is for https://github.com/cloudant/sync-android/issues/326. The test nefariously
+     * creates a tree with two roots with identical revision IDs (replicating the type of tree
+     * created by a bug with parallel replications and force insert). The two revisions have
+     * different bodies to help us assert easily which is which (note that revision IDs were not
+     * calculated based on body content in this library at the time this test was created).
+     * </p>
+     * <p>
+     * One those two revisions are in place a deletion of the revision ID is force inserted as would
```
tomblench: Once

tomblench:
+1 apart from comment typo

Issue #325, opened 2016-08-08
Add logging for sqlite version


ricellis:
It could be helpful for debugging to know what version of SQLite is being used. It varies with Android version, but may also be fragmented by vendor specific builds using different variants.
The `sqlite3` command is available in emulators, but not typically for devices when using `adb`. One option is to use something like:
```java
Cursor cursor = SQLiteDatabase.openOrCreateDatabase(":memory:", null).rawQuery("SELECT sqlite_version() AS sqlite_version", null);
String sqliteVersion = "";
while(cursor.moveToNext()){
   sqliteVersion += cursor.getString(0);
}
```

Issue #292, opened 2016-06-22
Perform readability validation on JavaSE also


ricellis:
The validation that a database has been correctly opened is currently only performed on Android. We should do this on JavaSE as well.
See com.cloudant.sync.sqlite.SQLDatabaseFactory#validateOpenedDatabase

Currently doing so causes failures, particularly in IndexUpdater tests, which I anticipate may be resolved by #286.
I think it is safe to do the validation on the `SQLDatabase` and not the `SQLDatabaseQueue` because it happens before it is returned for other access.

tomblench:
NB: `validateOpenedDatabase` is only called for the SQLCipher case of `openSqlDatabase`

ricellis:
I believe the validation was originally added to verify the supplied key was appropriate to read the database in the encrypted case, but I don't think there is anything to stop us running it in all cases if we feel that it is valuable to do so.

tomblench:
Note that `validateOpenedDatabase` appears not to exist any more, probably was lost some time around #501

PR #328, opened 2016-08-10
Increase diagnostics

rhyshort:
## What

Increase the ability to diagnose problems when the changes feed throws an `IllegalStateException`.

## How

- Name threads created by thread pools for SQLDatabaseQueue and for pull replications.
- Name the ReplicatorThread
- Provide the SQLite version and include it in the IllegalState thrown by the changes feed.

## Tests

No new tests, manually checked that the threads get renamed using the debugger in idea.

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
@@ -510,7 +510,7 @@ public class DatastoreImpl implements Datastore {
            logger.log(Level.SEVERE, "Failed to get changes",e);
             if(e.getCause()!= null){
                 if(e.getCause() instanceof IllegalStateException) {
-                    throw (IllegalStateException) e.getCause();
+                    throw new IllegalStateException(e);
                 }
             }
         }
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
@@ -61,6 +61,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.SortedMap;
 import java.util.TreeMap;
@@ -492,8 +493,12 @@ public class DatastoreImpl implements Datastore {
                         }
                         List<DocumentRevision> results = getDocumentsWithInternalIdsInQueue(db, ids);
                         if(results.size() != ids.size()) {
-                            throw new IllegalStateException(String.format("The number of documents %d does not match number of ids %d, " +
-                                    "something must be wrong here.", results.size(), ids.size()));
+                            throw new IllegalStateException(String.format(Locale.ENGLISH,
+                                    "The number of documents does not match number of ids, " +
+                                    "something must be wrong here. Number of IDs: %s, number of documents: %s",
+                                    ids.size(),
+                                    results.size()
+                                    ));
                         }
 
                         return new Changes(lastSequence, results);
@@ -508,11 +513,9 @@ public class DatastoreImpl implements Datastore {
             logger.log(Level.SEVERE, "Failed to get changes",e);
         } catch (ExecutionException e) {
            logger.log(Level.SEVERE, "Failed to get changes",e);
-            if(e.getCause()!= null){
                 if(e.getCause() instanceof IllegalStateException) {
-                    throw new IllegalStateException(e);
+                    throw new IllegalStateException("Failed to get changes, SQLite version: "+queue.getSQLiteVersion(),e);
                 }
-            }
         }
 
         return null;
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
@@ -28,6 +28,7 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
@@ -123,7 +124,12 @@ class GetRevisionTaskThreaded implements Iterable<DocumentRevsList> {
             @Override
             public DocumentRevsList executeRequest(BulkGetRequest request) {
                 // since this is part of a thread pool, we'll rename each thread as it takes a task.
-                Thread.currentThread().setName("GetRevisionThread: "+GetRevisionTaskThreaded.this.sourceDb.getIdentifier());
+                try {
+                    Thread.currentThread().setName("GetRevisionThread: " + GetRevisionTaskThreaded.this.sourceDb.getIdentifier());
+
+                } catch (SecurityException e){
+                    logger.log(Level.WARNING, "Could not rename pull strategy pool thread", e);
+                }
                 return new DocumentRevsList(GetRevisionTaskThreaded.this.sourceDb.getRevisions
                         (request.id, request.revs, request.atts_since,
                                 GetRevisionTaskThreaded.this.pullAttachmentsInline));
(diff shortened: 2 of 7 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -483,8 +484,12 @@ public Changes call(SQLDatabase db) throws Exception {
                         }
                         List<DocumentRevision> results = getDocumentsWithInternalIdsInQueue(db, ids);
                         if(results.size() != ids.size()) {
-                            throw new IllegalStateException("The number of document does not match number of ids, " +
-                                    "something must be wrong here.");
+                            throw new IllegalStateException(String.format(Locale.ENGLISH,
+                                    "The number of document does not match number of ids, " +
```
brynh: `document` -> `documents`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -501,7 +506,7 @@ public Changes call(SQLDatabase db) throws Exception {
            logger.log(Level.SEVERE, "Failed to get changes",e);
             if(e.getCause()!= null){
```
brynh: Although not introduced by this PR, this null check is redundant.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
```diff
@@ -121,6 +122,8 @@ public GetRevisionTaskThreaded(CouchDB sourceDb,
                 DocumentRevsList>(executorService, this.requests, threads + 1) {
             @Override
             public DocumentRevsList executeRequest(BulkGetRequest request) {
+                // since this is part of a thread pool, we'll rename each thread as it takes a task.
+                Thread.currentThread().setName("GetRevisionThread: "+GetRevisionTaskThreaded.this.sourceDb.getIdentifier());
```
brynh: We should maybe catch the `SecurityException` that could be thrown by the call to `setName()` and not set a custom thread name in that case. I don't think `setName()` could fail on Android, but it probably could on desktop Java with a security manager.

ricellis: Has the formatter run on this? I feel like there ought to be some spaces around the `+` in here.

rhyshort: Erm pass. I don't know how intellij runs the formatter (I assumed it was automatic), should I be doing it manually?

rhyshort: Catch security exception in 34fb8be

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
rhyshort: removed in 34fb8be

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
rhyshort: fixed in 34fb8be

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/GetRevisionTaskThreaded.java
ricellis: Formatter is not automatic by default, you can set it to run on all lines or modified lines.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
```diff
@@ -189,4 +195,30 @@ public boolean isShutdown() {
             throw new RejectedExecutionException("Database is closed");
         }
     }
+
+    /**
+     * Returns the SQLite Version.
+     * @return The SQLite version or "Unknown" if the version could not be determined.
+     */
+    public String getSQLiteVersion() {
```
ricellis: I'd prefer to see this run once when opening the database (avoiding the need to queue it) and then storing the version as a constant for any use later. We could include it in the thread name for instance.

brynh:
+1

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
rhyshort: We could do that, but it would change the start up profile of the datastore, since currently we just queue operations for migrations etc. It may cause ANRs in android for example if the datastore is created on the UI thread. I'd expect other calls into the queue via the datastore eg `changes` to be on a background thread.  

We could get around that by not waiting on the version number to be returned, and by using an in memory database to get the version, but we'd have to be careful that on older devices that we still support because of their android version that we don't cause ANRs. 

We'd also then have to set the thread name via the SQLCallable, since we couldn't wait for the name to be returned. It would mean that the SQLCallable would become coupled to the Queue, which I think is the wrong way to go.

Instead of asking for the version every time, what we could do is cache the result in a field so if it's not `null` we return straight away otherwise wait for the task to be processed on the queue.

ricellis: Save the result to a field for now - although I suspect retrieving the version to be much faster than a query because it is probably just a constant we should avoid the need to go native every time we ask for it.

We could consider moving extraction of the version into the validation as part of #292 and having it available from the `SQLDatabase` as well, which would open up opportunities for us to add the version to exceptions at a lower level, something to consider while we figure out the places it would be useful to record it.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
```diff
@@ -189,4 +195,30 @@ public boolean isShutdown() {
...
+        try {
+            return this.submit(new SQLQueueCallable<String>() {
+                @Override
+                public String call(SQLDatabase db) throws Exception {
+                    Cursor cursor = db.rawQuery("SELECT sqlite_version() AS sqlite_version", null);
```
ricellis: Do we really need `AS sqlite_version` here? We don't appear to use the alias.

rhyshort: I don't know tbh, I just copied the query from the ticket #325

ricellis: It works without it if I run it using sqlite3 command line.

rhyshort: done in 651b477

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
rhyshort: Result saved to field in 651b477

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
```diff
@@ -200,25 +201,32 @@ public boolean isShutdown() {
...
+                        return sqliteVersion;
                     }
-                    return sqliteVersion;
-                }
-            }).get();
-        } catch (InterruptedException e) {
-            logger.log(Level.WARNING, "Could not determine SQLite version", e);
-        } catch (ExecutionException e) {
-            logger.log(Level.WARNING, "Could not determine SQLite version", e);
+                }).get();
```
ricellis: I think you need to `return` after this to avoid overwriting your value from the db with `"unknown"`. Or set `"unknown"` inside both the catch blocks.

rhyshort: fixed in 95fa7dc

ricellis:
+1

PR #330, opened 2016-08-12
Prevent force insertion of duplicate revisions

### LINKED ISSUE #329: Force insert will insert revisions we already have.
`DatastoreImpl.forceInsert` will insert any revision provided even if it is already in the database. This can cause multiple rooted trees of the same revision if there are two replicators running from the same source in parallel.  

It can be reproduced by doing the following:

```java
DocumentRevision rev = /* create revision*/
datastore.forceInsert(rev, "1-rev"); 
datastore.forceInsert(rev, "1-rev");  // inserts second root with same revision.
```

To prevent insertion of a revision we already have, forceInsert should check if it contains a revisions with the matching id and rev, if it does it should abort the insert.

rhyshort:
## What

Added additional testing for preventing duplicate insertions into the database by force insert.

## Why

Without checking for the document id and revision pair in the database it is possible that two replicators running in parallel will insert the same document at the same revision into the database, which can cause multiple roots to be created. This exposed a problem with the
`pickWinnerOfConlficts` method which will be fixed as part of #327.
The database schema constraints were updated in #338 and will prevent this type of insertion with errors.

## Testing
New tests:
* `DatastoreImplForceInsertTest.forceInsert_sameRevisionTwice()` - to validate that a `DocumentException` results if the same revision is force inserted multiple times
* `PullStrategyTest.dual_pull_replications()` - to validate that two pull replications running concurrently will have one error and one complete

## Issues
Fixes #329

### DIFF
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreImplForceInsertTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreImplForceInsertTest.java
@@ -17,11 +17,16 @@ package com.cloudant.sync.datastore;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.hasItems;
 import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
+import static org.hamcrest.core.Is.is;
 
 import com.cloudant.android.ContentValues;
+import com.cloudant.sync.event.Subscribe;
+import com.cloudant.sync.notifications.DocumentModified;
+import com.cloudant.sync.sqlite.Cursor;
 import com.cloudant.sync.sqlite.SQLDatabase;
 import com.cloudant.sync.sqlite.SQLQueueCallable;
 import com.cloudant.sync.util.CouchUtils;
+import com.cloudant.sync.util.DatabaseUtils;
 import com.cloudant.sync.util.JSONUtils;
 import com.cloudant.sync.util.TestUtils;
 
@@ -251,6 +256,17 @@ public class DatastoreImplForceInsertTest {
         DocumentRevision obj = datastore.getDocument(OBJECT_ID);
         Assert.assertEquals(remoteRevisionId6, obj.getRevision());
         Assert.assertTrue(Arrays.equals(bodyOne.asBytes(), obj.getBody().asBytes()));
+    }
+
+    @Test(expected = DocumentException.class)
+    public void forceInsert_sameRevisionTwice() throws Exception {
+        DocumentRevision rev = createDbObject("1-rev", bodyOne);
+        EventSubscriber eventSubscriber = new EventSubscriber();
+        datastore.getEventBus().register(eventSubscriber);
+        datastore.forceInsert(rev, "1-rev");
+        Assert.assertThat(datastore.getDocumentCount(), is(1));
+        Assert.assertThat(eventSubscriber.eventCount, is(1));
+        datastore.forceInsert(rev, "1-rev");
     }
 
     @Test
@@ -620,4 +636,12 @@ public class DatastoreImplForceInsertTest {
         }
     }
 
+    public static class EventSubscriber {
+        int eventCount = 0;
+
+        @Subscribe
+        public void event(DocumentModified documentModified) {
+            eventCount++;
+        }
+    }
 }
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/PullStrategyTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/PullStrategyTest.java
@@ -36,9 +36,11 @@ import org.junit.experimental.categories.Category;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 @Category(RequireRunningCouchDB.class)
 public class PullStrategyTest extends ReplicationTestBase {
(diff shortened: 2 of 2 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1504,19 +1506,21 @@ public int compare(String r1, String r2) {
...
+                                                                              List<String> revHistory)
+            throws AttachmentException, SQLException {
         logger.entering("DocumentRevision",
                 "doForceInsertNewDocumentWithHistory()",
                 new Object[]{rev, revHistory});
 
         long docNumericID = insertDocumentID(db,rev.getId());
         long parentSequence = 0L;
         for (int i = 0; i < revHistory.size() - 1; i++) {
+            //TODO: We could just continue here, if we have a dup we can just ignore
```
tomblench: We should probably have a warning in the log here - it might help us in the future to know that customers are running multiple replications in parallel etc.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1163,18 +1164,22 @@ public Object call(SQLDatabase db) throws Exception {
...
                             // TODO fetch the parent doc?
                             documentUpdated = new DocumentUpdated(null, item.rev);
                         } else {
-                            seq = doForceInsertNewDocumentWithHistory(db, item.rev, item.revisionHistory);
-                            item.rev.initialiseSequence(seq);
+                            result  = doForceInsertNewDocumentWithHistory(db, item.rev, item.revisionHistory);
                             documentCreated = new DocumentCreated(item.rev);
                         }
 
+                        item.rev.initialiseSequence(result.sequence);
```
tomblench: I think the call to `initialiseSequence` was back up there so that the `document{Created,Updated}` event would have the correct sequence on it.

Also, we should probably check if it's a duplicate before deciding whether to publish the event.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreImplForceInsertTest.java
```diff
@@ -251,6 +257,39 @@ public void forceInsert_sameLengthOfPath_remoteRevisionWins() throws Exception {
         Assert.assertTrue(Arrays.equals(bodyOne.asBytes(), obj.getBody().asBytes()));
     }
 
+    public void forceInsert_sameRevisionTwice() throws Exception {
```
tomblench: Check that events get published correctly?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
rhyshort: The events don't hold the seq directly, so using the value of reference semantics we can mutate the DocumentRevs before sending the events. And we don't send the events if its a duplicate, because the remaining part of the loop is skipped if there is a duplicate.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreImplForceInsertTest.java
rhyshort: done in 560fb02

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
rhyshort: done in 6199c90

tomblench: OK, you can remove the comment now

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreImplForceInsertTest.java
```diff
@@ -251,6 +259,44 @@ public void forceInsert_sameLengthOfPath_remoteRevisionWins() throws Exception {
...
+        final String sql = "SELECT docs.docid, revid from revs join docs on revs.doc_id  == docs.doc_id where docs.docid = ? and revs.revid = ? ";
+
+
+        final DocumentRevision rev = createDbObject("1-rev", bodyOne);
+        EventSubscriber eventSubscriber = new EventSubscriber();
+        datastore.getEventBus().register(eventSubscriber);
+        datastore.forceInsert(rev, "1-rev");
+        Assert.assertThat(datastore.getDocumentCount(), is(1));
+        Assert.assertThat(eventSubscriber.eventCount, is(1));
+        datastore.getEventBus().unregister(eventSubscriber);
```
tomblench: An easier way to do this is to skip these two lines (the `unregister` and the `new`) and just check that `eventCount` is still 1

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
tomblench: OK makes sense

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1504,19 +1506,28 @@ public int compare(String r1, String r2) {
...
      *                   as well) sorted in ascending order.
      */
-    private long doForceInsertNewDocumentWithHistory(SQLDatabase db,
-                                                     DocumentRevision rev,
-                                                     List<String> revHistory)
-            throws AttachmentException{
+    private InsertRevisionCallable.Result doForceInsertNewDocumentWithHistory(SQLDatabase db,
+                                                                              DocumentRevision rev,
+                                                                              List<String> revHistory)
+            throws AttachmentException, SQLException {
```
tomblench: Why has the method signature changed to include this exception?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
ricellis: Because the sequence is now held in the `InsertRevisionCallable.Result` we no longer use the `long seq` field so that can be deleted.

ricellis:
Added test for concurrent replications in 18575ea

ricellis:
I'm +1 on @rhyshort added test.
@rhyshort @tomblench are you happy with the test I added?

rhyshort:
Seems fine to me

ricellis:
Looks like the travis testing isn't fast enough to hit the window for the dual replications, updated the assertion for at least one completion in 309d648.

tomblench:
+1

PR #334, opened 2016-08-15
Simplify `pickWinnerOfConflicts`:

tomblench:
- Mark winning sequence current
- Mark all non winning leaf sequences not current

Also remove `changeDocumentToBeCurrent`, `changeDocumentToBeNotCurrent`
and simplify `getCurrent` to take a sequence number argument.

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
@@ -754,7 +754,7 @@ public class DatastoreImpl implements Datastore {
                         , docId));
             }
             // if we got here, parent rev was deleted
-            this.setCurrent(db, potentialParent, false);
+            this.setCurrent(db, potentialParent.getSequence(), false);
             callable.revId = CouchUtils.generateNextRevisionId(potentialParent.getRevision());
             callable.docNumericId = potentialParent.getInternalNumericId();
             callable.parentSequence = potentialParent.getSequence();
@@ -855,7 +855,7 @@ public class DatastoreImpl implements Datastore {
             throw new ConflictException("Revision to be updated is not current revision.");
         }
 
-        this.setCurrent(db, preRevision, false);
+        this.setCurrent(db, preRevision.getSequence(), false);
         String newRevisionId = this.insertNewWinnerRevision(db, body, preRevision);
         return this.getDocumentInQueue(db, preRevision.getId(), newRevisionId);
     }
@@ -895,7 +895,7 @@ public class DatastoreImpl implements Datastore {
         if (prevRevision.isDeleted()) {
             throw new DocumentNotFoundException("Previous Revision is already deleted");
         }
-        setCurrent(db, prevRevision, false);
+        setCurrent(db, prevRevision.getSequence(), false);
         String newRevisionId = CouchUtils.generateNextRevisionId(prevRevision.getRevision());
         // Previous revision to be deleted could be winner revision ("current" == true),
         // or a non-winner leaf revision ("current" == false), the new inserted
@@ -1362,7 +1362,7 @@ public class DatastoreImpl implements Datastore {
             long seq = getSequenceInQueue(db, newRevision.getId(), revId);
             if (seq == -1) {
                 seq = insertStubRevision(db,docNumericID, revId, parentSeq);
-                this.changeDocumentToBeNotCurrent(db, parentSeq);
+                this.setCurrent(db, parentSeq, false);
             }
             parentSeq = seq;
         }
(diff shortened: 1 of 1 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1423,15 +1423,14 @@ private long insertDocumentHistoryToNewTree(SQLDatabase db, DocumentRevision new
         callable.available = !newRevision.isDeleted();
         long newLeafSeq = callable.call(db);
 
-        pickWinnerOfConflicts(db, docNumericID, newRevision.getId(), previousLeafSeq);
+        pickWinnerOfConflicts(db, docNumericID, newRevision.getId());
```
emlaver: Do we need to keep the `previousLeafSeq` variable? 
https://github.com/cloudant/sync-android/pull/334/files#diff-3171a584301b2c5d64276721c439450aR1405

rhyshort:
@tomblench Looking at the call sites for `setCurrent`, I expect the `currentValue` arg to be the value we are setting the revision to, eg [Line 858](https://github.com/cloudant/sync-android/pull/334/files#diff-3171a584301b2c5d64276721c439450aR858). I expect the current flag for that doc to be set to false. Part of me thinks that the `currentValue` parameter should be the value we are setting the flag to, just to match my expectation from the method call. Thoughts?

tomblench:
@rhyshort setting the current flag in memory and setting it on the database should probably be two distinct steps. There are times where you only want to set it in the database because you're not going to return a document revision. 

Also I'd be loathe to make `setCurrent ` take a `DocumentRevision` as an argument - I prefer these SQL helpers to just take the minimum primitive values they need to get their job done. We've seen an anti-pattern in the past where temporary `DocumentRevision` etc objects need to be created in order to call methods like this and I'm keen to avoid that.

rhyshort:
That's not what I was asking. But anyway, looking at the cod again, its a tad confusing, the method signature has the boolean flag `currentValue` this makes me think that the value that is being passed in is the value it is currently in the database. Where as its actually the value we should change current to (I think). So to make it clearer the method signature should be changed to make it clearer, whether thats just the argument name or more I don't know.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1432,15 +1432,14 @@ private long insertDocumentHistoryToNewTree(SQLDatabase db, DocumentRevision new
...
         return newLeafSeq;
     }
 
 
     private void pickWinnerOfConflicts(SQLDatabase db,
                                        long docNumericId,
-                                       String docId,
-                                       long previousWinnerSeq) throws DatastoreException {
+                                       String docId) throws DatastoreException {
```
ricellis: I can't see where in this method we use the `String docId` any more (or indeed where it was used before this change). Can we get rid of it?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
rhyshort: I think we should remove it. looking at the call site of this method, we have already checked if the parent exists, otherwise we wouldn't hit this call. So I think it's safe to remove and it'll save us a database access to :)

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1857,10 +1850,10 @@ private String insertNewWinnerRevision(SQLDatabase db,DocumentBody newWinner,
         return newRevisionId;
     }
 
-    private void setCurrent(SQLDatabase db,DocumentRevision winner, boolean currentValue) {
+    private void setCurrent(SQLDatabase db, long sequence, boolean currentValue) {
```
ricellis: Regarding the other discussion around the naming of `currentValue` I can see how it could be confusing if taken in the context of something like a `checkAndSet`. The meaning here is clearly `valueOfCurrentToWrite` and that could be added in javadoc for the method or just renaming the field to `valueOfCurrent` to make it clear that current is the thing we're talking about and not some state of the value.

tomblench: Good idea - see a0fcbf1

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
tomblench: Thanks for spotting, see aebc7d2

ricellis:
+1 from me

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
tomblench: Thanks for spotting @emlaver and @rhyshort - done in bb863d4

rhyshort:
+1

PR #336, opened 2016-08-16
Added 100 to 200 migration

### LINKED ISSUE #331: Migration of datastores impacted by #326 and #329
Datastores impacted by the bug #329 may be in a state with multiple identical revisions of the same document. Further to that if they were subsequently impacted by #326 then they may not have a current winner for a document.
This issue is to create a migration step to re-graft all branches of duplicate revisions to a single copy and re-evaluate the winner.

ricellis:
*What*

Added migration for repairing issues caused by #326 and #329; fixes #331.

*How*

* Created a version 100 to 200 implementation of `Migration` that:
    * Removes duplicate revisions and repairs the document trees.
    * Re-evaluates the correct current winner after the repair.
    * Moves all attachments from duplicate revisions to the lowest sequence and then deletes all but one of any duplicate attachment entries. Attachment duplication or attachment to other identical revisions could have been caused in the multipart pull case.
* Updated CHANGES.md.

*Testing*

Added `DatastoreSchemaTests#migration100DatabaseWithDuplicatesTo200` that:
* Unzips an existing datastore that includes duplicate revision IDs and some docs
with no current winner as well as duplicate attachments.
* Validates that migration of that datastore succeeds.
* Validates that after migration no documents remain with duplicate revisions.
* Validates that after migration no documents remain without a current winner.
* Validates that after migration duplicate attachments have been correctly consolidated.

`DatastoreSchemaTests#migration100ComplexDatabaseWithoutDuplicatesTo200` as above but with a complex database that is in a normal state.

Database200MigrationTest adds test cases for specific duplicate revisions in trees.


CHANGES.md
```diff
@@ -15,6 +15,11 @@
   directory path in the scope of the `DatastoreManager` class.
 - [FIX] Corrected a case where two root nodes with identical revision IDs prevented selection of the
   correct new winning revision.
+- [FIX] Added migration on `Datastore` opening to repair datastores that have documents with
+  multiple identical revision IDs (as caused by issue 329). The migration will delete duplicate
```
rhyshort: I think we should make issue 329 a link.

CHANGES.md
```diff
@@ -15,6 +15,11 @@
...
+  revisions and correct the document tree. It will also re-evaluate the winning revision resolving
+  issues caused by 326. Finally it will correct any duplicate attachment entries caused by
```
rhyshort: make 326 a link.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase100To200.java
```diff
@@ -0,0 +1,231 @@
...
+public class MigrateDatabase100To200 implements Migration {
+
+    @Override
+    public void runMigration(SQLDatabase db) throws Exception {
+        Cursor c = null;
+        try {
+            // Find docs with duplicate revision ids and get the lowest sequence for that revision
+            List<DocRevSequence> lowestDuplicateRevs = new ArrayList<DocRevSequence>();
+            c = db.rawQuery("SELECT doc_id, revid, min(sequence) FROM revs GROUP BY doc_id, revid" +
+                    " HAVING COUNT(*) > 1;", null);
```
rhyshort: I think on android `null` needs to be an empty array, it caught me out with the tests. Waiting on CI to confirm though.

CHANGES.md
tomblench: As a rule we've tended not to mention specific issue numbers in the CHANGES, because the focus is on telling a library consumer what has changed from their perspective. So I'd generally shy away from discussing internals too much here.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -187,15 +190,18 @@ public DatastoreImpl(String dir, String name, KeyProvider provider) throws SQLEx
...
         // Increment the hundreds position if a schema change means that older
         // versions of the code will not be able to read the migrated database.
-        if(dbVersion >= 200){
-            throw new DatastoreException(String.format("Database version is higher than the version supported " +
-                    "by this library, current version %d , highest supported version %d",dbVersion, 99));
+        int highestSupportedVersionExclusive = 300;
+        if (dbVersion >= highestSupportedVersionExclusive) {
+            throw new DatastoreException(String.format("Database version is higher than the " +
+                    "version supported by this library, current version %d , highest supported " +
+                    "version %d", dbVersion, highestSupportedVersionExclusive));
```
tomblench: Should that be `highestSupportedVersionExclusive-1`?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase100To200.java
```diff
@@ -0,0 +1,231 @@
...
+                // Finally resolve the winner for the document based on all the merged branches
+                reevaluateWinner(db, lowest.doc_id);
+            }
+        } finally {
+            DatabaseUtils.closeCursorQuietly(c);
+        }
+    }
+
+    /**
+     * This method copies the code from
```
tomblench: Once https://github.com/cloudant/sync-android/pull/334/ is merged you can use the new `pickWinnerOfConflicts` instead.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase100To200.java
```diff
@@ -0,0 +1,231 @@
...
+ * <p>
+ * It also deletes duplicate entries with the same sequence and attachment filename, which could
+ * have been force inserted into the attachments table during a multipart attachment pull during
+ * identical replications running in parallel.
+ * </p>
+ */
+public class MigrateDatabase100To200 implements Migration {
+
+    @Override
+    public void runMigration(SQLDatabase db) throws Exception {
```
tomblench: These migrations should probably have logging so we can see an "audit trail" of what happened during migration. Something like "No duplicates found" or "X duplicates found...deleted Y duplicate revs with ID Z"

CHANGES.md
ricellis: Since we have tended not to mention issue numbers or go into huge internal detail how about simply:
>- [FIX] Added migration on `Datastore` opening to repair datastores with duplicated revisions or attachments caused by an issue running parallel pull replications with the same source and target.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
ricellis: It should, fixed in f7bae1a

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase100To200.java
ricellis: I was hoping that would be the case :)

CHANGES.md
rhyshort: I wonder if we should explain that it might cause new revisions to be selected as the winner?

ricellis: >Note that in rare circumstances for some documents this may result in a different, but corrected, winning revision than before migration.

?

rhyshort: yeah sounds good

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase100To200.java
rhyshort: CI passed all the tests, guess this call doesn't need to be an empty array.

ricellis: The difference appears to be `rawQuery` vs `execSQL`

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreSchemaTests.java
```diff
@@ -251,4 +254,135 @@ private static File f(String filename) {
...
+     * <li>
+     * 307 a2359c3503e34c008ec448834583e482 1-b83f5c820588498ca2228282237d4f90:
+     * sequence 1051 (highest) has 2 identical attachment entries, seq 606 has no attachments
+     * </li>
+     * <li>
+     * 308 badad1b3842e4056b013587b49c93308 1-aaf13ff7c5dc44ba8626f684fb73b9d3:
+     * sequence 607 (lowest) 1050 (highest) each have one copy of each of two different attachments
+     * </li>
+     * </ul>
+     *
```
rhyshort: I think a diagram of these special cases would make it easier to understand where the attachments sit in the tree

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase100To200.java
ricellis: Added logging in 027310f & a6d3f73
It looks like:
>Aug 17, 2016 2:02:10 PM com.cloudant.sync.datastore.migrations.MigrateDatabase100To200 runMigration
INFO: Found 537 duplicated revisions.
Aug 17, 2016 2:02:10 PM com.cloudant.sync.datastore.migrations.MigrateDatabase100To200 runMigration
INFO: Deleted 2 duplicate revisions of 585:1492
...
Aug 17, 2016 2:12:19 PM com.cloudant.sync.datastore.migrations.MigrateDatabase100To200 runMigration
INFO: Migrated 2 attachments to 308:607
Aug 17, 2016 2:12:19 PM com.cloudant.sync.datastore.migrations.MigrateDatabase100To200 runMigration
INFO: Deleted 1 duplicate revisions of 308:607
Aug 17, 2016 2:12:19 PM com.cloudant.sync.datastore.migrations.MigrateDatabase100To200 runMigration
INFO: Found 2 attachments with duplicates on 308:607
Aug 17, 2016 2:12:19 PM com.cloudant.sync.datastore.migrations.MigrateDatabase100To200 runMigration
INFO: Found 2 copies of attachment on 308:607
Aug 17, 2016 2:12:19 PM com.cloudant.sync.datastore.migrations.MigrateDatabase100To200 runMigration
INFO: Deleted 1 copies of attachment on 308:607
Aug 17, 2016 2:12:19 PM com.cloudant.sync.datastore.migrations.MigrateDatabase100To200 runMigration
INFO: Found 2 copies of attachment on 308:607
Aug 17, 2016 2:12:19 PM com.cloudant.sync.datastore.migrations.MigrateDatabase100To200 runMigration
INFO: Deleted 1 copies of attachment on 308:607
Aug 17, 2016 2:12:19 PM com.cloudant.sync.datastore.migrations.MigrateDatabase100To200 runMigration
INFO: Deleted duplicates for 2 attachments on 308:607
...

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreSchemaTests.java
ricellis: Added in 6d42e68

CHANGES.md
ricellis: Updated in 02f330e

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/Database200MigrationTest.java
```diff
@@ -0,0 +1,530 @@
...
+     * 1-xxxxxxx
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testDatabaseWithComplexTree() throws Exception {
+
+        // sets up the datastore with 1-x and 1-x revisions (Duplicates)
+
+        final String sql = "INSERT INTO revs (doc_id, parent, current, deleted, available, revid, json) SELECT doc_id, parent, current, deleted, available, revid, json FROM revs where revs.revid = ?";
```
ricellis: This constant is repeated in many tests, I suggest pulling it up to the class level.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/Database200MigrationTest.java
```diff
@@ -0,0 +1,530 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * Created by rhys on 15/08/2016.
+ */
+public class Database200MigrationTest {
+
+    private Datastore ds;
```
ricellis: Why not just make this a DatastoreImpl and do the cast once in setUp, I think it will reduce the clutter.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/Database200MigrationTest.java
```diff
@@ -0,0 +1,530 @@
...
+    @After
+    public void tearDown() throws Exception {
+        ds.close();
+        TestUtils.deleteTempTestingDir(dir);
+    }
+
+    /**
+     * Tree:
+     *
+     * abc:
```
ricellis: `1-abc`?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/Database200MigrationTest.java
```diff
@@ -0,0 +1,530 @@
...
+        }).get();
+
+        DocumentRevision rev = new DocumentRevision(rootRevision.id, "1-abc");
+        rev.setBody(rootRevision.getBody());
+        ((DatastoreImpl) ds).forceInsert(rev, "1-abc");
+        runMigration();
+
+
+        int count = revisionCount();
+        Assert.assertThat(count, is(2));
```
ricellis: I think this needs more description about why the expected result is 2.

As a general rule I find having a "should" message on assertions adds clarity when reading test code and in error messages. 
e.g.
`Assert.assertThat("A duplicate revision should be removed", count, is(2))`
results in:
 `A duplicate revision should be removed: Expected 2 was 3`
vs:
 ` Expected 2 was 3`

As an aside it confuses me every time that the hamcrest assertion is in the opposite order to the normal JUnit `expected, actual`!

ricellis: We should probably also assert that the remaining revisions are the ones we expect, e.g. this test would pass if we were left with two copies of `1-xxxxxx` and no `1-abc`.

ricellis:
@rhyshort in general for the testing I think the comments https://github.com/cloudant/sync-android/pull/336/files#r75153045 and https://github.com/cloudant/sync-android/pull/336#discussion_r75153277 apply for all the tests.

I also think it is pretty confusing that the trees don't show the `rootRevision` and the duplicates are also `1-` revisions. That makes it pretty hard to follow what is happening. It might be worth updating the diagrams/revs.

rhyshort:
@ricellis They do show the root revision. The root revision is just the initial revision that is inserted to build the the tree from, so in the case of duplicates at the root level, you get multiple roots.

ricellis:
@rhyshort I was confused by the assertion for 2 entries after migration. As discussed you'll fix the tree to show the other missing revision that was confusing me!

rhyshort:
@ricellis yup fixed in 71e85f5

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/Database200MigrationTest.java
rhyshort: done in 71e85f5

rhyshort: Everything but check the exact revs are there.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/Database200MigrationTest.java
rhyshort: done in 71e85f5

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/Database200MigrationTest.java
rhyshort: done in 71e85f5

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/Database200MigrationTest.java
rhyshort: revs check done in cb36853

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/Database200MigrationTest.java
```diff
@@ -100,7 +102,7 @@ public Void call(SQLDatabase db) throws Exception {
 
 
         int count = revisionCount();
-        Assert.assertThat(count, is(2));
+        Assert.assertThat("That a duplicate root has been removed.", count, is(2));
```
ricellis: This message reads really nicely in the code, but it makes for a really confusing failure message because it makes it sound like the failure is because the duplicate has been removed not because it hasn't.

>java.lang.AssertionError: That a duplicate root has been removed.
Expected: is <2>
     but: was <3>

This is why I recommended using "should" because it tends to read OK both ways e.g.

>java.lang.AssertionError: The duplicate root should be removed.
Expected: is <2>
     but: was <3>

`Assert.assertThat("The duplicate root should be removed.", count, is(2)`

Do you see what I mean?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreSchemaTests.java
```diff
@@ -405,4 +406,218 @@ public Integer call(SQLDatabase db) throws Exception {
...
+     * - Multiple leaf nodes (conflicts)
+     * - Attachments
+     * - Deleted revs
+     * @throws Exception
+     */
+    @Test
+    public void migration100ComplexDatabaseWithoutDuplicatesTo200() throws Exception {
+        // Extract database to temp folder
+        File temp_folder = new File(TestUtils.createTempTestingDir(this.getClass().getName()));
+        File zippedWithDups = f("fixture/v100ComplexWithoutDuplicates.zip");
```
ricellis: I know this is just a copy/paste, but we should rename this field for this test to avoid confusion.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreSchemaTests.java
```diff
@@ -405,4 +406,218 @@ public Integer call(SQLDatabase db) throws Exception {
...
+
+        } finally {
+            datastore.close();
+        }
+    }
+
+    // utility methods used by tests
+    private void assertLeafCount(Datastore datastore, String docId, int expectedLeafCount) {
+        int actualLeafCount = ((DatastoreImpl)datastore).getAllRevisionsOfDocument(docId).leafRevisions().size();
+        Assert.assertEquals(expectedLeafCount, actualLeafCount);
```
ricellis: Add a message, "Should get the expected number of leaves" so we get an obvious description of what has gone wrong if it fails:
>java.lang.AssertionError: Should get the expected number of leaves 
Expected :2357
Actual   :2310

is much easier to deal with than if there is no message.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreSchemaTests.java
```diff
@@ -405,4 +406,218 @@ public Integer call(SQLDatabase db) throws Exception {
...
+        int actualAttachmentCount = 0;
+        Set<String> revIds = ((DatastoreImpl)datastore).getAllRevisionsOfDocument(docId).leafRevisionIds();
+        for (String revId : revIds) {
+            try {
+                actualAttachmentCount += datastore.getDocument(docId, revId).attachments.size();
+            } catch (DocumentNotFoundException dnfe) {
+                Assert.fail("Failed to find revision for document ID " + docId + " and revision ID " + revId);
+            }
+        }
+        Assert.assertEquals(expectedAttachmentCount, actualAttachmentCount);
```
ricellis: Message please see https://github.com/cloudant/sync-android/pull/336#discussion_r75153045 for my opinion on this

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreSchemaTests.java
```diff
@@ -405,4 +406,218 @@ public Integer call(SQLDatabase db) throws Exception {
...
+    }
+
+    private void assertWinner(Datastore datastore, String docId, String expectedRevId) {
+        try {
+            String actualRevId = datastore.getDocument(docId).getRevision();
+            Assert.assertEquals(expectedRevId, actualRevId);
```
ricellis: And one more message

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/Database200MigrationTest.java
rhyshort: I do, but i think its generally more important to read better in code, than read better in the report.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreSchemaTests.java
tomblench: All done in 28f03a3

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatastoreSchemaTests.java
tomblench: Fixed in f2a5cec

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/Database200MigrationTest.java
tomblench: "assert that that a duplicate root has been removed" ← code
"java.lang.AssertionError: That a duplicate root has been removed" ← message
neither of those look good to me

ricellis: If anything the failure message in the report is more important as it is, more often than not, read in isolation. On the other hand whilst reading code the surrounding code and nearby javadoc etc provides some additional context that is not available in the report.
You don't have to follow my advice about using "should", but I would like the messages to at least make sense in a failure case.

ricellis:
I'm +1 on the extra tests thanks @rhyshort & @tomblench

rhyshort:
+1 on the migration, just needs to pick winner to split out before merging.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/migrations/MigrateDatabase100To200.java
ricellis: @tomblench @rhyshort Extracted the common pick winner code in c6e1430. I made it a static instead of a callable because it is (currently) always called from inside other callables so it doesn't ever make sense to submit it to the queue because it would always get blocked behind its caller on the single thread executor.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/PickWinningRevision.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+/**
+ * Callable that evaluates leaf nodes to identify the winning revision for a particular document ID.
+ * Marks all revisions as not current, before marking a single winner.
+ *
+ * @api_private
+ */
+public class PickWinningRevision {
```
tomblench: Is there an implicit convention where these classes are called `XCallable`? Also I wonder if the class and method should be consistently named - pick or mark?

rhyshort: Yes, the naming scheme is `xCallable` and you then call `call` on the object, so I think we'd get `new PickWinningRevisionCallable().call()` although since there is currently no configuration required, `PickWinningRevisionCallable.call()` is fine.

rhyshort: I take the `PickWinningRevisionsCallable.call()` back, you configure the object with the `docNumericId`

ricellis: As per my comment at https://github.com/cloudant/sync-android/pull/336#discussion_r75481031 this is not actually a `Callable`.

I preferred `mark` to `pick` for the method because it actually changes something - I'm happy to update the class to match or rename both if someone can think of something better.

ricellis: Following our discussion and with vague thoughts about what we'll do in #291 made pick winner a callable after all in a0e57d4

rhyshort:
+1

PR #338, opened 2016-08-22
333 add schema constraint

tomblench:
_What_: Migrate schema version 200 to new tables for `revs` and `attachments`

_Why_: These schema constraints should have been in the original DDL to prevent duplicate revs and attachments (see also #329). Their omission means that certain logic errors were not caught by SQLite. By migrating to new tables with these constraints added (using the recipe outlined at https://www.sqlite.org/lang_altertable.html) we help to ensure the database can't get into an inconsistent state in the future.

_Testing_: Existing/updated tests pass

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreConstants.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreConstants.java
@@ -125,4 +125,46 @@ class DatastoreConstants {
         };
     }
 
+    // NB schema version 100 updates are defined in MigrateDatabase6To100 class
+
+    public static String[] getSchemaVersion200() {
+        return new String[]{
+                // migrate revs to new schema which is identical except for added constraint
+                // "UNIQUE (doc_id, revid)"
+                "    CREATE TABLE new_revs ( " +
+                "        sequence INTEGER PRIMARY KEY AUTOINCREMENT, " +
+                "        doc_id INTEGER NOT NULL REFERENCES docs(doc_id) ON DELETE CASCADE, " +
+                "        parent INTEGER REFERENCES revs(sequence) ON DELETE SET NULL, " +
+                "        current BOOLEAN, " +
+                "        deleted BOOLEAN DEFAULT 0, " +
+                "        available BOOLEAN DEFAULT 1, " +
+                "        revid TEXT NOT NULL, " +
+                "        json BLOB, " +
+                "        UNIQUE (doc_id, revid)); ",
+                "    INSERT INTO new_revs (sequence, doc_id, parent, current, deleted, available, revid, json) SELECT sequence, doc_id, parent, current, deleted, available, revid, json FROM revs; ",
+                "    DROP TABLE revs; ",
+                "    ALTER TABLE new_revs RENAME TO revs; ",
+                "    CREATE INDEX revs_by_id ON revs(revid, doc_id); ",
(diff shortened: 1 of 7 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreConstants.java
```diff
@@ -125,4 +125,46 @@
...
+        return new String[]{
+                // migrate revs to new schema which is identical except for added constraint
+                // "UNIQUE (doc_id, revid)"
+                "    CREATE TABLE new_revs ( " +
+                "        sequence INTEGER PRIMARY KEY AUTOINCREMENT, " +
+                "        doc_id INTEGER NOT NULL REFERENCES docs(doc_id) ON DELETE CASCADE, " +
+                "        parent INTEGER REFERENCES revs(sequence) ON DELETE SET NULL, " +
+                "        current BOOLEAN, " +
+                "        deleted BOOLEAN DEFAULT 0, " +
+                "        available BOOLEAN DEFAULT 1, " + // TODO should we remove `available`?
```
ricellis: I think the answer is not at this time.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/ForceInsertTest.java
```diff
@@ -84,14 +86,20 @@ public void notification_forceinsertWithAttachments() throws Exception {
 
         ArrayList<String> revisionHistory = new ArrayList<String>();
         revisionHistory.add(doc1_rev1.getRevision());
+        revisionHistory.add("2-revision");
+
+        // We need to create a "clean" document for force insert of the attachment
+        // manager will add the attachments to seq 1 instead of seq 2.
```
ricellis: This comment needs to be deleted as I don't think the attachments are on seq 1 and we don't really have any indication in this code what is seq 1 and what is seq 2.

Besides that I think it would be worth saying that we are force inserting 2-revision which has attachments and what we are expecting a notification for.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
```diff
@@ -206,6 +206,8 @@ public static void updateSchema(SQLDatabase database, Migration migration, int v
                 }
             } finally {
                 database.endTransaction();
+                // re-enable foreign keys
+                database.execSQL("PRAGMA foreign_keys = ON;");
```
ricellis: Do we need to surround `endTransaction()` in a try/finally as well to ensure we turn the foreign keys back on? I worry if it could throw a RuntimeException even if it doesn't declare any other exceptions - although at that point we probably have bigger problems.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreConstants.java
tomblench: Done in 	ab8f6bf

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/ForceInsertTest.java
tomblench: Done in 0f1221c

ricellis:
+1

PR #340, opened 2016-08-25
refactor datastore callables

tomblench:
See #291

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLQueueCallable.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLQueueCallable.java
@@ -22,10 +22,43 @@ import java.util.concurrent.Callable;
  *
  * @api_private
  */
-public abstract class SQLQueueCallable<T> implements Callable<T> {
+class SQLQueueCallable<T> implements Callable<T> {
 
     private SQLDatabase db;
-    private boolean runInTransaction = false;
+    private boolean runInTransaction;
+    private SQLCallable<T> sqlCallable;
+
+    /**
+     * <p>
+     *     Create a SQLQueueCallable object which will invoke the SQLCallable on the SQLDatabase
+     * </p>
+     * <p>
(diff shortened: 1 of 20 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLQueueCallable.java
```diff
@@ -22,54 +22,59 @@
...
+     *     Create a SQLQueueCallable object which will invoke the SQLCallable on the SQLDatabase
+     * </p>
+     * <p>
+     *     The SQLCallable will be invoked in a new transaction if runInTransaction is true
+     * </p>
+     * @param db The SQLDatabase to use for invoking
+     * @param runInTransaction Whether to invoke the SQLCallable in a new transaction
+     * @param sqlCallable The SQLCallable to invoke
+     */
+    SQLQueueCallable(SQLDatabase db, boolean runInTransaction, SQLCallable<T> sqlCallable) {
```
rhyshort: Small stylistic thing, shouldn't `runInTransaction` be the final argument so its easier to recognise thats its method overloading?

tomblench: Good point, done in 5431756

rhyshort:
+1

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLQueueCallable.java
```diff
@@ -22,54 +22,59 @@
  *
  * @api_private
  */
-public abstract class SQLQueueCallable<T> implements Callable<T> {
+public class SQLQueueCallable<T> implements Callable<T> {
```
ricellis: Since this wrapper is only used by the `SQLDatabaseQueue` now it can be made less visible.

ricellis:
Just the one visibility thing we discussed
+1

PR #342, opened 2016-09-01
Move most SQL code out of DatastoreImpl

tomblench:
This is achieved by moving the SQL code out of the various `*InQueue`
methods or methods in anonymous classes. Instead each 'operation'
becomes a `*Callable` class which overrides the `call()` method.

This change widens access of some fields and methods from
package-private to public, due to the fact that callables live in their
own package. This may not be entirely desirable and other work-arounds
could be considered.

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
@@ -1,16 +1,16 @@
 /**
  * Copyright © 2016 IBM Corp. All rights reserved.
- *
+ * <p/>
  * Original iOS version by  Jens Alfke, ported to Android by Marty Schoch
  * Copyright (c) 2012 Couchbase, Inc. All rights reserved.
- *
+ * <p/>
  * Modifications for this distribution by Cloudant, Inc., Copyright (c) 2013 Cloudant, Inc.
- *
+ * <p/>
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
- *
+ * <p/>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p/>
  * Unless required by applicable law or agreed to in writing, software distributed under the
  * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
  * either express or implied. See the License for the specific language governing permissions
@@ -21,10 +21,20 @@ package com.cloudant.sync.datastore;
 
 import com.cloudant.android.Base64InputStreamFactory;
 import com.cloudant.android.ContentValues;
+import com.cloudant.sync.datastore.callables.DeleteDocumentCallable;
 import com.cloudant.sync.datastore.callables.GetAllDocumentIdsCallable;
+import com.cloudant.sync.datastore.callables.GetAllRevisionsOfDocumentCallable;
+import com.cloudant.sync.datastore.callables.GetDocumentCallable;
+import com.cloudant.sync.datastore.callables.GetDocumentCountCallable;
+import com.cloudant.sync.datastore.callables.GetDocumentsWithInternalIdsCallable;
+import com.cloudant.sync.datastore.callables.GetLastSequenceCallable;
+import com.cloudant.sync.datastore.callables.GetLocalDocumentCallable;
+import com.cloudant.sync.datastore.callables.GetNumericIdCallable;
 import com.cloudant.sync.datastore.callables.GetPossibleAncestorRevisionIdsCallable;
+import com.cloudant.sync.datastore.callables.GetSequenceCallable;
 import com.cloudant.sync.datastore.callables.InsertRevisionCallable;
 import com.cloudant.sync.datastore.callables.PickWinningRevisionCallable;
+import com.cloudant.sync.datastore.callables.SetCurrentCallable;
 import com.cloudant.sync.datastore.encryption.KeyProvider;
 import com.cloudant.sync.datastore.encryption.NullKeyProvider;
 import com.cloudant.sync.datastore.migrations.MigrateDatabase100To200;
@@ -60,7 +70,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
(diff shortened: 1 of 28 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -80,32 +88,35 @@
     private static final String LOG_TAG = "BasicDatastore";
     private static final Logger logger = Logger.getLogger(DatastoreImpl.class.getCanonicalName());
 
-    private static final String FULL_DOCUMENT_COLS = "docs.docid, docs.doc_id, revid, sequence, json, current, deleted, parent";
+    public static final String FULL_DOCUMENT_COLS = "docs.docid, docs.doc_id, revid, sequence, " +
+            "json, current, deleted, parent";
 
-    private static final String GET_DOC_NUMERIC_ID =
+    public static final String GET_DOC_NUMERIC_ID =
```
ricellis: Do you think there is any value in moving these into the `SQLCallable`s that use them? IIRC most are used only once. If they are more widely shared a better option might be putting them in a separate constants class. I guess if we wanted them more tightly bound to the callables they could be added to the `SQLCallable` interface or an abstract super. I just think it is weird keeping the SQL statements here now that this class basically needs to know nothing about SQL.

tomblench: There's an argument for keeping them all in one place (maybe), but it certainly isn't `DatastoreImpl` any more.

I'm happy to either move them out to a constants class or have them declared in each Callable.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -209,40 +223,14 @@ public long getLastSequence() {
...
         } catch (InterruptedException e) {
-            logger.log(Level.SEVERE,"Failed to get last Sequence",e);
+            logger.log(Level.SEVERE, "Failed to get last Sequence", e);
             throw new RuntimeException(e);
         } catch (ExecutionException e) {
             logger.log(Level.SEVERE, "Failed to get last Sequence", e);
-            if(e.getCause()!=null){
-                if(e.getCause() instanceof IllegalStateException){
+            if (e.getCause() != null) {
+                if (e.getCause() instanceof IllegalStateException) {
```
ricellis: I know this is unrelated and has only been drawn to my attention by the whitespace change, but doesn't this effectively swallow all `DatastoreException`s thrown by the callable?
I know we are limited by the API in what we can throw here, but I think that the callable should throw the `SQLException` (which will be chained as the cause of the `ExecutionException` and we should wrap the cause of **every** `ExecutionException` in a `DatastoreException` and throw it.

ricellis: I think this would apply in other places too, perhaps another issue might be more appropriate before we grow this even further.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -482,39 +339,44 @@ public Changes call(SQLDatabase db) throws Exception {
...
+                throw new IllegalStateException("Failed to get changes, SQLite version: " + queue
+                        .getSQLiteVersion(), e);
+            }
         }
 
         return null;
 
     }
 
+    // TODO this is only called by test code - this should be removed and the test code should use the callable directly
```
ricellis: Agreed.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -595,21 +406,23 @@ public int compare(DocumentRevision documentRevision, DocumentRevision
             throw new IllegalArgumentException("limit must be >= 0");
         }
         try {
-            return queue.submit(new SQLCallable<List<DocumentRevision>>(){
+            return queue.submit(new SQLCallable<List<DocumentRevision>>() {
```
ricellis: Can't this be broken out into a class of its own as well?

tomblench: There are a few more to do, wasn't sure about making the diff even bigger

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -621,27 +434,29 @@ public int compare(DocumentRevision documentRevision, DocumentRevision
...
 
     @Override
-    public List<DocumentRevision> getDocumentsWithIds(final List<String> docIds)  throws
+    public List<DocumentRevision> getDocumentsWithIds(final List<String> docIds) throws
             DocumentException {
         Preconditions.checkState(this.isOpen(), "Database is closed");
         Preconditions.checkNotNull(docIds, "Input document id list cannot be null");
         try {
-            return queue.submit(new SQLCallable<List<DocumentRevision>>(){
+            return queue.submit(new SQLCallable<List<DocumentRevision>>() {
```
ricellis: what about this one?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
ricellis: Looking at how many times we write the pattern:
```java
try {
  return queue.submit(callable).get();
            logger.log(Level.SEVERE, "Identical message", e);
             throw new RuntimeException(e);
} catch (ExecutionException e) {
             logger.log(Level.SEVERE, "Identical message", e);
}
```
I'd be sorely tempted to wrap the `Future` returned from the `SQLDatabaseQueue.submit*` methods in another type with a modified get that does this try/catch/log for us.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
ricellis: Sounds OK to me to try and catch the rest in a second pass.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -967,50 +721,24 @@ private long insertStubRevision(SQLDatabase db, long docNumericId, String revId,
...
-            logger.log(Level.SEVERE, String.format("Error getting local document with id: %s",docId),e);
-            throw new DatastoreException("Error getting local document with id: " + docId, e);
-        } finally {
-            DatabaseUtils.closeCursorQuietly(cursor);
-        }
-    }
-
-    private List<DocumentRevision> getRevisionsFromRawQuery(SQLDatabase db, String sql, String[] args)
-            throws DocumentNotFoundException, AttachmentException, DocumentException, DatastoreException {
+    // TODO - move?
```
ricellis: Yeah I think it should move too, but it can be part of the second pass.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1786,15 +1545,17 @@ private String insertNewWinnerRevision(SQLDatabase db,DocumentBody newWinner,
      *
      * @see #pickWinnerOfConflicts(long)
      */
-    private void setCurrent(SQLDatabase db, long sequence, boolean valueOfCurrent) {
+    public static void setCurrent(SQLDatabase db, long sequence, boolean valueOfCurrent) {
```
ricellis: I note that this hasn't been marked TODO and is not so obviously a callable because it doesn't already create an anonymous `SQLCallable` as in some other places, but I think this could also be moved into a callable, which I think suits it better than a public static method.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1786,15 +1545,17 @@ private String insertNewWinnerRevision(SQLDatabase db,DocumentBody newWinner,
...
         ContentValues updateContent = new ContentValues();
         updateContent.put("current", valueOfCurrent ? 1 : 0);
         String[] whereArgs = new String[]{String.valueOf(sequence)};
         db.update("revs", updateContent, "sequence=?", whereArgs);
     }
 
-    private static DocumentRevision getFullRevisionFromCurrentCursor(Cursor cursor,
-                                                                          List<? extends Attachment> attachments) {
+    // TODO where should this be moved - DocumentRevision?
```
ricellis: There are a few of these statics hanging around, I think as part of the next pass we should list them all and see if we can break them into some reasonable groups that might better suggest where they belong (or perhaps allow us to place them appropriately in a hierarchies based on the callables that might need them (e.g. `abstract DocumentRevisionCallable` might be a possible way of containing to a specific group of functions).

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentRevision.java
```diff
@@ -242,7 +241,7 @@ long getParent() {
      *
      * @return  the internal numeric ID of this document revision.
      */
-    long getInternalNumericId(){
+    public long getInternalNumericId(){
```
ricellis: I'm opposed to this particular change because this class is API public. I think the internal numeric ID exposes way to much implementation detail to be part of a public API.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentRevision.java
```diff
@@ -18,6 +18,7 @@
 package com.cloudant.sync.datastore;
 
 import com.cloudant.common.SimpleChangeNotifyingMap;
+import com.cloudant.sync.sqlite.Cursor;
```
ricellis: Is this import used? I don't see a change that warrants adding it.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/SavedAttachment.java
```diff
@@ -30,7 +30,7 @@
  *
  * @api_public
  */
-class SavedAttachment extends Attachment {
+public class SavedAttachment extends Attachment {
```
ricellis: I'm confused why this was API public before when the class itself wasn't public!

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/SavedAttachment.java
```diff
@@ -57,7 +57,7 @@
     private final File file;
     private final AttachmentStreamFactory attachmentStreamFactory;
 
-    protected SavedAttachment(long seq, String name, byte[] key, String type, Encoding encoding,
+    public SavedAttachment(long seq, String name, byte[] key, String type, Encoding encoding,
```
ricellis: This is another awkward one I hope we might be able to reconsider this when we revisit the `AttachmentManager`.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/callables/AttachmentsForRevisionCallable.java
```diff
@@ -0,0 +1,70 @@
+package com.cloudant.sync.datastore.callables;
```
ricellis: Copyright statement needed.

ricellis: Possibly for all the new callables.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
rhyshort: I think placing it in the callable will make it easier to understand whats going on, and it has the bonus of making people use the callable rather than just run the SQL from the constants class.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
rhyshort: Looking through the changes there are lots of places we need to make this correction, so lets push it to another issue, where we should talk API impact etc.

ricellis:
I'm fine with the `DeleteDocumentCallable` changes, but noticed that the on device build failed now. Looks like the big pull replication test didn't complete in time - seems to have stalled after one batch. I don't see how that is related to the change, so possibly just an intermittent but it probably needs validating that it isn't a persistent failure.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/callables/AttachmentsForRevisionCallable.java
tomblench: Done in f9ee071

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/SavedAttachment.java
tomblench: This should have been api_private. See 19ed5c3

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/SavedAttachment.java
tomblench: Agreed. `AttachmentsForRevisionCallable` has been migrated partly because it was needed for other callables - `GetAllRevisionsOfDocumentCallable` and `GetDocumentCallable`. But I'm going to hold off on migrating any of the other `AttachmentManager` stuff because it will make this PR too big.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentRevision.java
tomblench: Fixed in 22a96ba

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
tomblench: Moved this into a callable in b50d8f8

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/callables/AttachmentsForRevisionCallable.java
ricellis: Thanks, although it pains me to say that the guidelines are to use (C) or © I'm not sure what isn't correct about the lowercase (c) - admittedly we have not fixed that throughout the library I was hoping to just get them right as we go through.

tomblench: Done in b8edf2d

ricellis:
+1

rhyshort:
+1

Issue #308, opened 2016-07-12
Attempt to remove guava dependency


ricellis:
The `compile group: 'com.google.guava', name: 'guava', version:'15.0'` dependency adds approximately 14.5k methods and 1.8 MB to the library.
While there are quite a lot of usages, it may be possible to remove the dependency and slim down the library.

ricellis:
A lot of the usages are in test code, so another option would also be to remove the runtime dependency and just have a test dependency.

ricellis:
The classes we're using at present are:
com.google.common.base.Preconditions
com.google.common.base.Strings
com.google.common.collect.ForwardingMap
com.google.common.io.Resources
com.google.common.collect.ImmutableMap
 com.google.common.collect.ImmutableSet
com.google.common.base.Joiner
com.google.common.collect.ArrayListMultimap
 com.google.common.collect.HashMultimap
 com.google.common.collect.Lists
com.google.common.collect.Multimap
com.google.common.collect.Sets
com.google.common.eventbus.EventBus

ricellis:
Did Immutable*, Lists and EventBus in these branches:
308-remove-guava-eventbus
308-remove-guava-lists
308-remove-guava-immutables

ricellis:
`Resources` and `Sets` are in 308-remove-guava-res-sets

ricellis:
`ForwardingMap` done in 308-remove-guava-forwarding.

ricellis:
`Multimap` classes are in 308-remove-guava-multimap
`Joiner` is in 308-remove-guava-joiner
`Strings` and `Preconditions` in 308-remove-guava-precon

ricellis:
Last part is in 308-remove-guava-dep

PR #344, opened 2016-09-07
308 remove guava eventbus

ricellis:
*What*

Removed Guava `EventBus`.

*How*

The `EventBus` APIs were already replaced before the 1.0.0 release so this just means updating the internal implementation to track and reflectively invoke subscribers instead of delegating to the Guava `EventBus` implementation.

*Testing*

Existing tests pass, added new `EventBusTest` to validate internal `EventBus` implementation.
Fixed a test ordering issue that was exposed by this change.

*Issues*

Part of #308

### DIFF
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright (c) 2016 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package com.cloudant.sync.event;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Before;
+import org.junit.Test;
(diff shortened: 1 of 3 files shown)


rhyshort:
I think all this needs is a test with a badly behaving subscriber to to ensure we don't break the internals in the future.

ricellis:
Badly behaved how?

rhyshort:
Throws some kind of exception.

ricellis:
Good point, test added in 68f52e2.

rhyshort:
+1

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
```diff
@@ -0,0 +1,173 @@
...
+        TestEvent e = new TestEvent();
+        bus.post(e);
+
+        assertEquals("There should be only one event received", 1, sub.eventsReceived.size());
+        assertNotEquals("The event received should not be the latest one sent", e, sub
+                .eventsReceived.get(0));
+    }
+
+    @Test
+    public void multipleEvents() throws Exception {
```
emlaver: Add test description.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
```diff
@@ -0,0 +1,173 @@
...
+        // Post the events
+        for (TestEvent e : events) {
+            bus.post(e);
+        }
+
+        assertEquals("All the events should be received", events, sub.eventsReceived);
+    }
+
+    /**
+     * Test that a subscriber throwing an exception does not c
```
emlaver: Fix incomplete description.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
ricellis: Done in 21b42d1.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
ricellis: Fixed in 21b42d1.

PR #346, opened 2016-09-07
Replaced guava ImmutableMap and ImmutableSet

ricellis:
*What*

Replaced test code usage of `ImmutableMap` and `ImmutableSet` with new
`CollectionFactory`.

*Testing*

Updated tests to use `CollectionFactory`, tests pass.

*Issues*

Part of #308


emlaver:
LGTM +1

PR #347, opened 2016-09-07
308 remove guava res sets

ricellis:
*What*

Removed use of `Sets` and `Resources`

*How*

Replaced `Resources.newInputStreamSupplier(url).getInput()` with the `ClassLoader`'s `getResourceAsStream`.

Replaced `accumulator = Sets.intersection(accumulator, childIds)` with `accumulator.retainAll(childIds)`
Replaced `accumulator = Sets.union(accumulator, childIds)`with `accumulator.addAll(childIds)`.
As it happens the pattern of using `Sets.union` in that way in a loop was explicitly advised against in the Guava javadoc so it is probably best that we replace it anyway.

*Testing*

Existing tests continue to pass.

*Issues*

Part of #308


tomblench:
+1

PR #348, opened 2016-09-07
Removed ForwardingMap

ricellis:
*What*

Removed usage of Guava `ForwardingMap`.

*How*

Replaced `SimpleChangeNotifyingMap` class usages with `ChangeNotifyingMap`
interface.
Created new instances of `ChangeNotifyingMap` with `SimpleChangeNotifyingMap.wrap` method that uses a reflection proxy to delegate to another `Map` and implements the `hasChanged` method and checks for map changes in an `InvocationHandler`.

*Testing*

Added `resolveConflictsForDocument_twoConflictAndNewWinner_newWinnerInsertedWithModifiedAttachments` to exercise the `hasChanged` path which was untested before.

*Issues*

Part of #308

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/common/SimpleChangeNotifyingMap.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/common/SimpleChangeNotifyingMap.java
@@ -1,48 +1,54 @@
+/*
+ * Copyright © 2016 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
 package com.cloudant.common;
 
-import com.google.common.collect.ForwardingMap;
-
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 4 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/common/SimpleChangeNotifyingMap.java
```diff
@@ -1,48 +1,54 @@
...
-        this.mapChanged = false;
-    }
-
-    protected Map<K, V> delegate() {
-        return delegateMap;
-    }
-
-    public boolean hasChanged() {
-        return mapChanged;
+public class SimpleChangeNotifyingMap {
```
tomblench: This is quite a clever trick, but I wonder if we can't just façade all of the `Map` methods? There are only about 14 so there's not too much boilerplate. Any comments on that?

ricellis: Well I guess I consider 14 basically identical methods too much boilerplate :)
Either approach would work and both approaches would need modification if later new `Map` methods need to modify the `hasChanged` flag. I wondered if one could argue that by façading everything we would be forced to update for new `Map` methods, but with `default` methods I don't think that is the case, so either way this class could need longer term maintenance.
Was there some particular advantage/disadvantage you had in mind?

tomblench: OK I think this approach is worth sticking with. Is there a reason for `isChanged` to be an `AtomicBoolean`?

ricellis: I was concerned about thead safety in the case that methods accessing the map came from a different thread than those calling `hasChanged`, but nearly concurrently - a `volatile boolean` would also handle that scenario - maybe I'm being too paranoid! Do you have a preference or think that we don't need to worry about it?

rhyshort:
+1

tomblench:
+1

cloudant-sync-datastore-core/src/main/java/com/cloudant/common/SimpleChangeNotifyingMap.java
ricellis: As discussed changed to `volatile boolean` in 067360d

PR #349, opened 2016-09-09
Replaced Guava Joiner with Misc static method

ricellis:
*What*

Replaced Guava `Joiner` with `Misc` static method

*How*

Added `Misc.join` method to join a collection of strings together with a
separator, skipping `null`s and not appending a final terminator.
Replaced usages of `Joiner` with `Misc.join`.

*Testing*

Existing tests pass.
Added new tests for `join` method.

*Issues*

Part of #308


rhyshort:
It makes me sad that we can't just use the goodies baked into the platforms, but hey oh, LGTM 👍

PR #350, opened 2016-09-09
308 remove guava multimap

ricellis:
*What*

Replaced Guava `Multimap` usages.

*How*

* Reworked the batching of `revsDiff` to first break down by doc ID then by groups of rev IDs up to the placeholder limit. Previously in cases where a rev ID clashed between documents the SQL statement could return revisions from the database that did not match the list of revisions passed in. Since the revision list is filtered by the `revsDiff` the impact was simply calling remove for entries that didn't exist so it did not cause any issues, but it seems more clear to avoid the ambiguity.
* Added `ValueListMap` to provide `Map<String, List<String>>` implementation
with functionality to add values to a key.
* Refactored methods using `Multimap` to use `Map<String, List<String>>`.
* Replaced `Multimap` creation with `ValueListMap`.

*Testing*

Existing tests continue to pass. No new tests as the usage of the map should be covered by the revs diff testing.

*Issues*
Part of #308

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/common/ValueListMap.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/common/ValueListMap.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright © 2016 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package com.cloudant.common;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * <p>
+ * A map implementation that stores a list of values for each key. To remain consistent
+ * with existing Map interfaces the put* methods perform replacement of keys with new
+ * lists of values, the addValue* methods should be used to add new values to the existing keys.
+ * </p>
+ * <p>
+ * Note that when using put to add a list of values for a key that later calls to addValue* will
+ * modify the original list. Consequently if the list is not modifiable then these operations will
+ * fail (typically with an UnsupportedOperationException).
+ * </p>
+ * <p>
+ * This implementation does not currently support removing some values from a key.
+ * </p>
+ *
+ * @param <K> the type of key
+ * @param <V> the type of values
+ * @api_private
+ */
+public class ValueListMap<K, V> extends ConcurrentHashMap<K, List<V>> {
+
+    /**
+     * Add a value to the map under the existing key or creating a new key if it does not
+     * yet exist.
+     *
+     * @param key   the key to associate the values with
+     * @param value the value to add to the key
+     */
+    public void addValueToKey(K key, V value) {
+        this.addValuesToKey(key, Collections.singletonList(value));
+    }
+
+    /**
+     * Add a collection of one or more values to the map under the existing key or creating a new key
+     * if it does not yet exist in the map.
+     *
+     * @param key             the key to associate the values with
+     * @param valueCollection a collection of values to add to the key
+     */
+    public void addValuesToKey(K key, Collection<V> valueCollection) {
+        if (!valueCollection.isEmpty()) {
+            // Create a new collection to store the values (will be changed to internal type by call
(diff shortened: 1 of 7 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/common/ValueListMap.java
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @param key   the key to associate the values with
+     * @param value the value to add to the key
+     */
+    public void addValueToKey(K key, V value) {
+        this.addValuesToKey(key, Collections.singletonList(value));
+    }
+
+    /**
+     * Add one or more collections of values to the map under the existing key or creating a new key
```
tomblench: should read "a collection of one or more values"

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1525,95 +1525,83 @@ boolean isOpen() {
...
-
-        String[] args = new String[revisions.keySet().size() + revisions.size()];
-        String[] keys = revisions.keySet().toArray(new String[revisions.keySet().size()]);
-        String[] values = revisions.values().toArray(new String[revisions.size()]);
-        System.arraycopy(keys, 0, args, 0, revisions.keySet().size());
-        System.arraycopy(values, 0, args, revisions.keySet().size(), revisions.size());
+                "SELECT revs.revid FROM docs, revs " +
+                        "WHERE docs.doc_id = revs.doc_id AND docs.docid = %s AND revs.revid IN " +
+                        "(%s) ",
+                DatabaseUtils.makePlaceholders(1),
```
tomblench: `DatabaseUtils.makePlaceholders(1)` should be replaced by the literal `?` in the string.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1525,95 +1525,83 @@ boolean isOpen() {
...
+                DatabaseUtils.makePlaceholders(revs.size()));
+
+        String[] args = new String[1 + revs.size()];
+        args[0] = docId;
+        // Copy the revisions into the end of the args array
+        System.arraycopy(revs.toArray(new String[revs.size()]), 0, args, 1, revs.size());
```
tomblench: this can become `String[] args = revs.toArray();` after the simplification above.

cloudant-sync-datastore-core/src/main/java/com/cloudant/common/ValueListMap.java
ricellis: Fixed in 82d1224

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
ricellis: Done in 2ba31b3.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
ricellis: We still need to pass the docId in these args to fill in the `?` placeholder. We could use `String[] args = revs.toArray()` if we substitute the docId directly into the selection string instead of `?` but in that case I think we need to do some string escaping as well.

tomblench: OK it looks ugly but now I understand that's the best way of doing it.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1525,95 +1525,81 @@ boolean isOpen() {
...
+
+                    ValueListMap<String, String> missingRevs = new ValueListMap<String, String>();
+
+                    // Break down by docId first to avoid potential rev ID clashes between doc IDs
+                    for (Map.Entry<String, List<String>> entry : revisions.entrySet()) {
+                        String docId = entry.getKey();
+                        List<String> revs = entry.getValue();
+                        // Partition into batches to avoid exceeding placeholder limit
+                        List<List<String>> batches = CollectionUtils.partition(revs,
+                                SQLITE_QUERY_PLACEHOLDERS_LIMIT - 1);
```
rhyshort: I think we need to make a comment saying why its `SQLITE_QUERY_PLACEHOLDERS_LIMIT - 1` rather than just `SQLITE_QUERY_PLACEHOLDERS_LIMIT`

rhyshort:
one small comment, once thats done +1

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
ricellis: Clarified in 4814abd.

Issue #354, opened 2016-09-13
Refactor - fix exception handling in Callables


tomblench:
See https://github.com/cloudant/sync-android/pull/342#discussion_r77188251

See #291 and #342

tomblench:
Done in #356 and #420 - the public API now throws the correct exceptions

Issue #280, opened 2016-06-22
Unchecked exception audit


ricellis:
There are far too many `RuntimeException` and other unchecked exceptions.
We should go through and remove these where possible to make error handling easier.
We should also pay particular attention to the package com.cloudant.sync.query which could use some exception handling improvements.

tomblench:
Note also that some SQL errors don't manifest themselves in obvious ways because the user has to look at the log to see what exactly happened. This is due to the way that `insert` and `insertWithOnConflict` do not throw exceptions (this is by design - it is compatible with the Android SQL API).

See below for an example log. The user can't programmatically retrieve the underlying error because `InsertRevisionCallable.call` throws this exception:

```
throw new IllegalStateException("Unknown error inserting new revision, please check log");
```

```
SEVERE: Error inserting to: revs, parent=25 current=false deleted=true revid=26-9f17ab457a9f4d38e0b8fa02e9af6db9 available=true json=[B@64accba4 doc_id=1, 
com.almworks.sqlite4java.SQLiteException: [2067] DB[1] step() [INSERT  INTO "revs"(parent,current,deleted,revid,available,json,doc_id) VALUES (?,?,?,?,?,?,?)]DB[1][C] [UNIQUE constraint failed: revs.doc_id, revs.revid]
	at com.almworks.sqlite4java.SQLiteConnection.throwResult(SQLiteConnection.java:1436)
	at com.almworks.sqlite4java.SQLiteConnection$BaseController.throwResult(SQLiteConnection.java:1689)
	at com.almworks.sqlite4java.SQLiteStatement.stepResult(SQLiteStatement.java:1402)
	at com.almworks.sqlite4java.SQLiteStatement.step(SQLiteStatement.java:301)
	at com.cloudant.sync.sqlite.sqlite4java.SQLiteWrapper.executeSQLStatement(SQLiteWrapper.java:331)
	at com.cloudant.sync.sqlite.sqlite4java.SQLiteWrapper.insertWithOnConflict(SQLiteWrapper.java:314)
	at com.cloudant.sync.sqlite.sqlite4java.SQLiteWrapper.insert(SQLiteWrapper.java:325)
	at com.cloudant.sync.datastore.callables.InsertRevisionCallable.call(InsertRevisionCallable.java:67)
	at com.cloudant.sync.datastore.DatastoreImpl.insertDocumentHistoryIntoExistingTree(DatastoreImpl.java:1377)
[...]

PR #356, opened 2016-09-14
291 refactor datastore callables part 3

tomblench:
Everything except AttachmentManager into callables

**TODO**
* ~~Check exception correctness~~ `DatastoreImpl` signatures haven't changed so they still throw the same exceptions. See https://github.com/cloudant/sync-android/issues/354 and https://github.com/cloudant/sync-android/issues/280 - these are the correct follow on PRs to use to address the other exception issues
* ~~Fix up `catch(InterruptedException e)`/ `catch (ExecutionException e)` pattern in `DatastoreImpl`~~
* ~~javadoc comments~~
* ~~Copyright headers~~
* ~~Run some proper replication tests~~ - pull and pull replication looks good with the Airfi dataset

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Changes.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Changes.java
@@ -35,7 +35,22 @@ public class Changes {
 
     private final List<DocumentRevision> results;
 
-    protected Changes(long lastSequence, List<DocumentRevision> results) {
+    /**
+     * <p>
+     * Construct a list of changes
+     * </p>
+     * <p>
+     * Note that this constructor is for internal use. To get a set of changes from a given sequence
+     * number, use {@link com.cloudant.sync.datastore.Datastore#changes}
+     * </p>
+     * @param lastSequence the last sequence number of this change set
+     * @param results the list of {@code DocumentRevision}s in this change set
+     *
+     * @see com.cloudant.sync.datastore.Datastore#changes
+     *
+     * @api_private
+     */
+    public Changes(long lastSequence, List<DocumentRevision> results) {
         Misc.checkNotNull(results, "Changes results");
         this.lastSequence = lastSequence;
         this.results = results;
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
@@ -19,23 +19,33 @@
 
 package com.cloudant.sync.datastore;
 
-import com.cloudant.android.Base64InputStreamFactory;
-import com.cloudant.android.ContentValues;
+import com.cloudant.common.ChangeNotifyingMap;
 import com.cloudant.common.ValueListMap;
+import com.cloudant.sync.datastore.callables.ChangesCallable;
+import com.cloudant.sync.datastore.callables.CompactCallable;
+import com.cloudant.sync.datastore.callables.DeleteAllRevisionsCallable;
 import com.cloudant.sync.datastore.callables.DeleteDocumentCallable;
+import com.cloudant.sync.datastore.callables.DeleteLocalDocumentCallable;
+import com.cloudant.sync.datastore.callables.ForceInsertCallable;
 import com.cloudant.sync.datastore.callables.GetAllDocumentIdsCallable;
+import com.cloudant.sync.datastore.callables.GetAllDocumentsCallable;
 import com.cloudant.sync.datastore.callables.GetAllRevisionsOfDocumentCallable;
+import com.cloudant.sync.datastore.callables.GetConflictedDocumentIdsCallable;
 import com.cloudant.sync.datastore.callables.GetDocumentCallable;
 import com.cloudant.sync.datastore.callables.GetDocumentCountCallable;
+import com.cloudant.sync.datastore.callables.GetDocumentsWithIdsCallable;
 import com.cloudant.sync.datastore.callables.GetDocumentsWithInternalIdsCallable;
 import com.cloudant.sync.datastore.callables.GetLastSequenceCallable;
 import com.cloudant.sync.datastore.callables.GetLocalDocumentCallable;
-import com.cloudant.sync.datastore.callables.GetNumericIdCallable;
 import com.cloudant.sync.datastore.callables.GetPossibleAncestorRevisionIdsCallable;
 import com.cloudant.sync.datastore.callables.GetSequenceCallable;
+import com.cloudant.sync.datastore.callables.InsertDocumentIDCallable;
+import com.cloudant.sync.datastore.callables.InsertLocalDocumentCallable;
 import com.cloudant.sync.datastore.callables.InsertRevisionCallable;
-import com.cloudant.sync.datastore.callables.PickWinningRevisionCallable;
+import com.cloudant.sync.datastore.callables.ResolveConflictsForDocumentCallable;
+import com.cloudant.sync.datastore.callables.RevsDiffBatchCallable;
 import com.cloudant.sync.datastore.callables.SetCurrentCallable;
+import com.cloudant.sync.datastore.callables.UpdateDocumentFromRevisionCallable;
 import com.cloudant.sync.datastore.encryption.KeyProvider;
 import com.cloudant.sync.datastore.encryption.NullKeyProvider;
 import com.cloudant.sync.datastore.migrations.MigrateDatabase100To200;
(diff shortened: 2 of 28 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Changes.java
```diff
@@ -35,7 +35,7 @@
 
     private final List<DocumentRevision> results;
 
-    protected Changes(long lastSequence, List<DocumentRevision> results) {
+    public Changes(long lastSequence, List<DocumentRevision> results) {
```
ricellis: Perhaps add a comment indicating that API users are not expected to construct this and should call `com.cloudant.sync.datastore.Datastore#changes` since it is now publicly visible.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -656,14 +591,17 @@ private DocumentRevision updateDocumentBody(SQLDatabase db, String docId,
...
         if (!preRevision.isCurrent()) {
             throw new ConflictException("Revision to be updated is not current revision.");
         }
 
         new SetCurrentCallable(preRevision.getSequence(), false).call(db);
-        String newRevisionId = this.insertNewWinnerRevision(db, body, preRevision);
+        InsertRevisionCallable insertRevisionCallable = this.insertNewWinnerRevisionAdaptor(body, preRevision);
+        String newRevisionId = insertRevisionCallable.revId;
+        insertRevisionCallable.call(db);
         return new GetDocumentCallable(docId, newRevisionId, this.attachmentsDir, this.attachmentStreamFactory).call(db);
```
ricellis: Should this be a TODO as well to build the new `DocumentRevision` instead of retrieving the whole document again?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1421,77 +982,57 @@ public String extensionDataFolder(String extensionName) {
...
-                    }
 
-                    String revIdKeep = newWinner.getRevision();
-                    if (revIdKeep == null) {
-                        throw new IllegalArgumentException("Winning revision must have a revision" +
-                                " id");
-                    }
+            // before starting the tx, get the 'new winner' and see if we need to prepare its
+            // attachments
+            DocumentRevisionTree docTree = queue.submit(new GetAllRevisionsOfDocumentCallable(docId, attachmentsDir, attachmentStreamFactory)).get();
```
ricellis: Before all parts of this resolution were in a single transaction. Do you think it is safe to do it in parts? I guess the revisions of the document could change between this get and the later resolve and update of attachments.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1421,77 +982,57 @@ public String extensionDataFolder(String extensionName) {
...
+                                AttachmentManager.findNewAttachments(attachments));
+                final List<SavedAttachment> existingAttachments =
+                        AttachmentManager.findExistingAttachments(attachments);
+
+                queue.submit(new UpdateDocumentFromRevisionCallable(newWinner,
+                        preparedNewAttachments,
+                        existingAttachments, attachmentsDir, attachmentStreamFactory)).get();
+            }
+        } catch (AttachmentException e) {
+            // TODO
```
ricellis: Think we should at least add a log message here before merging even if a more robust solution is intended in a later pass.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
```diff
@@ -1505,7 +1046,8 @@ public Object call(SQLDatabase db) throws Exception {
 
     }
 
-    private String insertNewWinnerRevision(SQLDatabase db, DocumentBody newWinner,
+    // TODO should this be a callable itself?
```
ricellis: I don't think so because it doesn't run against the DB, which I think was the pattern for the callables.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentRevision.java
```diff
@@ -312,6 +312,13 @@ public void initialiseSequence(long sequence) {
         }
     }
 
+    /**
+     * @return Whether the body has been modified
```
ricellis: modified since when?

`DocumentRevision` is api_public so we should try to be clear about what the method does or is for or possibly just say it is for internal use.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/callables/GetDocumentsWithIdsCallable.java
```diff
@@ -0,0 +1,57 @@
...
+        this.docIds = docIds;
+        this.attachmentsDir = attachmentsDir;
+        this.attachmentStreamFactory = attachmentStreamFactory;
+    }
+
+    @Override
+    public List<DocumentRevision> call(SQLDatabase db) throws Exception {
+        String sql = String.format("SELECT " + DatastoreImpl.FULL_DOCUMENT_COLS + " FROM revs, docs" +
+                " WHERE docid IN ( %1$s ) AND current = 1 AND docs.doc_id = revs" +
+                ".doc_id " +
```
ricellis: formatting is a bit weird here

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/Changes.java
tomblench: See 0aaed809265ba04360a26bbef1693d5952dd08d4

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
tomblench: See 23ce739c0e7eff73d8378e486bb1a3aa8d92698b - we can sweep these all up into a ticket/gist to use in a later PR

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
tomblench: Thanks for spotting, restored in 78a8c67e4ef1b3f8b4a4c778273f98c0698cbe7f

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
tomblench: Due to changes in 78a8c67 this should no longer be present

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
tomblench: Agreed. Done in 78a8c67e4ef1b3f8b4a4c778273f98c0698cbe7f

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/callables/GetDocumentsWithIdsCallable.java
tomblench: Fixed in 9d866f07e3360b3e1915384a956f08815aed005f

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreImpl.java
ricellis: The first `getAllRevisionsOfDocumentCallable` is still not part of the tx, was that oversight or intentional?

tomblench: Oh I was following the comment but yes in the original version it was inside the transaction. It possibly doesn't make much/any difference because `getAllRevisionsOfDocumentCallable` doesn't modify the database, but I could put it back in the transaction to be consistent with the original code.

ricellis: Yeah considering the scenarios:

1. `rev1`, `rev2` <- `getAllRevisionsOfDocumentCallable`
2. New `rev3` inserted
3. Transaction
    1. Resolve conflicts between `rev1` and `rev2`
4. Conflict exists with `rev3`

vs

1. Transaction
    1. `rev1`, `rev2` <- `getAllRevisionsOfDocumentCallable`
    2. Resolve conflicts between `rev1` and `rev2`
2. New `rev3` inserted
3. Conflict exists with `rev3`

The outcome isn't different even though it *seems* like the second case provides some protection against a new conflict being created during the resolution so you're right there probably isn't anything further to be gained.

Issue #363, opened 2016-09-23
`IndexManager` constructor can return with null `dbQueue` field


tomblench:
See also #280

If an exception is thrown, the caller isn't notified that something went wrong and `dbQueue` is null meaning every subsequent call will fail:

``` java
        try {
            queue = new SQLDatabaseQueue(filename, keyProvider);
            queue.updateSchema(new SchemaOnlyMigration(QueryConstants.getSchemaVersion1()), 1);
            queue.updateSchema(new SchemaOnlyMigration(QueryConstants.getSchemaVersion2()), 2);
            textSearchEnabled = ftsAvailable(queue);
        } catch (Exception e) {
            logger.log(Level.SEVERE, "Failed to open database", e);
            queue = null;
        }
```

ricellis:
@tomblench can this be closed?

PR #366, opened 2016-09-26
IndexManager ctor throws exceptions

tomblench:
Change the constructor to throw exceptions rather than swallowing
them. Add test to check correct behaviour when exceptions thrown.

See #363


tomblench:
Doing this in #384 instead

PR #367, opened 2016-09-27
Refactor access to Crud and Indexing

tomblench:
Datastore -> Crud
DatastoreImpl -> CrudImpl

Create new class Datastore with fields crud and query, through which the
CRUD and Query/Index methods are accessed.


tomblench:
Closing, this work is being done on other PRs

PR #369, opened 2016-09-30
Add Authentication Documentation

### LINKED ISSUE #335: Clarify documentation for different authentication methods
Since we moved to cookie authentication by default, this has caused some issues for customers using proxies (custom configured or Envoy). 

There should be some clear and easy to follow instructions (either in `README` or `doc` directory) showing how to enable basic authentication where needed.

Note also that the `CHANGELOG` never documented this potentially breaking change, probably because we didn't anticipate use of proxies.

See also https://github.com/cloudant-labs/envoy/issues/124

rhyshort:
## What
Added documentation about how sync-android authenticates with the server
by default. Added information on how to customise the authentication
method through the interceptor API.

## Why

By using cookie authentication by default, it may break some users who use a proxy for authentication.

## Issues

Resolves #335

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Add the maven repo and a compile time dependency on the appropriate datastore ar
         </dependencies>
     </project>
     ```
-1. Target is Android and you want datastore encryption. 
+1. Target is Android and you want datastore encryption.
     * Gradle
     ```groovy
     repositories {
@@ -148,7 +148,7 @@ Add the maven repo and a compile time dependency on the appropriate datastore ar
         </dependencies>
     </project>
     ```
-    
+
 You can see a fuller example in the sample application's [build.gradle][sabg].
 
 [sabg]: https://github.com/cloudant/sync-android/blob/master/sample/todo-sync/build.gradle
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -265,6 +265,27 @@ replicator.start();
 
 Read more in [the replication docs](https://github.com/cloudant/sync-android/blob/master/doc/replication.md).
 
+#### Authentication
+
+Sync-android uses session cookies by default to authenticate with the server if
+credentials are provided with in the URL. If you need more fine grain control
```
tomblench: Fine-grained control and Basic Auth are different concepts. How about:
"If you want more fine-grained control over authentication, provide an interceptor to perform the authentication for each request. For example, if you are using a proxy..."

README.md
```diff
@@ -265,6 +265,27 @@ replicator.start();
 
 Read more in [the replication docs](https://github.com/cloudant/sync-android/blob/master/doc/replication.md).
 
+#### Authentication
+
+Sync-android uses session cookies by default to authenticate with the server if
```
tomblench: You should clarify that this happens when the user provides the username and password in the URL

README.md
rhyshort: thinking about this a little bit more, maybe we shouldn't mention proxy here, since we mention it later with for proxy authentication, perhaps we should say

> If you want more fine-grained control over authentication, provide an interceptor to perform the authentication for each request. For example, if you are using middleware such as Envoy....

tomblench: Agreed

rhyshort: done in 6419caf

README.md
```diff
@@ -265,6 +265,27 @@ replicator.start();
 
 Read more in [the replication docs](https://github.com/cloudant/sync-android/blob/master/doc/replication.md).
 
+#### Authentication
```
tomblench: This should be a `###` level heading to match the others

PR #371, opened 2016-10-05
Refactor Datastore and IndexManager - New IndexManager API

tomblench:
Part 2 of _n_ PRs for feature-284-refactor-datastore-indexmanager. See parent issue #284.

The overall theme for this PR is to fix any shortcomings in the public `IndexManager` API. Mostly inspired by @rhyshort - see https://gist.github.com/rhyshort/237fb3d2c00b0c6f3437957635b076c8

- Add `FieldSort` class and use this class instead of a map to request result sorting
- Change `QueryException` to a checked exception and add `throws` clauses for it to IndexManager methods.
- Change Index definition type to be the more specific `Map<String,Map<String,Object>>` and remove casts which this change obviates.
- Update a huge amount of test code which the above breaks

Not done:
- Actual throwing of Exceptions marked on interface methods (instead of currently returning `null` to indicate failure). For a later PR.
- Some extra changes suggested in https://github.com/cloudant/sync-android/issues/305 - should we do them here?

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
@@ -19,11 +19,42 @@ public class FieldSort {
         this.sort = sort;
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        FieldSort fieldSort = (FieldSort) o;
+
+        if (!field.equals(fieldSort.field)) {
+            return false;
+        }
+        return sort == fieldSort.sort;
+
+    }
+
+    @Override
+    public int hashCode() {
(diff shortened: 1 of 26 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+public class FieldSort {
+
+    public final String field;
+    public final Direction sort;
+
+    public FieldSort(String field) {
+        this.field = field;
+        this.sort = Direction.ASCENDING;
```
rhyshort: This should call down into the two args constructor.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
```diff
@@ -0,0 +1,60 @@
+package com.cloudant.sync.query;
```
rhyshort: Missing copyright

PR #372, opened 2016-10-17
284 Refactor IndexManager - "User Friendly API"

tomblench:
Part 3 of _n_ PRs for feature-284-refactor-datastore-indexmanager. See parent issue #284.

The overall theme for this PR is to finish API changes initiated in #371 and ensure that the `IndexManager` has a "user-friendly" API.

- Return `List<Index>` from `listIndexes` so user can interact with a more intuitive, self-documenting type of object.
- Remove over-generic `Map<String, String> IndexSettings` option from `ensureIndexed` and replace it with the an argument for the one current index setting - `tokenize`.
- Throw exceptions instead of returning `null` where possible. Re-throw any `ExecutionException`s as `QueryException`s where needed.
- (Continuing from previous point). Use `Misc.*` pre-conditions to do null checks, argument validity checks at the beginning of public API methods.
- Update test code for new API.
- Do some things in the more correct Java idiomatic way: implement `equals` and `hashCode` for `Index` and remove `compareIndexTypeTo`, remove `Index.getInstance()`, etc.

### DIFF
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexCreatorTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexCreatorTest.java
@@ -43,7 +43,6 @@ public class IndexCreatorTest extends AbstractIndexTestBase {
         assertThat(indexes.isEmpty(), is(true));
     }
 
-    // TODO check that these exceptions are correct
     @Test
     public void preconditionsToCreatingIndexes() throws QueryException {
         // doesn't create an index on null fields
@@ -85,8 +84,8 @@ public class IndexCreatorTest extends AbstractIndexTestBase {
             fieldNames = Arrays.<FieldSort>asList(new FieldSort("age"), new FieldSort("pet"), new
                     FieldSort("age"));
             im.ensureIndexed(fieldNames, "basic");
-            Assert.fail("Expected ensureIndexed to throw a QueryException");
-        } catch (QueryException qe) {
+            Assert.fail("Expected ensureIndexed to throw a IllegalArgumentException");
+        } catch (IllegalArgumentException qe) {
             ;
         }
 
@@ -267,12 +266,11 @@ public class IndexCreatorTest extends AbstractIndexTestBase {
 //        assertThat(indexes.keySet(), containsInAnyOrder("textIndex", "jsonIndex"));
     }
 
-    @Test
+    @Test(expected = QueryException.class)
     public void correctlyLimitsTextIndexesToOne() throws QueryException {
         String indexName = im.ensureIndexed(Arrays.<FieldSort>asList(new FieldSort("name"), new FieldSort("age")), "basic", IndexType.TEXT);
         assertThat(indexName, is("basic"));
         indexName = im.ensureIndexed(Arrays.<FieldSort>asList(new FieldSort("name"), new FieldSort("age")), "anotherIndex", IndexType.TEXT);
-        assertThat(indexName, is(nullValue()));
     }
 
     @Test
@@ -304,8 +302,8 @@ public class IndexCreatorTest extends AbstractIndexTestBase {
         // rejects indexes with $ at start
         try {
             im.ensureIndexed(Arrays.<FieldSort>asList(new FieldSort("$name"), new FieldSort("datatype")), "basic");
-            Assert.fail("Expected ensureIndexed to throw a QueryException");
-        } catch (QueryException qe) {
+            Assert.fail("Expected ensureIndexed to throw a IllegalArgumentException");
+        } catch (IllegalArgumentException qe) {
             ;
         }
 
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexCreatorTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexCreatorTest.java
@@ -19,6 +19,7 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
 
+import org.junit.Assert;
 import org.junit.Test;
 
 import java.util.ArrayList;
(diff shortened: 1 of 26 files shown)


cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexCreatorTest.java
```diff
@@ -31,124 +38,126 @@
...
     public void emptyIndexList() {
-        Map<String, Map<String, Object>> indexes = im.listIndexes();
+        List<Index> indexes = im.listIndexes();
         assertThat(indexes, is(notNullValue()));
         assertThat(indexes.isEmpty(), is(true));
     }
 
     @Test
-    public void preconditionsToCreatingIndexes() {
+    public void preconditionsToCreatingIndexes() throws QueryException {
```
rhyshort: I think this test needs to be broken up into its individual tests using expected class to make sure the exceptions get thrown.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexCreatorTest.java
```diff
@@ -162,123 +171,108 @@ public void createIndexWhenIndexNameExistsIdxDefinitionSame() {
...
-        petField.put("pet", "desc");
-
-        // removes directions from the field specifiers
-        List<String> fields;
-        fields = IndexCreator.removeDirectionsFromFields(Arrays.<FieldSort>asList(new FieldSort("name", FieldSort.Direction.ASCENDING),
-                                                                               new FieldSort("pet", FieldSort.Direction.DESCENDING),
-                                                                               new FieldSort("age")));
-        assertThat(fields, containsInAnyOrder("name", "pet", "age"));
-    }
+    public void createIndexWhereFieldNameContainsDollarSign() throws QueryException {
```
rhyshort: I think this test needs to be spilt into 3, and use the expected class to verify the index isn't created for the on that should fail.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexManagerTest.java
```diff
@@ -30,84 +32,99 @@
...
 
     @Test
-    public void enusureIndexedGeneratesIndexName() {
+    public void enusureIndexedGeneratesIndexName() throws QueryException {
         assertThat(im.ensureIndexed(Arrays.<FieldSort>asList(new FieldSort("name"))), is(notNullValue()));
     }
 
     @Test
-    public void deleteFailOnNoIndexName() {
+    public void deleteFailOnNoIndexName() throws QueryException {
```
rhyshort: This test should also be split.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexTest.java
```diff
@@ -40,105 +40,92 @@ public void setUp() {
...
-        Index index = Index.getInstance(fieldNames, "");
-        assertThat(index, is(nullValue()));
-    }
-
-    @Test
-    public void returnsNullWhenInvalidIndexSettings() {
-        Map<String, String> indexSettings = new HashMap<String, String>();
-        indexSettings.put("foo", "bar");
-        Index index = Index.getInstance(fieldNames, indexName, IndexType.TEXT, indexSettings);
+        Index index = new Index(fieldNames, "");
```
rhyshort: This test needs to renamed, and the assert below needs to be removed too.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexTest.java
```diff
@@ -40,105 +40,92 @@ public void setUp() {
...
     }
 
     @Test
     public void correctlyIgnoresIndexSettings() {
-        Map<String, String> indexSettings = new HashMap<String, String>();
-        indexSettings.put("tokenize", "porter");
         // json indexes do not support index settings.  Index settings will be ignored.
-        Index index = Index.getInstance(fieldNames, indexName, IndexType.JSON, indexSettings);
-        assertThat(index.indexSettings, is(nullValue()));
+        Index index = new Index(fieldNames, indexName, IndexType.JSON, "porter");
```
rhyshort: Looking at the API for this here, it looks a bit awkward, and odd that a constructor will not set a value rather than throw an `IllegalArgumentException`.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexTest.java
```diff
@@ -40,105 +40,92 @@ public void setUp() {
...
-        indexSettings.put("tokenize", "porter");
         // text indexes support the tokenize setting.
-        Index index = Index.getInstance(fieldNames, indexName, IndexType.TEXT, indexSettings);
-        assertThat(index.indexSettings.size(), is(1));
-        assertThat(index.indexSettings.get("tokenize"), is("porter"));
+        Index index = new Index(fieldNames, indexName, IndexType.TEXT, "porter");
+        assertThat(index.tokenize, is("porter"));
     }
-
+    // TODO these should test that .equals() does the right thing
```
rhyshort: Shouldn't these tests be enabled now that `.equals()` is implemented?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
```diff
@@ -1074,19 +1077,23 @@ private void createIndex(String indexName, List<FieldSort> fieldNames) {
     }
 
     @SuppressWarnings("unchecked")
```
rhyshort: I think we can remove this annotation

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexUpdaterTest.java
```diff
@@ -1074,19 +1077,23 @@ private void createIndex(String indexName, List<FieldSort> fieldNames) {
...
-        Map<String, Map<String, Object>> indexes = im.listIndexes();
-        assertThat(indexes, hasKey(indexName));
-
-        Map<String, Object> index = indexes.get(indexName);
-        fields = (List<String>) index.get("fields");
-        assertThat(fields.size(), is(fieldNames.size() + 2));
-        for (FieldSort fieldSort : fieldNames) {
-            assertThat(fields, Matchers.hasItem(fieldSort.field));
+        List<Index> indexes = im.listIndexes();
+        // TODO use custom/better hamcrest matcher rather than interating through
```
rhyshort: Have you created the matcher for this?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/QueryCoveringIndexesTest.java
```diff
@@ -96,7 +97,12 @@ public void setUp() throws Exception{
     @Test
     public void returnsNullForNoQuery() throws Exception {
         setUpBasicQueryData();
-        assertThat(im.find(null), is(nullValue()));
+        try {
+            im.find(null);
+            Assert.fail("find should throw QueryException");
+        } catch (QueryException qe) {
```
rhyshort: Use the expected class attribute for this.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexTest.java
tomblench: The existing tests seem a bit pointless but we can replace them with some equivalents which exercise `equals`.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexManagerTest.java
tomblench: This test can't be split as it tests a certain sequence of events - create index, delete non-existent index, check that first index is still there.

rhyshort: ok

PR #379, opened 2016-10-20
Moved FTS check to SQLDatabaseFactory

### LINKED ISSUE #378: Remove nested transaction from FTS check
The refactoring in branch **feature-284-refactor-datastore-indexmanager** has left a single place where there is a nested transaction (which also happens not to be completed in a finally block): `com.cloudant.sync.query.IndexManagerImpl#ftsAvailable`

This method in itself probably only needs to be run once given that it is checking for the availability of full text search in the compiled SQLite which is something that isn't going to change during running.

We should clean up the transaction and if possible make this check happen only once.

ricellis:
*What*

Moved FTS check to `SQLDatabaseFactory`, removing nested transaction

*How*

* Moved FTS check to `SQLDatabaseFactory`.
* Changed to perform FTS check only once in static initializer.
* Removed nested transaction from FTS check.
* Simplified FTS check to always rollback the table creation instead of
also dropping the table.
* Used an in-memory database for FTS check table creation.
* Updated subclasses of `SQLDatabase` to allow null paths for creation of
in-memory databases.
* Added internal create method to `SQLDatabaseFactory` to allow creation
of in-memory database.
* Updated copyright statements.

*Testing*

* Updated `com.cloudant.sync.query.IndexManagerTest#validateTextSearchIsAvailable`
to assert the new field.
* Refactored query tests for the feature branch to use the `IndexManager` supplied by the `CloudantSync` object instead of constructing duplicates. Using duplicates was leading to errors during the migration phase of the `IndexManager` init that were swallowed by tests by severely impacting the test speed.

*Issues*

Fixes #378

### DIFF
--- a/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLite.java
+++ b/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLite.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013 Cloudant, Inc. All rights reserved.
+ * Copyright © 2013, 2016 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
@@ -33,7 +33,12 @@ public class AndroidSQLite extends SQLDatabase {
 
 
     public static AndroidSQLite createAndroidSQLite(String path) {
-        SQLiteDatabase db = SQLiteDatabase.openDatabase(path, null, SQLiteDatabase.CREATE_IF_NECESSARY);
+        SQLiteDatabase db;
+        if (path != null) {
+            db = SQLiteDatabase.openDatabase(path, null, SQLiteDatabase.CREATE_IF_NECESSARY);
+        } else {
+            db = SQLiteDatabase.create(null);
+        }
         return new AndroidSQLite(db);
     }
 
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
@@ -1,14 +1,16 @@
-//  Copyright (c) 2014 Cloudant. All rights reserved.
-//
-//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
-//  except in compliance with the License. You may obtain a copy of the License at
-//
-//    http://www.apache.org/licenses/LICENSE-2.0
-//
-//  Unless required by applicable law or agreed to in writing, software distributed under the
-//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
-//  either express or implied. See the License for the specific language governing permissions
-//  and limitations under the License.
+/*
+ * Copyright © 2014, 2016 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
 
 //
 // The metadata for an index is represented in the database table as follows:
@@ -70,12 +72,12 @@ import java.util.regex.Pattern;
  *  - execute queries
  *  - update indexes (usually done automatically)
  *
+ *
  *  @api_public
  */
 public class IndexManagerImpl implements IndexManager {
 
     private static final String INDEX_TABLE_PREFIX = "_t_cloudant_sync_query_index_";
-    private static final String FTS_CHECK_TABLE_NAME = "_t_cloudant_sync_query_fts_check";
     public static final String INDEX_METADATA_TABLE_NAME = "_t_cloudant_sync_query_metadata";
 
     private static final String EXTENSION_NAME = "com.cloudant.sync.query";
@@ -87,8 +89,6 @@ public class IndexManagerImpl implements IndexManager {
     private final Pattern validFieldName;
 
     private final SQLDatabaseQueue dbQueue;
-
-    private boolean textSearchEnabled;
 
     /**
      *  Constructs a new IndexManager which indexes documents in 'datastore'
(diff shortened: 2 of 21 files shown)


cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLite.java
```diff
@@ -33,7 +33,9 @@
 
 
     public static AndroidSQLite createAndroidSQLite(String path) {
-        SQLiteDatabase db = SQLiteDatabase.openDatabase(path, null, SQLiteDatabase.CREATE_IF_NECESSARY);
+        SQLiteDatabase db = (path != null)
+                ? SQLiteDatabase.openDatabase(path, null, SQLiteDatabase.CREATE_IF_NECESSARY)
```
emlaver: Since `if/else` blocks are used in the rest of the code, wouldn't it be better to use that style to keep things consistent?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
```diff
@@ -1,38 +1,41 @@
...
+/**
+ * The metadata for an index is represented in the database table as follows:
+ *
+ * index_name|index_type|field_name|last_sequence
+ * -----------------------------------------------------------
+ * name|json|_id|0
+ * name|json|_rev|0
+ * name|json|firstName|0
+ * name|json|lastName|0
+ * age|json|age|0
```
emlaver: Should we keep the spacing and format in the original commit for this table?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
```diff
@@ -1,38 +1,41 @@
...
+ *
+ * The index itself is a single table,with a column for docId and each of the indexed fields:
+ *
+ * _id|_rev|firstName|lastName
+ * --------------------------------------------------------
+ * miker|1-blah|Mike|Rhodes
+ * johna|3-blob|John|Appleseed
+ * joeb|2-blip|Joe|Bloggs
```
emlaver: Keep table spacing and format like https://github.com/cloudant/sync-android/pull/379/files#diff-b7df9e23ebfd740a214a3db0f44ee84eL25 ?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
ricellis: Reverted in 2c68723884df10ac118f5866fd5b88bfea5d35f1.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
ricellis: Reverted in 2c68723884df10ac118f5866fd5b88bfea5d35f1.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLite.java
ricellis: I tend to prefer the ternary operator for cases like this where one of two values are being assigned, but I don't mind much the outcome is the same.
Changed to if/else in 07aedeee952055d401a4bea457ed65f4371af37c along with the complementary assignment in `SQLiteWrapper`.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
```diff
@@ -48,12 +80,25 @@
      * @throws SQLException if the database cannot be opened.
      */
     public static SQLDatabase createSQLDatabase(String dbFilename, KeyProvider provider) throws IOException, SQLException {
+        makeSureFileExists(dbFilename);
+        return internalCreateSQLDatabase(dbFilename, provider);
+    }
+
+    /**
+     * Internal method for creating a SQLDatabase that allows a null filename to create an in-memory
+     * database which can be useful for performing checks, but creating in-memory databases is not
```
brynh: Although our current usage is fine and it's private so it's only internally we'd be affected, this could either do with a comment stating that if `dbFilename` is null, `provider` must be a `NullKeyProvider`, or we need to update the `createAndroidSQLite` method in `AndroidSQLCipherSQLite` to handle creation of an encrypted in-memory database by calling `SQLiteDatabase.create` instead of `SQLiteDatabase.openOrCreateDatabase`.  As it stands, I think we'd get an `IllegalArgumentException` if it were called with a null `dbFilename` and a `provider` with a non-null key.

cloudant-sync-datastore-javase/src/main/java/com/cloudant/sync/sqlite/sqlite4java/SQLiteWrapper.java
```diff
@@ -84,7 +84,12 @@ SQLiteConnection getConnection() {
 
     SQLiteConnection createNewConnection() {
         try {
-            SQLiteConnection conn = new SQLiteConnection(new File(this.databaseFilePath));
+            SQLiteConnection conn;
+            if (this.databaseFilePath != null) {
+                conn = new SQLiteConnection(new File(this.databaseFilePath));
+            } else {
+                conn = new SQLiteConnection(null);
```
brynh: We needn't pass in null here, we could call the no args constructor.

cloudant-sync-datastore-javase/build.gradle
```diff
@@ -1,7 +1,7 @@
 dependencies {
     compile project(':cloudant-sync-datastore-core')
 
-    compile group: 'com.almworks.sqlite4java', name: 'sqlite4java', version:'1.0.392'
+    compile group: 'com.almworks.sqlite4java', name: 'sqlite4java', version: '1.0.392'
```
ricellis: corrected the others as well in 92ea3c919a6ffc7e8ec67f0df638751001ab6709

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
ricellis: Added comment in b19786cd545eac2208a498ee5df58ff9816414c6. I think that ought to be sufficient, but it it isn't we could always add a check to `internalCreateSQLDatabase` like:
```java
if (dbFilename == null && !(provider instanceof NullKeyProvider)) {
     throw new IllegalArgumentException("In-memory datastore must use NullKeyProvider.");
 }
```

cloudant-sync-datastore-javase/src/main/java/com/cloudant/sync/sqlite/sqlite4java/SQLiteWrapper.java
ricellis: Done in 1f6a3500396eb9d452c9b45c28aedd403984cf96

Issue #294, opened 2016-06-23
Create entire storage path if it does not already exist


ricellis:
Initializing a `DatastoreManager` currently creates only a single directory of the path by calling `directoryPath.mkdir();` if it doesn't exist.
It would be nicer to create any necessary parent directories too by calling `mkdirs()`.
We should document the behaviour and the exceptions that could be thrown if it is not possible to create the paths.

rhyshort:
Done in cbe43c0

PR #380, opened 2016-10-21
Remove DatastoreManager

tomblench:
Part 4 of _n_ PRs for feature-284-refactor-datastore-indexmanager. See parent issue #284.

The overall theme for this PR is to remove the DatastoreManager and ensure that all of the key classes the user interacts with are named sensibly whilst we have the opportunity to introduce the breaking changes that this causes.

The DatastoreManager object introduces needless complexity for users of the API. Instead add static methods for getting/closing/deleting instances of DocumentStore on the DocumentStore object itself.

Changed behaviour or removed arbitrary restrictions:
- Removed datastore name. The user just passes in a path and that path is used to store db.sync etc
- Due to above, there is a subtle change in behaviour around directory creation. If the user calls getInstance with "/foo/bar/baz", if "/foo/bar/" exists, "baz" will be created, then db.sync under that
- Also due to above we need to explain how to construct a new path equivalent to the datastore location under the old API.
- Removal of restriction on database names. Because it's just a plain path now, and it was a pointless restriction.
- Some eventbus changes (see below)

Some tests currently failing, including `EndToEndEncryptionTest`, which we can clean up later.

Open questions:
- Is it correct to have a static `EventBus` on `DocumentStore`
- What events should be posted on the `DocumentStore` eventbus vs `Database` instance eventbus. An example here would be the DatastoreClosed event.
- What else should be renamed, should events be renamed?
- Is there still a need for a method to list `DocumentStore`s? I have provided `listInstances`


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+
+public class DocumentStore {
+
+    public final Database database;
+    public final IndexManager query;
+    protected final String databaseName; // only used for events
+    private final File location; // needed for close
+
+    /* This map should only be accessed inside synchronized(openDatastores) blocks */
+    private static final Map<File, DocumentStore> documentStores = new HashMap<File, DocumentStore>();
```
rhyshort: Shouldn't this be an `ConcurrentMap`?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+    public void close() {
+        synchronized (documentStores) {
+            DocumentStore ds = documentStores.get(location);
+            if (ds != null) {
+                System.out.println("removing "+location);
+                ((DatabaseImpl)database).close();
+                ((IndexManagerImpl)query).close();
+                documentStores.remove(location);
+            }
+            // TODO warn if not in map?
```
rhyshort: If this isn't in the map, it has already been closed, I think we should error out somehow to give a hint to the developer that they should only close this once.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+                throw new IOException(msg);
+            } else {
+                FileUtils.deleteDirectory(location);
+                eventBus.post(new DatabaseDeleted(databaseName));
+            }
+        }
+    }
+
+    public static List<File> listInstances() {
+        return new ArrayList<File>(documentStores.keySet());
```
rhyshort: I don't think we should return a list of files for this API. We should either return a List of `DocumentStore`s or a change the name.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+        Misc.checkArgument(location.isDirectory(), "Input path is not a valid directory");
+        Misc.checkArgument(location.canWrite(), "Datastore directory is not writable");
+        return created;
+    }
+
+    public void close() {
+        synchronized (documentStores) {
+            DocumentStore ds = documentStores.get(location);
+            if (ds != null) {
+                System.out.println("removing "+location);
```
rhyshort: This should either be a formal log or removed.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+        eventBus.post(new DatabaseClosed(databaseName));
+    }
+
+    public void delete() throws IOException {
+        synchronized (documentStores) {
+            DocumentStore ds = documentStores.remove(location);
+            if (ds != null) {
+                ((DatabaseImpl)database).close();
+                ((IndexManagerImpl)query).close();
```
rhyshort: Again, suppress the unchecked warnings here.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+            DocumentStore ds = documentStores.get(location);
+            if (ds != null) {
+                System.out.println("removing "+location);
+                ((DatabaseImpl)database).close();
+                ((IndexManagerImpl)query).close();
+                documentStores.remove(location);
+            }
+            // TODO warn if not in map?
+        }
+        eventBus.post(new DatabaseClosed(databaseName));
```
rhyshort: We post this in the impl too, should we really be generating two events for the same thing?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+            throw new DatastoreNotCreatedException("Database not found: " + location, e);
+        } catch (SQLException e) {
+            throw new DatastoreNotCreatedException("Database not initialized correctly: " + location, e);
+        } catch (DatastoreException e) {
+            throw new DatastoreNotCreatedException("Datastore not initialized correctly: " + location, e);
+        }
+    }
+
+    private static boolean checkPathAndCreateIfNeeded(File location) {
+        //logger.fine("Datastore path: " + directoryPath);
```
rhyshort: Remove or reinstate. I think I'd prefer to keep it.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatabaseImpl.java
```diff
@@ -166,31 +166,28 @@
...
      */
     private final AttachmentStreamFactory attachmentStreamFactory;
-
+/*
     protected DatabaseImpl(String dir, String name) throws SQLException, IOException,
             DatastoreException {
         this(dir, name, new NullKeyProvider());
     }
-
+*/
```
rhyshort: Let's remove this constructor

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatabaseImpl.java
```diff
@@ -166,31 +166,28 @@
...
     /**
      * Constructor for single thread SQLCipher-based datastore.
      * @param dir The directory where the datastore will be created
-     * @param name The user-defined name of the datastore
      * @param provider The key provider object that contains the user-defined SQLCipher key
      * @throws SQLException
      * @throws IOException
      */
-    protected DatabaseImpl(String dir, String name, KeyProvider provider) throws SQLException,
+    protected DatabaseImpl(File dir, KeyProvider provider) throws SQLException,
```
rhyshort: I wonder if we should lower to scope to `package`

rhyshort:
My two cents on a couple of the questions, I think we need to rename to events to be consistent, but I can see that's another change someone going to 2.0 is going to need to make so we might want to maintain compatibility for events maybe? 

I think having the static event bus is the right thing, so we can post the deleted events.

I'm not sure if there is a need to list instances, but it might be useful to users, so something to mull over, or just include it since its not a big API to maintain. 

With the file behaviour change, I wonder if we should just go all out and make all the directories rather than the final one. Not entirely sure if that would be unexpected or not.

ricellis:
>Is it correct to have a static EventBus on DocumentStore

Yes. The alternatives are scrapping `DatastoreOpened`/`Closed` events or unify all events on a single global bus. Having a global `EventBus` would require filtering capabilities beyond just the event type (for example to enable subscribing to `DocumentModified` events from a single `DocumentStore`). I don't think the second approach is viable for us right now, but I do think we should rename the `EventBus` instances so it is clear what they are for and possibly add checks for the types of events so that subscribers looking for an event that will never be issued would be rejected (i.e. a whitelist of event classes for given `EventBus` instances). Although I think that would be a different PR.

>What events should be posted on the DocumentStore eventbus vs Database instance eventbus. An example here would be the DatastoreClosed event.

The` DocumentStore` static `EventBus` should have:
* `DatabaseClosed`
* `DatabaseCreated`
* `DatabaseDeleted`
* `DatabaseModified`
* `DatabaseOpened`

The database (`DocumentStore` instance) `EventBus` should have:
* `DocumentCreated`
* `DocumentDeleted`
* `DocumentModified`
* `DocumentUpdated`

>What else should be renamed, should events be renamed?

The `Database*` events already didn't match the `Datastore` naming scheme, but this is a good opportunity to make them match so it is clear which class the events are associated with.

>Is there still a need for a method to list DocumentStores? I have provided listInstances

Listing the known `DocumentStore`s necessitated knowing some top level path. If we're removing that concept we can't effectively list the stores any more, unless we take an argument of a path. Your replacement lists the open `DocumentStore`s and I think that is fine, but the method name needs to reflect that, although I'm struggling to think of a good use case for it.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * Created by tomblench on 27/09/2016.
+ */
```
emlaver: Class documentation needed.

ricellis:
Regarding the directory creation behaviour I think we should just go ahead and do #294 
either here or elsewhere in the 2.0 milestone.

rhyshort:
@ricellis agreed.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+                ((IndexManagerImpl)query).close();
+                documentStores.remove(location);
+            }
+            // TODO warn if not in map?
+        }
+        eventBus.post(new DatabaseClosed(databaseName));
+    }
+
+    public void delete() throws IOException {
+        synchronized (documentStores) {
```
ricellis: I think this method should just call `close()` then do the on disk deletion, rather than duplicating the `close()` code.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        Misc.checkArgument(location.isDirectory(), "Input path is not a valid directory");
+        Misc.checkArgument(location.canWrite(), "Datastore directory is not writable");
+        return created;
+    }
+
+    public void close() {
+        synchronized (documentStores) {
+            DocumentStore ds = documentStores.get(location);
```
ricellis: I think this could `.remove(location)` immediately and then close it down rather than calling `get` then `remove`.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatabaseImpl.java
tomblench: Done in 248a5a8

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatabaseImpl.java
tomblench: Done in 9df5d06

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
tomblench: Done in 4f4fbcc

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
tomblench: Done in 4f4fbcc

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
tomblench: Done in 4f4fbcc, with unit test in 629a26a

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+        synchronized (documentStores) {
+            DocumentStore ds = documentStores.remove(location);
+            if (ds != null) {
+                ((DatabaseImpl)database).close();
+                ((IndexManagerImpl)query).close();
+            }
+            if (!location.exists()) {
+                String msg = String.format(
+                        "Datastore %s doesn't exist on disk", location
+                );
```
tomblench: Done in 4f4fbcc

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
tomblench: I decided to remove it because it isn't very useful. See 606bb9f

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
tomblench: Removed in 4f4fbcc

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+        return created;
+    }
+
+    public void close() {
+        synchronized (documentStores) {
+            DocumentStore ds = documentStores.get(location);
+            if (ds != null) {
+                System.out.println("removing "+location);
+                ((DatabaseImpl)database).close();
+                ((IndexManagerImpl)query).close();
```
tomblench: Done in b05d5ef

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
tomblench: Done in b05d5ef

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
tomblench: Done in 	b9337d6

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
tomblench: @emlaver done in 0f71e9d

Some methods still need documenting, we'll cover these in a documentation pass on a later PR.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,138 @@
...
+    private static final Map<File, DocumentStore> documentStores = new HashMap<File, DocumentStore>();
+
+    private static EventBus eventBus = new EventBus();
+
+    private DocumentStore(File location, KeyProvider keyProvider) throws DatastoreException, IOException, SQLException {
+        this.location = location;
+        this.databaseName = location.getName();
+        this.database = new DatabaseImpl(location, keyProvider);
+        // this will be used to observe purge() events from database and forward them to indexmanager
+        this.database.getEventBus().register(this);
```
tomblench: @rhyshort this will be used when we implement purge which isn't part of this feature branch, we can take it out now if you don't think it's relevant here

tomblench:
@rhyshort @emlaver @emlaver implemented changes now, please notify me if you have any more feedback

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
rhyshort: Since purge isn't going to be part of 2.0, i'd prefer it if we dropped these two lines.

tomblench: Done in 025c6a4

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatabaseImpl.java
```diff
@@ -217,8 +209,7 @@ protected DatabaseImpl(String dir, String name, KeyProvider provider) throws SQL
 
     @Override
     public String getDatastoreName() {
-        Misc.checkState(this.isOpen(), "Database is closed");
-        return this.datastoreName;
+        return this.datastoreDir.getName();
```
ricellis: I wonder if we should return something longer here. Per the javadoc
>This is just the last name in the pathname's name sequence.

Now that we don't have the `DatastoreManager` context we have lost the filesystem's enforced uniqueness of a an individual datastore name (folder) within that context.

I wonder if this new name method should return the whole path?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,192 @@
...
+                DocumentStore ds = documentStores.get(location);
+                boolean created = false;
+                if (ds == null) {
+                    created = checkPathAndCreateIfNeeded(location);
+                    ds = new DocumentStore(location, provider);
+                    documentStores.put(location, ds);
+                    eventBus.post(new DatabaseOpened(ds.databaseName));
+                }
+                if (created) {
+                    eventBus.post(new DatabaseCreated(ds.databaseName));
```
ricellis: I think it's weird that this event would be posted after the `DatabaseOpened` in the creation case. Should probably move up.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,192 @@
...
+     * a non-null key will result in an exception.
+     * </p>
+     * @param location The location on the file system where the underlying files should be stored.
+     *                 Must be a directory.
+     * @param provider KeyProvider object. Use a {@link NullKeyProvider} if the database shouldn't
+     *                 be encrypted.
+     * @return An existing or newly created store.
+     * @throws DatastoreNotCreatedException if the database cannot be opened
+     */
+    public static DocumentStore getInstance(File location, KeyProvider provider) throws DatastoreNotCreatedException {
```
ricellis: DatastoreNotCreatedException does not seem like the correct type for cases where it exists, but can't be opened for some reason.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,192 @@
...
+                ((IndexManagerImpl)query).close();
+            }
+            else {
+                throw new IllegalStateException("DocumentStore "+location+" already closed");
+            }
+        }
+        eventBus.post(new DatabaseClosed(databaseName));
+    }
+
+    public void delete() throws IOException {
```
ricellis: It seems a bit inconsistent that `delete()` will throw an `IOException` for some problem (e.g. file locked), but the failure to create or open for the same reason would throw a specific `Datastore*` exception.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -0,0 +1,192 @@
...
+        } catch (Exception e) {
+            // caught exception could be something benign like datastore already closed, so just log
+            logger.log(Level.WARNING, "Caught exception whilst closing DocumentStore in delete()", e);
+        }
+        if (!location.exists()) {
+            String msg = String.format("Datastore %s doesn't exist on disk", location);
+            logger.warning(msg);
+            throw new IOException(msg);
+        } else {
+            FileUtils.deleteDirectory(location);
```
ricellis: If the checked exception type changes then we'll need to catch the `IOException` from a failed deletion and throw the new type here as well.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatabaseImpl.java
```diff
@@ -167,30 +167,22 @@
...
     /**
      * Constructor for single thread SQLCipher-based datastore.
      * @param dir The directory where the datastore will be created
-     * @param name The user-defined name of the datastore
      * @param provider The key provider object that contains the user-defined SQLCipher key
      * @throws SQLException
      * @throws IOException
      */
-    protected DatabaseImpl(String dir, String name, KeyProvider provider) throws SQLException,
+    DatabaseImpl(File dir, KeyProvider provider) throws SQLException,
```
ricellis: I think this should be private, or it can probably stay package if it moves into an internal package. I realise a bunch of tests are calling `new DatabaseImpl` but really they should be migrated where possible (which I think should be everywhere) to use the `getInstance().database` API (and I guess casting to `DatabaseImpl` in some cases).

PR #384, opened 2016-11-03
284 refactor datastore indexmanager 5 fix exception handling

### LINKED ISSUE #294: Create entire storage path if it does not already exist

tomblench:
Part 5 of _n_ PRs for feature-284-refactor-datastore-indexmanager. See parent issue #284.

This PR is for:
- Fix exception handling in IndexManager (addresses #363)
- Fix exception handling in DocumentStore by introducing NotDeleted/NotClosed events
- Fix other minor shortcomings mentioned by Rich in https://github.com/cloudant/sync-android/pull/380 which weren't addressed before merge
- Fix issues with opening new/existing DocumentStores, Databases, and underling SQLite databases. Fixes #294


cloudant-sync-datastore-android-encryption/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLCipherSQLite.java
```diff
@@ -47,7 +49,7 @@
      * @param provider Provider object that contains the key to encrypt the SQLCipher database
      * @return
      */
-    public static AndroidSQLCipherSQLite createAndroidSQLite(String path, KeyProvider provider) {
+    public static AndroidSQLCipherSQLite openOrCreate(File path, KeyProvider provider) {
```
rhyshort: This method implies, that if the database exists, we'll open it, if it doesn't we'll just create it. I think we need a better name, not sure what though.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatabaseImpl.java
```diff
@@ -209,7 +209,7 @@
 
     @Override
     public String getDatastoreName() {
-        return this.datastoreDir.getName();
+        return this.datastoreDir.toString();
```
rhyshort: Since we are returning the path now, I don't think it is right to have this method called `getName`.  I think `getPath` might be more appropriate.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreNotDeletedException.java
```diff
@@ -0,0 +1,23 @@
+package com.cloudant.sync.datastore;
```
rhyshort: Missing copyright

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreNotOpenedException.java
```diff
@@ -0,0 +1,23 @@
+package com.cloudant.sync.datastore;
```
rhyshort: missing copyright

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -170,9 +167,15 @@ public void delete() throws IOException {
         if (!location.exists()) {
             String msg = String.format("Datastore %s doesn't exist on disk", location);
             logger.warning(msg);
-            throw new IOException(msg);
+            throw new DatastoreNotDeletedException(msg);
```
rhyshort: I feel like a `FileNotFound` would be a better exception, although there is a part of me that thinks we should crash because the state isn't right. Calling `delete` twice seems like a programmer error than something that we should be recovering from. Thoughts?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
```diff
@@ -94,26 +95,17 @@
...
+    public IndexManagerImpl(Database database) throws IOException, SQLException {
         this.database = database;
         validFieldName = Pattern.compile(INDEX_FIELD_NAME_PATTERN);
 
-        final String filename = ((DatabaseImpl) database).extensionDataFolder(EXTENSION_NAME) + File.separator
-                                                                              + "indexes.sqlite";
-        final KeyProvider keyProvider = ((DatabaseImpl) database).getKeyProvider();
-
-        SQLDatabaseQueue queue = null;
+        final File file = new File(((DatabaseImpl) database).extensionDataFolder(EXTENSION_NAME), "indexes.sqlite");
```
rhyshort: Thinking about this some more, I don't think we should couple this to the implementation of the `Database` perhaps we should have a package level interface which exposes the extensions file path, and that is what gets  passed into this method.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
```diff
@@ -94,26 +95,17 @@
...
 
-        try {
-            queue = new SQLDatabaseQueue(filename, keyProvider);
-            queue.updateSchema(new SchemaOnlyMigration(QueryConstants.getSchemaVersion1()), 1);
-            queue.updateSchema(new SchemaOnlyMigration(QueryConstants.getSchemaVersion2()), 2);
-        } catch (Exception e) {
-            logger.log(Level.SEVERE, "Failed to open database", e);
-            queue = null;
-        }
+        final KeyProvider keyProvider = ((DatabaseImpl) database).getKeyProvider();
```
rhyshort: I think key provider should also be added to the internal interface.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
```diff
@@ -70,106 +72,69 @@ private static boolean isFtsAvailable() {
...
-     * @throws IOException if the file does not exists, and also
-     *         can not be created
+     * @throws IOException if the file cannot be opened, or the database files or its intermediate
+     *                     directories cannot be created.
      * @throws SQLException if the database cannot be opened.
      */
-    public static SQLDatabase createSQLDatabase(String dbFilename, KeyProvider provider) throws IOException, SQLException {
-        makeSureFileExists(dbFilename);
-        return internalCreateSQLDatabase(dbFilename, provider);
+    public static SQLDatabase openOrCreateSQLDatabase(File dbFile, KeyProvider provider) throws IOException, SQLException {
```
rhyshort: Again, this really sounds like its only going to open the database if it already exists, `openSQLDatabase` feels alot better to my mind.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
```diff
@@ -48,30 +48,30 @@
...
      * @throws IOException If a problem occurs creating the database
      * @throws SQLException If the database cannot be opened.
      */
-    public SQLDatabaseQueue(final String filename, KeyProvider provider) throws IOException, SQLException {
+    public SQLDatabaseQueue(final File file, KeyProvider provider) throws IOException, SQLException {
         queue = Executors.newSingleThreadExecutor(new ThreadFactory() {
             @Override
             public Thread newThread(Runnable r) {
-                return new Thread(r, "SQLDatabaseQueue - "+ filename);
+                return new Thread(r, "SQLDatabaseQueue - "+ file);
```
rhyshort: do we really want the full path to be printed for the thread name here?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatabaseManagerTest.java
```diff
@@ -47,11 +44,11 @@ public void tearDown() {
...
     }
 
-    @Test(expected = IllegalArgumentException.class)
-    public void nonWritableDatastoreManagerPathThrows() throws DatastoreNotCreatedException {
+    @Test(expected = DatastoreNotOpenedException.class)
+    public void nonWritableDatastoreManagerPathThrows() throws DatastoreNotOpenedException, IOException {
         File f = new File(TEST_PATH, "c_root_test");
         try {
-            f.mkdir();
+            FileUtils.forceMkdir(f);
```
rhyshort: Why are we forcing the creation of the directory? Surely it should be writeable?

tomblench: "Makes a directory, including any necessary but nonexistent parent directories" (https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html#forceMkdir(java.io.File)) seems like the right thing, no? This is necessary for #294.

cloudant-sync-datastore-android-encryption/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLCipherSQLite.java
tomblench: The name is an attempt to be consistent with the call to the underlying SQLite java interface - `SQLiteDatabase.openOrCreateDatabase`, and `SQLiteConnection.open(true)` where the boolean flag is called `allowCreate`. I'm happy to use a different name, but it should be the same across our 3 classes!

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseQueue.java
tomblench: The thinking comes from a previous comment by Rich. Previously the datastore names were unique per manager, but now we don't have that restriction. So there is a chance that users could have multiple datastores (documentstores?) open, eg `/a/path/db.sync` and `/b/path/db.sync`. If we follow our existing convention of using the lowest level directory as the name, then the user won't be able to tell the difference.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatabaseImpl.java
tomblench: Sounds fine, let's just return the `File` object in `getPath()`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
tomblench: This is to be consistent with `DatastoreNotOpenedException`, but I agree you have a point on the checked/unchecked argument.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
tomblench: It's probably more sensible to move this method into the `DocumentStore` which manages this - so `DocumentStore` should have an `extensionDataFolder` method, what do you think?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
tomblench: As above

tomblench:
@rhyshort do you think this is sufficient to prevent leaking connections? d3b1d5a

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatabaseManagerTest.java
ricellis: [`java.io.File.mkdirs()`](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#mkdirs%28%29)
?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
```diff
@@ -65,10 +65,30 @@
...
-        this.location = location;
-        this.databaseName = location.toString();
-        this.database = new DatabaseImpl(location, keyProvider);
-        this.query = new IndexManagerImpl(database);
+        try {
+            this.location = location;
+            this.databaseName = location.toString();
+            this.database = new DatabaseImpl(location, keyProvider);
+            this.query = new IndexManagerImpl(database);
+        } catch (DatastoreException e) {
```
rhyshort: Can we just catch `Exception` so we don't repeat the same code? (I wish we could use multicatch for this).

rhyshort:
@tomblench It looks like it will be sufficient, although I'd rather have one catch block

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DocumentStore.java
tomblench: As discussed, can't be done without changing the constructor signature to `throws Exception`

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/DatabaseManagerTest.java
tomblench: OK, changed to use `mkdirs()` in 9357ed5

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatabaseImpl.java
tomblench: Done in d322b41

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
tomblench: See how I fixed this in b40b64a and tell me if it makes sense

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
tomblench: See 6939c89

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreNotDeletedException.java
tomblench: Done in 8c4bca4

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/datastore/DatastoreNotOpenedException.java
tomblench: Done in 8c4bca4

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/sqlite/SQLDatabaseFactory.java
```diff
@@ -70,106 +72,69 @@ private static boolean isFtsAvailable() {
...
+        Misc.checkNotNull(dbFile, "dbFile");
+        System.out.println("*** creating "+dbFile.getParentFile());
```
brynh: I presume this line is left over from debugging and can be removed now.

cloudant-sync-datastore-android-encryption/src/main/java/com/cloudant/sync/sqlite/android/AndroidSQLCipherSQLite.java
rhyshort: I agree they should be same, and I'm more inclined to go with `open` with the documented behaviour that it creates the file if it isn't there.

tomblench:
@rhyshort are we done now?

PR #385, opened 2016-11-07
Handle missing revisions correctly.

rhyshort:
## ~What~

~Stop the replicator crashing when missing revisions are encountered.~

## ~Why~
~Missing revisions can occur for various reasons, one of them is that
revs limit has been reached on the server and the server is actively discarding
old revisions.~

## ~How~

~Insert stub revisions when missing revisions are encountered to keep the tree as expected.~

## ~Testing~

~One new test added, which mocks the changes feed with a revision that doesn't exist.~

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchClient.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchClient.java
@@ -3,6 +3,8 @@
  *
  * Copyright (C) 2011 Ahmed Yehia (ahmed.yehia.m@gmail.com)
  *
+ * Copyright © 2016 IBM Corp. All rights reserved.
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
@@ -409,6 +411,9 @@ public class CouchClient  {
 
         Map<String, Object> options = new HashMap<String, Object>();
         options.put("revs", true);
+        // by adding latest we should never receive a "missing" response from the server. A descendant
+        // of the revision requested will be returned even if it has been deleted.
+        options.put("latest", true);
         // only pull attachments inline if we're configured to
         if (pullAttachmentsInline) {
             options.put("attachments", true);
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/CouchClientWrapper.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/CouchClientWrapper.java
@@ -1,6 +1,8 @@
 /**
  * Copyright (c) 2013 Cloudant, Inc. All rights reserved.
  *
+ * Copyright © 2016 IBM Corp. All rights reserved.
+ *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
  *
@@ -48,7 +50,7 @@ public class CouchClientWrapper implements CouchDB {
     private final static String LOG_TAG = "CouchClientWrapper";
     private static final Logger logger = Logger.getLogger(CouchClientWrapper.class.getCanonicalName());
 
-    final CouchClient couchClient;
+    CouchClient couchClient;
 
     public CouchClientWrapper(CouchClient client) {
         Misc.checkNotNull(client, "Couch client");
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/MissingRevsReplicationTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/MissingRevsReplicationTest.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright © 2016 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package com.cloudant.sync.replication;
+
+
+import static org.mockito.Matchers.anyInt;
+import static org.mockito.Matchers.anyMapOf;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.spy;
+
+import com.cloudant.mazha.ChangesResult;
+import com.cloudant.mazha.CouchClient;
+import com.cloudant.sync.datastore.DocumentRevision;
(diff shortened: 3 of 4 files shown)


cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/PullStrategyTest.java
```diff
@@ -76,6 +78,56 @@ private void pull(PullStrategy replicator, int expectedDocs) throws Exception {
...
+        // inject our mocked changesresult via a 'spy'
+        // ChangesResultAnswer returns a missing revision each time.
+        customReplicator.sourceDb = spy(customReplicator.sourceDb);
+        doAnswer(new ChangesResultAnswer(changesResult1, changesResult2))
+                .when(customReplicator.sourceDb).
+                changes((PullFilter)
+                        anyObject(), anyObject(), anyInt());
+        // do the actual replication
+        customReplicator.run();
+
```
tomblench: So this test is useful because it checks that the replicator doesn't explode with an exception. But we should have a bunch of asserts at the end to make sure we got all of the expected documents, that the sequence number is correct, etc.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/CouchClientWrapper.java
```diff
@@ -165,8 +154,16 @@ public ChangesResult changes(PullFilter filter, Object lastSequence, int limit)
             if(openRev instanceof OkOpenRevision) {
                 documentRevs.add(((OkOpenRevision) openRev).getDocumentRevs());
             } else {
-                throw new RuntimeException("Missing open revision for document:" + documentId
-                           + ", revisions: " + Arrays.asList(revisionIds));
+                MissingOpenRevision missing = (MissingOpenRevision) openRev;
+                logger.info(String.format(Locale.ENGLISH,
+                        "Revision %s  for document %s is missing, generating stub.",
```
tomblench: Strange double space after first `%s`

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -461,13 +461,27 @@ public void putAttachmentStream(String id, String rev, String attachmentName, St
...
                     String id = doc.ok.getId();
                     if (!revsMap.containsKey(id)) {
                         revsMap.put(id, new ArrayList<DocumentRevs>());
                     }
                     revsMap.get(id).add(doc.ok);
+                } else {
+                    // error will be set here, so lets just access the map to get at the id
+                    @SuppressWarnings("unchecked")
+                    String id = (String)doc.error.get("id");
+                    // build the stub revision, this might have gone due to revs_limit.
```
tomblench: due to revs_limit being exceeded, or compaction, or something else? i think the comment is a bit misleading

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -461,13 +461,27 @@ public void putAttachmentStream(String id, String rev, String attachmentName, St
...
+                if (doc.ok != null){
                     String id = doc.ok.getId();
                     if (!revsMap.containsKey(id)) {
                         revsMap.put(id, new ArrayList<DocumentRevs>());
                     }
                     revsMap.get(id).add(doc.ok);
+                } else {
+                    // error will be set here, so lets just access the map to get at the id
+                    @SuppressWarnings("unchecked")
+                    String id = (String)doc.error.get("id");
```
tomblench: You could define an inner class for `error` inside `BulkGetResponse` and have the deserialisation done for you. Then you won't need the casts or `@SuppressWarnings("unchecked")`

rhyshort: I could, but I'd rather not, since we only need two properties out of the map, and we only need it here, I don't think there is much point to add a whole class to avoid two casts and a localised suppression.

rhyshort:
Update: After some discussions on slack, the best way to avoid a missing revision is to include the `latest=true` parameter with the request. That will result in an ancestor of the revision request (even if it is deleted) being returned.

There is still the potential to have a missing response from the sever however this will be for revisions, which don't exist, and the replicator will error accordingly.

The test case has been changed so it modifies the changes feed to return a non leaf revision and to verify that the expected leaf revision is inserted into the database with the correct generation.

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -409,6 +409,9 @@ public void putAttachmentStream(String id, String rev, String attachmentName, St
 
         Map<String, Object> options = new HashMap<String, Object>();
         options.put("revs", true);
+        // by adding latest we should never receive a "missing" response from the server. An ancestor
```
ricellis: I don't think this comment is quite right. The behaviour is that the latest descendant will be returned even if the ancestor we asked for has been deleted.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/MissingRevsReplicationTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Test
+    public void testReplicationWithMissingRevision() throws Exception {
+        // Create doc
+        Bar a = BarUtils.createBar(remoteDb, "testdoc", "alpha", 1);
+        // Update doc
+        final Bar b = BarUtils.updateBar(remoteDb, "testdoc", "beta", 2);
+
+        Bar c = BarUtils.updateBar(remoteDb, "testdoc", "gamma", 3);
+        // We have two remote revisions
```
ricellis: looks like 3 to me

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/MissingRevsReplicationTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+        // Do the pull replication
+        pull();
+        // Validate the document was created at the correct rev
+        DocumentRevision rev = datastore.getDocument("testdoc");
+        Assert.assertNotNull("The document should be present", rev);
+        Assert.assertEquals("The local revision should be the third generation",
+                c.getRevision(),
+                rev.getRevision());
+        // Validate that there is only a single revision
```
ricellis: This doesn't appear to be how the test works any more. A tree depth of 2 indicates 3 revisions.

rhyshort: fixed in 	8a80522

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/MissingRevsReplicationTest.java
rhyshort: fixed in 	8a80522

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchClient.java
rhyshort: fixed in 	8a80522

rhyshort:
Copyrights fixed in 8a80522 and changes updated in ffb07a0

PR #386, opened 2016-11-09
Avoid creating duplicate indexes with different generated names.

### LINKED ISSUE #377: Calling ensureIndexed multiple times for the same fields creates multiple indexes
The name of the method `ensureIndexed` makes it very easy to inadvertently create multiple indexes for the same field. Having multiple indexes increases the work required when updating the indexes and also increases the disk space needed to store the indexes.

For example:

``` java
IndexManager im = new IndexManager(ds);
im.ensureIndexed(Collections.singletonList((Object) "myField"));
```

creates an index with a uniquely generated name.

A later second call to:

``` java
im.ensureIndexed(Collections.singletonList((Object) "myField"));
```

returns a new name and generates a second index, although one already exists that indexes that field.

For completeness this is the behaviour when using named indexes:

``` java
im.ensureIndexed(Collections.singletonList((Object) "myField"), "myIndex");
im.ensureIndexed(Collections.singletonList((Object) "myField"), "myIndex");
```

results in only a single index being created called `myIndex`.

Calling:

``` java
im.ensureIndexed(Collections.singletonList((Object) "myField"), "myIndex");
im.ensureIndexed(Collections.singletonList((Object) "myField"), "mySecondIndex");
```

results in two indexes being created called `myIndex` and `mySecondIndex`.

I think this is open to discussion on what the appropriate behaviour should be if an attempt is made to create a named index for a field that already has a named index.
- Should it throw an exception?
- Should it return the name of the existing index?
- Should it keep existing behaviour and generate a new index?

Note that fixing this potentially has a migration implication in terms of cleaning up any of the duplicated indexes, although if any are referred to by (generated) name that would be very problematic. It may just be that deleting and re-creating indexes is the best approach.

rhyshort:
## What
Avoid creating duplicate indexes with different generated names, when calling `ensureIndexed` without a name parameter.

## How

- Implement comparable on FieldSort
- Ensure all `ensureIndex` methods call down into the same method.
- Compare the fields for an index with `_id` and `_rev` omitted, with the remaining fields sorted.

## Tests

New Tests added to `IndexManagerTests`

## Issues

Fixes #377

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
@@ -18,7 +18,7 @@ package com.cloudant.sync.query;
  * Created by tomblench on 28/09/2016.
  */
 
-public class FieldSort implements Comparable<FieldSort> {
+public class FieldSort {
 
     public final String field;
     public final Direction sort;
@@ -64,17 +64,6 @@ public class FieldSort implements Comparable<FieldSort> {
                 ", sort=" + sort +
                 '}';
     }
-
-    @Override
-    public int compareTo(FieldSort other) {
-
-        int fieldCompare = field.compareTo(other.field);
-        if( fieldCompare == 0){
-            return sort.compareTo(other.sort);
-        }
-        return fieldCompare;
-    }
-
 
     public enum Direction {
         ASCENDING,
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
@@ -18,7 +18,7 @@ package com.cloudant.sync.query;
  * Created by tomblench on 28/09/2016.
  */
 
-public class FieldSort {
+public class FieldSort implements Comparable<FieldSort> {
 
     public final String field;
     public final Direction sort;
@@ -64,6 +64,16 @@ public class FieldSort {
                 ", sort=" + sort +
                 '}';
     }
+
+    @Override
+    public int compareTo(FieldSort other) {
+
+        if(field.compareTo(other.field) == 0){
+            return sort.compareTo(other.sort);
+        }
+        return field.compareTo(other.field);
+    }
+
 
     public enum Direction {
         ASCENDING,
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
@@ -68,10 +68,11 @@ public class FieldSort implements Comparable<FieldSort> {
     @Override
     public int compareTo(FieldSort other) {
 
-        if(field.compareTo(other.field) == 0){
+        int fieldCompare = field.compareTo(other.field);
+        if( fieldCompare == 0){
             return sort.compareTo(other.sort);
         }
-        return field.compareTo(other.field);
+        return fieldCompare;
     }
 
 
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
@@ -262,7 +262,9 @@ public class IndexManagerImpl implements IndexManager {
         // if an index already exists that matches the request index definition.
         if (indexName == null) {
             List<Index> indexes = this.listIndexes();
-            Collections.sort(fieldNames);
+            // Create a copy of the field names list so we know its mutable.
+            List<FieldSort> mFieldNames = new ArrayList<FieldSort>(fieldNames);
+            Collections.sort(mFieldNames);
             for (Index index : indexes) {
                 Collections.sort(index.fieldNames);
                 if (!fieldNames.equals(filterMeta(index.fieldNames))) {
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
@@ -207,15 +207,6 @@ public class IndexManagerImpl implements IndexManager {
      */
     @Override
     public String ensureIndexed(List<FieldSort> fieldNames) throws QueryException {
-        List<Index> indexes = this.listIndexes();
-        Collections.sort(fieldNames);
-        for(Index index: indexes){
-            Collections.sort(index.fieldNames);
-            if (fieldNames.equals(filterMeta(index.fieldNames))){
-                return index.indexName;
-            }
-        }
-
         return this.ensureIndexed(fieldNames, null);
     }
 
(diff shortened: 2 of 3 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
```diff
@@ -205,9 +207,19 @@ public void close() {
      */
     @Override
     public String ensureIndexed(List<FieldSort> fieldNames) throws QueryException {
+        List<Index> indexes = this.listIndexes();
+        Collections.sort(fieldNames);
+        for(Index index: indexes){
```
tomblench: Doesn't account for the `indexType` or `tokenize` options on the Index. I wonder if it's easier to select the already existed index(es) directly with a SQL query.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexManagerTest.java
```diff
@@ -40,6 +42,50 @@ public void enusureIndexedGeneratesIndexName() throws QueryException {
     }
 
     @Test
+    public void ensureIndexedGeneratesSingleIndexForSameFields() throws QueryException {
```
tomblench: Should have a test for same fields but differing indexType/tokenize options

rhyshort: done in 2e43dd7

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
rhyshort: done in 2e43dd7

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
```diff
@@ -205,9 +207,19 @@ public void close() {
...
+            Collections.sort(index.fieldNames);
+            if (fieldNames.equals(filterMeta(index.fieldNames))){
+                return index.indexName;
+            }
+        }
+
         return this.ensureIndexed(fieldNames, null);
```
brynh: Do we need to do all the stuff before this `return` statement in this method? Can't we just do the `return` statement because the checks before this are included in the version of `ensureIndexed` ultimately called?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
```diff
@@ -65,6 +65,16 @@ public String toString() {
                 '}';
     }
 
+    @Override
+    public int compareTo(FieldSort other) {
+
+        if(field.compareTo(other.field) == 0){
```
rhyshort: done in 	3772026

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
rhyshort: Should have been removed in the previous commit; removed in cb057b6

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
```diff
@@ -256,6 +257,32 @@ public String ensureIndexed(List<FieldSort> fieldNames,
                                 String indexName,
                                 IndexType indexType,
                                 String tokenize) throws QueryException {
+
+        // If the index name is null, a name should be generated, but first we should check
+        // if an index already exists that matches the request index definition.
+        if (indexName == null) {
+            List<Index> indexes = this.listIndexes();
+            Collections.sort(fieldNames);
```
brynh: I think we need to create a mutable copies of the lists here before we sort them, as this crashes if the list is immutable.  The tests seem to be currently failing for this reason because they use `Collections.singletonList()`, which creates an immutable list [here](https://github.com/cloudant/sync-android/blob/cb057b68c12a517370ce9df5a5fd3fe6d01198be/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/query/IndexManagerTest.java#L46)

rhyshort: fixed in 3521314 and ccd315f

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
```diff
@@ -65,6 +65,17 @@ public String toString() {
                 '}';
     }
 
+    @Override
+    public int compareTo(FieldSort other) {
+
+        int fieldCompare = field.compareTo(other.field);
+        if( fieldCompare == 0){
```
tomblench: odd spacing here

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
```diff
@@ -256,6 +257,34 @@ public String ensureIndexed(List<FieldSort> fieldNames,
...
                                 IndexType indexType,
                                 String tokenize) throws QueryException {
+
+        // If the index name is null, a name should be generated, but first we should check
+        // if an index already exists that matches the request index definition.
+        if (indexName == null) {
+            List<Index> indexes = this.listIndexes();
+            // Create a copy of the field names list so we know its mutable.
+            List<FieldSort> mFieldNames = new ArrayList<FieldSort>(fieldNames);
+            Collections.sort(mFieldNames);
```
tomblench: I'm not too keen on this copy and sort. Also how do we know that `index.fieldNames` (inside the for loop) doesn't need to be copied before sorting as well?

I would prefer a different strategy such as:
- put them both into a `Set` and use `Set.equals()`
- don't sort them but use `a.containsAll(b) && a.size() == b.size()`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/FieldSort.java
rhyshort: method removed in 2759746

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
rhyshort: done in 2759746

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexManagerImpl.java
```diff
@@ -256,6 +260,32 @@ public String ensureIndexed(List<FieldSort> fieldNames,
...
                                 String tokenize) throws QueryException {
+
+        // If the index name is null, a name should be generated, but first we should check
+        // if an index already exists that matches the request index definition.
+        if (indexName == null) {
+            List<Index> indexes = this.listIndexes();
+            Set<FieldSort> fieldNamesSet = new HashSet<FieldSort>(filterMeta(fieldNames));
+            for (Index index : indexes) {
+                Set<FieldSort> indexFieldNamesSet = new HashSet<FieldSort>(filterMeta(index.fieldNames));
+                if (!(indexFieldNamesSet.containsAll(fieldNamesSet) && indexFieldNamesSet.size() == fieldNamesSet.size())) {
```
rhyshort: done in 0f0a3fb

brynh: Actually, scratch that last comment, shouldn't this just be `if (!indexFieldNamesSet.equals(fieldNamesSet))` as these are both sets.

PR #387, opened 2016-11-09
v2 API - refactor class and package names

tomblench:
Part 6 of n PRs for feature-284-refactor-datastore-indexmanager. See parent issue #284.

The focus of this PR is choosing appropriate names for classes and packages which have been sub-optimal in the past.

* Rename `IndexManager` to `Query`
* Rename `DatabaseX` events to `DocumentStoreX`
* Rename `DatastoreXException` exceptions to `DocumentStoreXException`
* Main top-level packages now `com.cloudant.sync.{documentstore,event,internal,query,replication}`

TODO

* ~consider using an interface on `DocumentRevision` to hide methods like `getParent` and `getInternalNumericId` which shouldn't be on the public API.~
* ~update findbugs config appropriately~
* ~update javadoc so that `com.cloudant.sync.internal` stuff is hidden~


rhyshort:
Do we really need the `ErrorInfo` class? It seems pretty pointless to me.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/QueryException.java
```diff
@@ -1,12 +1,12 @@
 package com.cloudant.sync.query;
```
rhyshort: Missing Copyright.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicationPolicyManager.java
```diff
@@ -1,8 +1,8 @@
 package com.cloudant.sync.replication;
```
rhyshort: Missing Copyright.

tomblench:
@rhyshort good point, do you think we should replace its use with a direct usage of `Throwable`s?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/notifications/DocumentStoreClosed.java
```diff
@@ -12,9 +12,9 @@
  * and limitations under the License.
  */
 
-package com.cloudant.sync.notifications;
+package com.cloudant.sync.event.notifications;
 
-import com.cloudant.sync.datastore.DocumentStore;
+import com.cloudant.sync.documentstore.DocumentStore;
```
rhyshort: Copyright needs updating

rhyshort:
Yeah I think so, since its only wrapping a throwable

tomblench:
@rhyshort all files without a licence now have them, and the number of licence variants has been reduced as much as possible. I think this is all the appropriate licence activity for now, but we still need to do #390 some time.

PR #392, opened 2016-11-17
Fix sample app for v2 API

tomblench:
_What_: Fix sample app so it builds and works on new v2 API

_Why_: As well as being a pre-requisite for merging the feature branch to master, this is a useful exercise to see what the API looks like from a user's point of view

_Testing_: App builds and runs correctly


sample/todo-sync/src/com/cloudant/todo/TasksModel.java
```diff
@@ -70,10 +69,9 @@ public TasksModel(Context context) {
                 DATASTORE_MANGER_DIR,
                 Context.MODE_PRIVATE
         );
-        DatastoreManager manager = DatastoreManager.getInstance(path.getAbsolutePath());
         try {
-            this.mDatastore = manager.openDatastore(TASKS_DATASTORE_NAME);
-        } catch (DatastoreNotCreatedException dnce) {
+            this.mDocumentStore = DocumentStore.getInstance(new File(path, TASKS_DATASTORE_NAME));
+        } catch (DocumentStoreNotOpenedException dnce) {
```
ricellis: I wondered if we also needed to catch the `DocumentStoreNotCreatedException` here, but I see we only have a NotOpened, I thought we were having both types?
If we do we could probably just catch the `DocumentStoreException`?

sample/todo-sync/src/com/cloudant/todo/TasksModel.java
```diff
@@ -150,15 +148,15 @@ public Task updateDocument(Task task) throws ConflictException {
...
     }
 
     /**
      * <p>Returns all {@code Task} documents in the datastore.</p>
      */
     public List<Task> allTasks() {
-        int nDocs = this.mDatastore.getDocumentCount();
-        List<DocumentRevision> all = this.mDatastore.getAllDocuments(0, nDocs, true);
+        int nDocs = this.mDocumentStore.database.getDocumentCount();
+        List<? extends DocumentRevision> all = this.mDocumentStore.database.getAllDocuments(0, nDocs, true);
```
ricellis: Why the wildcard? Are we exposing sub-types of `DocumentRevision`?

sample/todo-sync/src/com/cloudant/todo/TasksModel.java
```diff
@@ -233,8 +231,8 @@ public void reloadReplicationSettings()
         // Set up the new replicator objects
         URI uri = this.createServerURI();
 
-        mPullReplicator = ReplicatorBuilder.pull().to(mDatastore).from(uri).build();
-        mPushReplicator = ReplicatorBuilder.push().from(mDatastore).to(uri).build();
+        mPullReplicator = ReplicatorBuilder.pull().to(mDocumentStore.database).from(uri).build();
+        mPushReplicator = ReplicatorBuilder.push().from(mDocumentStore.database).to(uri).build();
```
tomblench: Agree this would look more user friendly. @ricellis what do you think?

ricellis: Yeah it would, although IIRC we don't update indexes automatically after a replication so we should be careful to not imply that intent with the argument. On the plus side passing the `DocumentStore` gives us scope to improve that.

rhyshort: #395 Raised to do this in future.

ricellis:
Guess this is getting a bit stale now. Is the plan to rebase this once walkthrough suggested changes are complete and then re-review?

tomblench:
See #437

PR #393, opened 2016-11-17
Fix find bugs warnings

### LINKED ISSUE #388: Resolve outstanding findbugs issues
Fix the remaining issues reported by findbugs and enable the most sensitive level of reporting.

rhyshort:
Fix findbugs warnings and enable findbugs at low level.

Fixes #388

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
@@ -237,9 +237,10 @@ public class DatabaseImpl implements Database {
             throw new RuntimeException(e);
         } catch (ExecutionException e) {
             logger.log(Level.SEVERE, "Failed to get last Sequence", e);
-            if (e.getCause() != null) {
-                if (e.getCause() instanceof IllegalStateException) {
-                    throw (IllegalStateException) e.getCause();
+            Throwable cause = e.getCause();
+            if (cause != null) {
+                if (cause instanceof IllegalStateException) {
+                    throw (IllegalStateException)cause;
                 }
             }
         }
@@ -928,7 +929,7 @@ public class DatabaseImpl implements Database {
                             if (newWinnerTx.getClass().equals(DocumentRevision.class)) {
                                 // user gave us a new DocumentRevision instance
                                 isNewOrUpdatedRevision = true;
-                            } else if (newWinnerTx.getClass().equals(InternalDocumentRevision.class)) {
+                            } else if (newWinnerTx instanceof InternalDocumentRevision) {
                                 // user gave us an existing InternalDocumentRevision instance - did the body or attachments change?
                                 InternalDocumentRevision newWinnerTxInternal = (InternalDocumentRevision)newWinnerTx;
                                 if (newWinnerTxInternal.isBodyModified()) {
@@ -964,9 +965,10 @@ public class DatabaseImpl implements Database {
                     }));
         } catch (ExecutionException e) {
             logger.log(Level.SEVERE, "Failed to resolve Conflicts", e);
-            if (e.getCause() != null) {
-                if (e.getCause() instanceof IllegalArgumentException) {
-                    throw (IllegalArgumentException) e.getCause();
+            Throwable cause = e.getCause();
+            if (cause != null) {
+                if (cause instanceof IllegalArgumentException) {
+                    throw (IllegalArgumentException) cause;
                 }
             }
         }
@@ -1223,9 +1225,10 @@ public class DatabaseImpl implements Database {
             return deletedRevision;
         } catch (ExecutionException e) {
             logger.log(Level.SEVERE, "Failed to delete document", e);
-            if (e.getCause() != null) {
-                if (e.getCause() instanceof ConflictException) {
-                    throw (ConflictException) e.getCause();
+            Throwable cause = e.getCause();
+            if (cause != null) {
+                if (cause instanceof ConflictException) {
+                    throw (ConflictException) cause;
                 }
             }
         }
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
@@ -618,24 +618,7 @@ public class DatabaseImpl implements Database {
     public String getPublicIdentifier() throws DocumentStoreException {
         Misc.checkState(this.isOpen(), "Database is closed");
         try {
-            return get(queue.submit(new SQLCallable<String>() {
-                @Override
-                public String call(SQLDatabase db) throws Exception {
-                    Cursor cursor = null;
-                    try {
-                        cursor = db.rawQuery("SELECT value FROM info WHERE key='publicUUID'", null);
-                        if (cursor.moveToFirst()) {
-                            return "touchdb_" + cursor.getString(0);
-                        } else {
-                            throw new IllegalStateException("Error querying PublicUUID, " +
-                                    "it is probably because the sqlDatabase is not probably " +
-                                    "initialized.");
-                        }
-                    } finally {
-                        DatabaseUtils.closeCursorQuietly(cursor);
-                    }
-                }
-            }));
+            return get(queue.submit(new GetPublicIdentifierCallable()));
         } catch (ExecutionException e) {
             logger.log(Level.SEVERE, "Failed to get public ID", e);
             throw new DocumentStoreException("Failed to get public ID", e);
@@ -844,29 +827,7 @@ public class DatabaseImpl implements Database {
 
         // TODO hoist down queue.submit to just surround the call to RevsDiffBatchCallable
         try {
-            return get(queue.submit(new SQLCallable<Map<String, List<String>>>() {
-                @Override
-                public Map<String, List<String>> call(SQLDatabase db) throws Exception {
-
-                    ValueListMap<String, String> missingRevs = new ValueListMap<String, String>();
-
-                    // Break down by docId first to avoid potential rev ID clashes between doc IDs
-                    for (Map.Entry<String, List<String>> entry : revisions.entrySet()) {
-                        String docId = entry.getKey();
-                        List<String> revs = entry.getValue();
-                        // Partition into batches to avoid exceeding placeholder limit
-                        // The doc ID will use one placeholder, so use limit - 1 for the number of
-                        // revs for the remaining placeholders.
-                        List<List<String>> batches = CollectionUtils.partition(revs,
-                                SQLITE_QUERY_PLACEHOLDERS_LIMIT - 1);
-
-                        for (List<String> revsBatch : batches) {
-                            missingRevs.addValuesToKey(docId, new RevsDiffBatchCallable(docId, revsBatch).call(db));
-                        }
-                    }
-                    return missingRevs;
-                }
-            }));
+            return get(queue.submit(new RevsDiffCallable(revisions)));
         } catch (ExecutionException e) {
             logger.log(Level.SEVERE, "Failed to do revsdiff", e);
         }
@@ -1263,4 +1224,52 @@ public class DatabaseImpl implements Database {
     }
 
 
+    private static class GetPublicIdentifierCallable implements SQLCallable<String> {
+        @Override
+        public String call(SQLDatabase db) throws Exception {
+            Cursor cursor = null;
+            try {
+                cursor = db.rawQuery("SELECT value FROM info WHERE key='publicUUID'", null);
+                if (cursor.moveToFirst()) {
+                    return "touchdb_" + cursor.getString(0);
+                } else {
+                    throw new IllegalStateException("Error querying PublicUUID, " +
+                            "it is probably because the sqlDatabase is not probably " +
+                            "initialized.");
+                }
+            } finally {
+                DatabaseUtils.closeCursorQuietly(cursor);
+            }
+        }
+    }
+
+    private static class RevsDiffCallable implements SQLCallable<Map<String, List<String>>> {
+        private final Map<String, List<String>> revisions;
+
+        public RevsDiffCallable(Map<String, List<String>> revisions) {
+            this.revisions = revisions;
+        }
+
+        @Override
+        public Map<String, List<String>> call(SQLDatabase db) throws Exception {
+
+            ValueListMap<String, String> missingRevs = new ValueListMap<String, String>();
+
+            // Break down by docId first to avoid potential rev ID clashes between doc IDs
+            for (Map.Entry<String, List<String>> entry : revisions.entrySet()) {
+                String docId = entry.getKey();
+                List<String> revs = entry.getValue();
+                // Partition into batches to avoid exceeding placeholder limit
+                // The doc ID will use one placeholder, so use limit - 1 for the number of
+                // revs for the remaining placeholders.
+                List<List<String>> batches = CollectionUtils.partition(revs,
+                        SQLITE_QUERY_PLACEHOLDERS_LIMIT - 1);
+
+                for (List<String> revsBatch : batches) {
+                    missingRevs.addValuesToKey(docId, new RevsDiffBatchCallable(docId, revsBatch).call(db));
+                }
+            }
+            return missingRevs;
+        }
+    }
 }
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DocumentRevisionBuilder.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DocumentRevisionBuilder.java
@@ -28,26 +28,6 @@ import java.util.List;
  */
 public class DocumentRevisionBuilder {
 
-
-    static {
-        // these values are defined http://docs.couchdb.org/en/latest/api/document/common.html
-
-        allowedPrefixes = Arrays.asList(new String[]{
-                "_id",
-                "_rev",
-                "_deleted",
-                "_attachments",
-                "_conflicts",
-                "_deleted_conflicts",
-                "_local_seq",
-                "_revs_info",
-                "_revisions"
-        });
-
-    }
-
-    private static List<String> allowedPrefixes;
-
     private String docId = null;
     private String revId = null;
     private DocumentBody body = null;
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/InternalDocumentRevision.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/InternalDocumentRevision.java
@@ -39,8 +39,6 @@ public class InternalDocumentRevision extends DocumentRevision implements
 
     private long parent = -1L;
 
-    protected ChangeNotifyingMap<String, Attachment> attachments;
-
     private boolean bodyModified = false;
 
     public InternalDocumentRevision(String docId, String revId, DocumentBody body,
@@ -204,7 +202,7 @@ public class InternalDocumentRevision extends DocumentRevision implements
 
     @Override
     public ChangeNotifyingMap<String, Attachment> getAttachments() {
-        return attachments;
+        return (ChangeNotifyingMap<String, Attachment>) attachments;
     }
 
     @Override
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/SavedAttachment.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/SavedAttachment.java
@@ -66,7 +66,7 @@ public class SavedAttachment extends Attachment {
 
         this.revpos = revpos;
         this.seq = seq;
-        this.key = Arrays.copyOf(key, key.length);
+        this.key = key == null? null : Arrays.copyOf(key, key.length);
         this.length = length;
         this.encodedLength = encodedLength;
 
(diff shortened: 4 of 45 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
```diff
@@ -1260,4 +1224,52 @@ public DocumentRevision deleteDocumentFromRevision(final DocumentRevision rev) t
     }
 
 
+    private static class GetPublicIdentifierCallable implements SQLCallable<String> {
```
tomblench: Callables go in their own files

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
```diff
@@ -928,7 +890,7 @@ public Void call(SQLDatabase db) throws Exception {
                             if (newWinnerTx.getClass().equals(DocumentRevision.class)) {
                                 // user gave us a new DocumentRevision instance
                                 isNewOrUpdatedRevision = true;
-                            } else if (newWinnerTx.getClass().equals(InternalDocumentRevision.class)) {
+                            } else if (newWinnerTx instanceof InternalDocumentRevision) {
```
tomblench: is instanceof correct here? we need to distinguish between subtypes and supertypes and I thought that getClass() was the better check for this

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/QueuingExecutorCompletionService.java
```diff
@@ -14,6 +14,7 @@
 
 package com.cloudant.sync.internal.replication;
 
+import java.util.Observable;
```
ricellis: I don't think this import is needed.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -108,8 +109,9 @@ private ExecuteResult(InputStream stream,
                 case 3:
                     // 3xx redirection
                     throw new CouchException("Unexpected redirection (3xx) code encountered", responseCode);
-                case 4:
+                case 4: case 5:
                     // 4xx errors normally mean we are not authenticated so we shouldn't retry
+                    // 5xx errors are transient server errors so we should retry
```
tomblench: logic is broken here, case 5 previously had `needsCouchException = true`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/InternalDocumentRevision.java
```diff
@@ -204,7 +202,7 @@ public int getGeneration() {
 
     @Override
     public ChangeNotifyingMap<String, Attachment> getAttachments() {
-        return attachments;
+        return (ChangeNotifyingMap<String, Attachment>) attachments;
```
tomblench: this cast looks ugly compared to using a more specific type on the field - what's the reason for it?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexCreator.java
```diff
@@ -394,4 +389,10 @@ private static String generateIndexName(Set<String> existingIndexNames) throws E
         }
     }
 
+    private static class ListIndexesCallable implements SQLCallable<List<Index>> {
```
tomblench: seems pretty pointless to have a callable for a 1 liner

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java
```diff
@@ -390,4 +319,100 @@ public DBParameter(String tableName, ContentValues contentValues) {
         }
     }
 
+    private static class SequenceNumberForIndexCallable implements SQLCallable<Long> {
```
tomblench: callables should (probably, if we're being consistent with past usage) go into their own files

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java.orig
```diff
@@ -0,0 +1,460 @@
+/*
```
tomblench: you checked in a file generated by patch/diff

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -420,8 +419,7 @@ public void putAttachmentStream(String id, String rev, String attachmentName, St
             options.put("att_encoding_info", true);
         }
         options.put("open_revs", jsonHelper.toJson(revisions));
-        return this.getDocument(id, options, new TypeReference<List<OpenRevision>>() {
-        });
+        return this.getDocument(id, options,new CouchClientTypeReference<List<OpenRevision>>());
```
ricellis: needs a space after `options,`
and
we can probably benefit from having some static fields representing the types we use most e.g. :
`static final TYPE_REF_LIST_OPEN_REVS = new CouchClientTypeReference<List<OpenRevision>>();`
and equivalently for `List<Response>`, `DocumentRevs`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java.orig
```diff
@@ -0,0 +1,460 @@
...
+ * Copyright © 2014 Cloudant, Inc. All rights reserved.
```
ricellis: This file is not a valid Java class any more, looks like bad merging.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexCreator.java
rhyshort: It saves the overhead of having the reference to the enclosing class, not sure how much of a overhead it is, @ricellis could probably explain more.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java
rhyshort: Yes they should, but that would be part of the issue to host them into callables and consolidate where required / appropriate  .

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java.orig
rhyshort: Removed in 92d7377

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
rhyshort: Wouldn't the else block on 120 take care of that? Perhaps you mean that `fatal` should only be set to `true` if the response code isn't 5xx?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
rhyshort: I think so, unless we are worried that the ProjectDocumentRevision could come into here?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java
```diff
@@ -199,9 +199,6 @@ private void updateIndex(final String indexName,
                                                             initialArgs,
                                                             indexName,
                                                             rev);
-                if (parameter == null) {
```
ricellis: I can see this and it's friend a few lines further on were removed because `com.cloudant.sync.internal.query.IndexUpdater#populateDBParameter` doesn't appear to be able to return null. However, I wonder if that method actually needs looking at because there is no `else` case for if we couldn't figure out the type and then not indexing that field.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
rhyshort: This changed slightly in a later commit, it now uses the `JavaType` class from Jackson, (`TypeReference`s don't work as static class, not sure why.) The rest is still valid, I'll add the fields as required.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/SavedAttachment.java
```diff
@@ -66,7 +66,7 @@ public SavedAttachment(long seq, String name, byte[] key, String type, Encoding
 
         this.revpos = revpos;
         this.seq = seq;
-        this.key = Arrays.copyOf(key, key.length);
+        this.key = key == null? null : Arrays.copyOf(key, key.length);
```
ricellis: spacing incorrect around `?`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DocumentRevisionBuilder.java
```diff
@@ -28,26 +28,6 @@
  */
 public class DocumentRevisionBuilder {
 
-
```
ricellis: Was all this just obsolete? If not where did it go?

rhyshort: It was completely unused. I think it stems from a time before we allowed users to `new` up `DocumentRevisions`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/util/Misc.java
```diff
@@ -112,8 +113,8 @@ public static void checkNotNull(Object param, String errorMessagePrefix) throws
      */
     public static void checkNotNullOrEmpty(String param, String errorMessagePrefix) throws
             IllegalArgumentException {
-        checkArgument(!isStringNullOrEmpty(param), (errorMessagePrefix != null ? errorMessagePrefix :
-                "Parameter") + " must not be " + param == null ? "null." : "empty.");
+        // Null is never passed as the parameter here, as per FindBugs and Intellij inspections.
```
rhyshort: Just double checked the usages, the front facing API uses this. First glance it was just the SQL code, so I removed the check, backing out and replacing with exclude.

rhyshort: done in 62136e3

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/SavedAttachment.java
rhyshort: fixed in 27b0d45

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexCreator.java
ricellis: Right it is a reduction in memory overhead. We can be pragamatic, if `IndexCreator` is a singleton then I think we're only ever going to have one copy of it and it's anonymous inner. I think the warning is more relevant for cases where we might generate hundreds of instances of inners each holding an unnecessary reference (in the case that they could be a replaced by a static class). Nonetheless we have a made a decision to extract callables so we should probably just be consistent.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java
rhyshort: That probably needs some investigation if that method is doing the "right thing" I think it used to add `null` to the content values if it couldn't figure it out, and then `null` values would be skipped later in the code iirc, @tomblench Does that sound familiar for the method `com.cloudant.sync.internal.query.IndexUpdater#populateDBParameter` ?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
tomblench: I probably meant case 5 previously had `this.fatal = false`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DocumentRevisionBuilder.java
rhyshort: raised #396 to investigate the potential loss of function.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java
tomblench: Correct, populateDBParameter would put a `null` into `contentValues` if the type wasn't one we understood (https://github.com/cloudant/sync-android/blob/master/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/IndexUpdater.java#L330). And then that `null` would be skipped later, all entirely pointless.

However, that's for one `contentValues` entry on `DBParameter`.  These null checks are correct to be removed since `populateDBParameter` never returns null.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
rhyshort: fixed in 582d769

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
rhyshort: 546ab10

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/util/Misc.java
```diff
@@ -113,7 +113,8 @@ public static void checkNotNull(Object param, String errorMessagePrefix) throws
      */
     public static void checkNotNullOrEmpty(String param, String errorMessagePrefix) throws
             IllegalArgumentException {
-        // Null is never passed as the parameter here, as per FindBugs and Intellij inspections.
+        checkArgument(!isStringNullOrEmpty(param), (errorMessagePrefix != null ? errorMessagePrefix :
+              "Parameter") + " must not be " + param == null ? "null." : "empty.");
         checkArgument(!isStringNullOrEmpty(param), errorMessagePrefix + " must not be empty.");
```
ricellis: You re-added the old line, but left the new code in place, now we might check twice.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/json/JSONHelper.java
```diff
@@ -104,4 +107,17 @@ public JavaType mapStringToObject(){
         return this.getTypeFactory().constructParametricType(Map.class,String.class,Object.class);
     }
 
+    public JavaType mapStringMissingRevisions(){
+        return this.getTypeFactory().constructParametricType(Map.class, String.class, CouchClient
```
ricellis: I still think we could make these static constants, but I'm happy for that to defer to some other time. In fact I think we could make this whole JSONHelper utility static instead of having loads of them.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
ricellis: Yes, this is ok unless we are concerned about sub-classes of `InternalDocumentRevision` that we should not cast or assume to be modified.

ricellis:
Approved, but don't forget to remove that redundant line.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/util/Misc.java
rhyshort: fixed in fbdae4e

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -108,9 +109,10 @@ private ExecuteResult(InputStream stream,
                 case 3:
                     // 3xx redirection
                     throw new CouchException("Unexpected redirection (3xx) code encountered", responseCode);
-                case 4:
+                case 4: case 5:
```
tomblench: I still don't like the logic here. Breaking out 1xx,2xx,3xx,4xx,5xx is the most logical way to partition things because they are distinct ranges of error codes with different means. It doesn't make sense to effectively say "if 4 or 5" and then later say "oh but only if 5". I think it's more logical to have a case statement breaking them down to the hundreds and then refining within each hundred.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/sqlite/SQLDatabaseFactory.java
```diff
@@ -89,7 +90,9 @@ private static boolean isFtsAvailable() {
     public static SQLDatabase openSQLDatabase(File dbFile, KeyProvider provider) throws IOException, SQLException {
         Misc.checkNotNull(dbFile, "dbFile");
         File dbDirectory = dbFile.getParentFile();
-        dbDirectory.mkdirs();
+        if (!dbDirectory.mkdirs()){
+            logger.info(String.format(Locale.ENGLISH, "Did not create directories for path: %s directories may already exist", dbFile));
```
tomblench: Not sure if this deserves a logger.info. The more interesting case is probably when it does create the directory.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/util/Misc.java
```diff
@@ -50,9 +51,11 @@ public static boolean isRunningOnAndroid() {
             while ((bytesRead = shaFis.read(buf)) != -1) {
                 sha1.update(buf, 0, bytesRead);
             }
+        } catch (RuntimeException e){
+            throw e;
         } catch (Exception e) {
             logger.log(Level.WARNING,"Problem calculating SHA1 for stream",e);
-            return null;
+            return new byte[]{};
```
tomblench: It's possible that throwing an exception is the more correct behaviour here

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/encryption/KeyUtilsTest.java
```diff
@@ -42,19 +42,6 @@ public EncryptionKey getEncryptionKey() {
     }
 
     @Test
-    public void sqlCipherKeyForKeyProvider_nullKeyBytes() {
```
tomblench: Why was this removed?

cloudant-sync-datastore-javase/src/main/java/com/cloudant/sync/internal/sqlite/sqlite4java/SQLiteWrapperUtils.java
```diff
@@ -84,7 +84,7 @@ public static SQLiteStatement bindArguments(SQLiteStatement stmt, Object[] bindA
             bindArgs = new Object[]{};
         }
 
-        final int count = bindArgs != null ? bindArgs.length : 0;
+        final int count =  bindArgs.length;
```
tomblench: remove extra space

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/datastore/encryption/KeyUtilsTest.java
rhyshort: Incorrect behaviour,  `EncryptionKey` cannot return null for keyBytes, the constructor ensures this, but the test deliberately bypasses the restriction imposed by the constructor.

cloudant-sync-datastore-javase/src/main/java/com/cloudant/sync/internal/sqlite/sqlite4java/SQLiteWrapperUtils.java
rhyshort: removed in 6861e95

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
rhyshort: backed out in fa092ba

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/sqlite/SQLDatabaseFactory.java
rhyshort: I'd rather have a log message for a failure case than the successful case since the next two checks should work just fine. The level can always be lowered to `FINE`, `FINER` or `FINEST`.

tomblench: Returning false isn't necessarily a failure case. It could mean that the directories were already there.

rhyshort: yes, which the current log message takes account for.

PR #399, opened 2016-11-21
Generate User-Agent at build time

### LINKED ISSUE #262: Update cloudant-http to 2.4.2
Update cloudant-http to 2.4.2

### LINKED ISSUE #362: sync-android and java-cloudant can't be used in the same application
As reported by a customer trying to use both libraries in an Android application, the build fails because there are two classes called `com.cloudant.library.LibraryVersion`.

Because of the layering of projects (the `LibraryVersion` class is actually used by `cloudant-http`), we can't just rename one of the `LibraryVersion` classes.

One potential solution is to unify the use of the `properties` file (currently `mazha.properties` and `client.properties`) so that they have the same name, location, and format. Then `HTTPConnection` loads this resource if available to set the user-agent string and we can get rid of the `*Version` classes.

rhyshort:
## What

Generate the client name and client version for the user agent header at build time.

## How

- Upgrade cloudant-http to 2.7.0
- Added properties file generation to build.gradle
- Change the properties file name to `com.cloudant.sync.client.properties`

## Testing
Manually tested that the user agent is as expected using the sample application and local couch instance.

## Issues 

Fixes #362


cloudant-sync-datastore-core/build.gradle
```diff
@@ -126,3 +126,54 @@ compileToolsJava {
...
 }
+
+//
+// Generate client.properties
+//
+
+//task for generating a client properties file
+class ClientProperties extends DefaultTask {
+
+    //allow this to be configured, default to client.properties
```
ricellis: This comment is not correct based on the default set here. I don't mind if the comment or the default changes, either works as the file is explicitly set anyway.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/ReplicationTestBase.java
```diff
@@ -178,11 +178,11 @@ protected void assertCookieInterceptorPresent(ReplicatorBuilder p, String expect
         respI.setAccessible(true);
         List<HttpConnectionRequestInterceptor> reqIList = (List)reqI.get(p);
         List<HttpConnectionRequestInterceptor> respIList = (List)respI.get(p);
-        Assert.assertEquals(1, reqIList.size());
-        Assert.assertEquals(CookieInterceptor.class, reqIList.get(0).getClass());
+        Assert.assertEquals(2, reqIList.size());
```
ricellis: Since this test is relying on the `UserAgentInterceptor` being added before the `CookieInterceptor`
I think it might be worth having a comment here making that assumption clear in case anyone needs to debug this later.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/PullReplicatorTest.java
```diff
@@ -158,7 +158,7 @@ public void testRequestInterceptorsThroughBuilder() throws Exception {
                 .addRequestInterceptors(interceptorCallCounter)
                 .addResponseInterceptors(interceptorCallCounter);
         if (TestOptions.COOKIE_AUTH) {
-            CookieInterceptor ci = new CookieInterceptor(TestOptions.COUCH_USERNAME, TestOptions.COUCH_PASSWORD);
+            CookieInterceptor ci = new CookieInterceptor(this.remoteDb.couchClient.getRootUri().toString(), TestOptions.COUCH_USERNAME, TestOptions.COUCH_PASSWORD);
```
ricellis: In the other places the `URL` parameter is last, not first.

rhyshort: done in f5af2fa

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/ReplicationTestBase.java
rhyshort: done in bf2bbde

cloudant-sync-datastore-core/build.gradle
rhyshort: done in bf2bbde

ricellis: I think you mean 6411374

brynh:
We should probably also upgrade cloudant-http in the [AndroidTest gradle file](https://github.com/cloudant/sync-android/blob/362-user-agent-interceptor/AndroidTest/app/build.gradle#L88) and mention the upgrade in CHANGES.md

rhyshort:
@brynh Done

PR #411, opened 2016-11-22
Added Jenkinsfile

ricellis:
*What*

Added `Jenkinsfile`.

*How*

Added `Jenkinsfile` and `VERSION` file.
Followed the java-cloudant template but with parallel testing for Android.

*Testing*

No new tests, existing tests pass on Java & Android using Jenkins pipeline.
Parallel testing of Android using local CouchDB 1.6 and Java using remote service.


Jenkinsfile
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+stage('Build') {
+    // Checkout, build and assemble the source and doc
+    node('android-compile') {
```
rhyshort: Wouldn't this label be more accurate if it was java-compile? I think that this is just to compile to source once.

Jenkinsfile
```diff
@@ -0,0 +1,112 @@
...
+    parallel(
+        Java:
+        {
+            node {
+                unstash name: 'built'
+                // findBugs
+                try {
+                    sh './gradlew -Dfindbugs.xml.report=true findbugsMain'
+                } finally {
+                    step([$class: 'FindBugsPublisher', pattern: '**/build/reports/findbugs/*.xml'])
```
rhyshort: More of a question really, could we do this on another node so the tests run even if findbugs fails?

Jenkinsfile
ricellis: We need to compile somewhere that has the Android SDK available (i.e. a node with `ANDROID_HOME` set and pointing to something valid.

Jenkinsfile
ricellis: Yes, we could. I was trying to save the overhead of getting another node because findbugs is lightweight in comparison to the tests.
At the moment this is in line with java-cloudant, I think it would be nice to consolidate some of the shared targets in future and perhaps that would be the right time to re-evaluate what nodes are needed.

Of course there is the alternative argument that we shouldn't waste test execution time when we know we are going to have to make changes and rebuild/retest anyway (i.e. the case when a findbugs is found and fixed).

rhyshort: ok, probably best to fail the findbugs and not waste the execution time, something we can revisit when we common up the code between this and java-cloudant.

Jenkinsfile
ricellis: Following on from our discussion, yes, we can do this anywhere for now because of the old Android level available from maven for us to compile against - so I'll remove the node label restriction.
At some point when we update the minimum Android level we'll need some changes.

ricellis: Done in 46b8739

PR #412, opened 2016-11-22
ReplicatorBuilder takes DocumentStore

### LINKED ISSUE #395: ReplicatorBuilder should take the full `DocumentStore` object.
Currently the `ReplicatorBuilder` takes only the `Database` object from the `DocumentStore` like:

```
mPullReplicator = ReplicatorBuilder.pull().to(mDocumentStore.database).from(uri).build();
 mPushReplicator = ReplicatorBuilder.push().from(mDocumentStore.database).to(uri).build();
```

Which looks a bit off, it would be more natural to take the entire `DocumentStore` object.

tomblench:
Fixes #395 

_What_: Make `ReplicatorBuilder` API more useable by taking `DocumentStore` object instead of `Database`.
_Testing_: All tests pass

### DIFF
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/ReplicatorIdTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/ReplicatorIdTest.java
@@ -14,16 +14,18 @@
 
 package com.cloudant.sync.internal.replication;
 
-import static org.mockito.Mockito.mock;
-
 import com.cloudant.common.CollectionFactory;
-import com.cloudant.sync.internal.documentstore.DatabaseImpl;
+import com.cloudant.sync.documentstore.DocumentStore;
 import com.cloudant.sync.replication.PullFilter;
 import com.cloudant.sync.replication.ReplicatorBuilder;
+import com.cloudant.sync.util.TestUtils;
 
+import org.junit.After;
 import org.junit.Assert;
+import org.junit.Before;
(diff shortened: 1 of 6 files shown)


cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/ReplicatorIdTest.java
```diff
@@ -14,33 +14,50 @@
...
+        TestUtils.deleteTempTestingDir(testDir);
+    }
+
     // source/target switched, so ids should be different
     @Test
     public void pullNotEqualToPush() throws Exception {
         PullStrategy pull = (PullStrategy) ((ReplicatorImpl) ReplicatorBuilder.pull()
                 .from(new URI
-                ("http://default-host/default-database")).to(mock(DatabaseImpl.class)).build
+                ("http://default-host/default-database")).to(documentStore).build
```
ricellis: What is the motivation for not keeping this as a mock, but using `DocumentStore`?

tomblench: @ricellis it's not currently possible to mock our DocumentStore implementation for the purposes of this test. In `PushStrategy` we can't mock `source.database` and in `PullStrategy` we can't mock `target.database` because they are final fields. So this is definitely an argument for turning `DocumentStore.database` and `DocumentStore.query` into getter methods.

PR #414, opened 2016-11-22
Throw exceptions in Misc.getSha1

### LINKED ISSUE #397: Should Misc.getSha1 throw an exception instead of returning an empty byte array?
Currently `Misc. getSha1` returns `null` when it could not calculate the sha1 for a stream, however findbugs regards this as a problem as a result it has been replaced with an empty byte array in pr #393. Some more thought should be put in to wether or not it should throw an exception and if so what exception should be thrown.

tomblench:
Fixes #397

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/util/Misc.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/util/Misc.java
@@ -14,8 +14,10 @@
 
 package com.cloudant.sync.internal.util;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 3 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/util/Misc.java
```diff
@@ -51,11 +53,10 @@ public static boolean isRunningOnAndroid() {
             while ((bytesRead = shaFis.read(buf)) != -1) {
                 sha1.update(buf, 0, bytesRead);
             }
-        } catch (RuntimeException e){
+        } catch (NoSuchAlgorithmException e){
+            throw new IllegalStateException(e);
+        } catch (IOException e) {
```
rhyshort: Remove the catch block for IOException.

Issue #418, opened 2016-11-23
DatabaseImpl.getDocument() exception cases confusing 


tomblench:
I've just noticed that this method will throw `DocumentNotFoundException` in one of two cases:

- The document doesn't exist in the DB
- There was some sort of internal SQL error whilst retrieving the document

This is unhelpful to users and needs to be fixed!

tomblench:
Done in #420

PR #420, opened 2016-11-24
Database APIs throw DocumentStoreException

tomblench:
See also #418 #354

_What_: Declare and throw correct exceptions from `Database`. `DocumentStoreException` is the generic exception for underlying issues with the SQL database etc. `DocumentNotFoundException` etc are used to signal specific error conditions.

_Why_: Some methods throw incorrect exception (eg don't throw the underlying cause) or just return `null` under certain error conditions. This PR aims to fix that.

_Testing_: Existing tests pass, added `DatabaseImplExceptionsTest`.

Future PR
- Rename `DocumentStoreException` to be consistent with the class now being called `Database`.
- Update javadoc
- ~~Revisit empty catch blocks~~

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java
@@ -119,7 +119,7 @@ class IndexUpdater {
         } catch (DocumentStoreException e) {
             String message = String.format(Locale.ENGLISH, "Failed to get changes feed from %d", lastSequence);
             logger.log(Level.SEVERE, message, e);
-            throw new QueryException(e.getCause());
+            throw new QueryException(message, e.getCause());
         }
     }
 
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java
@@ -14,6 +14,7 @@
 
 package com.cloudant.sync.internal.query;
 
+import com.cloudant.sync.documentstore.DocumentStoreException;
 import com.cloudant.sync.internal.android.ContentValues;
 import com.cloudant.sync.documentstore.Changes;
 import com.cloudant.sync.documentstore.Database;
@@ -32,6 +33,7 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.logging.Level;
@@ -108,12 +110,17 @@ class IndexUpdater {
 
         Changes changes;
         long lastSequence = sequenceNumberForIndex(indexName);
-
-        do {
-            changes = database.changes(lastSequence, 10000);
-            updateIndex(indexName, fieldNames, changes, lastSequence);
-            lastSequence = changes.getLastSequence();
-        } while (changes.size() > 0);
+        try {
+            do {
+                changes = database.changes(lastSequence, 10000);
+                updateIndex(indexName, fieldNames, changes, lastSequence);
+                lastSequence = changes.getLastSequence();
+            } while (changes.size() > 0);
+        } catch (DocumentStoreException e) {
+            String message = String.format(Locale.ENGLISH, "Failed to get changes feed from %d", lastSequence);
+            logger.log(Level.SEVERE, message, e);
+            throw new QueryException(e.getCause());
+        }
     }
 
     private void updateIndex(final String indexName,
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/documentstore/DatabaseImplExceptionsTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/documentstore/DatabaseImplExceptionsTest.java
@@ -1,3 +1,17 @@
+/*
+ * Copyright © 2016 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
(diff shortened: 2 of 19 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java
```diff
@@ -108,12 +110,17 @@ private void updateIndex(String indexName, List<FieldSort> fieldNames) throws Qu
...
+        try {
+            do {
+                changes = database.changes(lastSequence, 10000);
+                updateIndex(indexName, fieldNames, changes, lastSequence);
+                lastSequence = changes.getLastSequence();
+            } while (changes.size() > 0);
+        } catch (DocumentStoreException e) {
+            String message = String.format(Locale.ENGLISH, "Failed to get changes feed from %d", lastSequence);
+            logger.log(Level.SEVERE, message, e);
+            throw new QueryException(e.getCause());
```
rhyshort: Shouldn't we just wrap the `DocmentStoreException` rather than the cause here?

tomblench: @rhyshort I think it's worth unpacking the exception for a couple of reasons: making the cause chain shorter and not exposing `DocumentStoreException` which is not relevant to a Query API call.

rhyshort: Well it is relevant, we call down into the datastore code, so if it throws an exception it should come back up in the stack trace, it'll make our lives a lot easier when something goes wrong in this code and the log messages aren't available, but we have the stack.

tomblench: The query code shouldn't call the public API so there should be no `DocumentStoreException` in the first place. In the absence of an even bigger refactor, this is the best we can do. There's no loss if information from skipping out this exception.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/documentstore/DatabaseImplExceptionsTest.java
```diff
@@ -0,0 +1,196 @@
+package com.cloudant.sync.internal.documentstore;
```
brynh: Missing copyright header.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/documentstore/DatabaseImplExceptionsTest.java
```diff
@@ -0,0 +1,196 @@
...
+        this.datastore.getDocument("nosuchdocument");
+    }
+
+    // getDocument on non-existent revid should throw DocumentNotFoundException
+    @Test(expected = DocumentNotFoundException.class)
+    public void getDocumentWithRevisionShouldThrowDocumentNotFoundException() throws Exception {
+        DocumentRevision dr = new DocumentRevision("doc1");
+        dr.setBody(DocumentBodyFactory.create("{\"hello\":\"world\"}".getBytes()));
+        this.datastore.createDocumentFromRevision(dr);
+        this.datastore.getDocument("nosuchdocument", "nosuchrevid");
```
brynh: Shouldn't we use the valid doc id `doc1` here instead of `nosuchdocument` or the test might be passing because of an invalid doc id rather than an invalid rev id.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/documentstore/DatabaseImplExceptionsTest.java
```diff
@@ -0,0 +1,196 @@
...
+        DocumentRevision dr = new DocumentRevision("doc1");
+        dr.setBody(DocumentBodyFactory.create("{\"hello\":\"world\"}".getBytes()));
+        dr = this.datastore.createDocumentFromRevision(dr);
+        // remove the underlying database, triggering a SQL Exception
+        dropRevs();
+        dr.setBody(DocumentBodyFactory.create("{\"hello\":\"updated\"}".getBytes()));
+        this.datastore.updateDocumentFromRevision(dr);
+    }
+
+    // updateDocumentFromRevision after we remove the underlying SQL database should throw
```
brynh: Typo - `deleteDocumentFromRevision`

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/documentstore/DatabaseImplExceptionsTest.java
```diff
@@ -0,0 +1,196 @@
...
+    public void deleteDocumentFromRevisionShouldThrowDocumentStoreException() throws Exception {
+        DocumentRevision dr = new DocumentRevision("doc1");
+        dr.setBody(DocumentBodyFactory.create("{\"hello\":\"world\"}".getBytes()));
+        dr = this.datastore.createDocumentFromRevision(dr);
+        // remove the underlying database, triggering a SQL Exception
+        dropRevs();
+        this.datastore.deleteDocumentFromRevision(dr);
+    }
+
+    // updateDocumentFromRevision after we remove the underlying SQL database should throw
```
brynh: Typo `deleteDocument`

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/query/IndexUpdaterTest.java
```diff
@@ -205,7 +205,7 @@ public void run() {
                     rev.setBody(DocumentBodyFactory.create(bodyMap));
                     try {
                         ds.createDocumentFromRevision(rev);
-                    } catch (DocumentException de) {
+                    } catch (Exception de) {
```
brynh: Although it doesn't really matter, it might be worth changing variable `de` to `e` just to save anyone wondering in what `de` means in future.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/IndexUpdater.java
brynh: It might be worth adding the message to the exception here.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/documentstore/DatabaseImplExceptionsTest.java
```diff
@@ -166,7 +166,7 @@ public void deleteDocumentFromRevisionShouldThrowDocumentStoreException() throws
         this.datastore.deleteDocumentFromRevision(dr);
     }
 
-    // updateDocumentFromRevision after we remove the underlying SQL database should throw
+    // deleteDocumentFromRevision after we remove the underlying SQL database should throw
```
brynh: This one's still wrong, it should be `deleteDocument`, not `deleteDocumentFromRevision`

Issue #389, opened 2016-11-11
Accuracy of Floating point comparisons in Query.


rhyshort:
Query can match on unindexed documents, using the UnindexedMatcher. This matcher compares `double` for all number types. Although this triggers a warning in findbugs, this could be a much bigger problem for example: `compareEq(new Double(1.01), new Float(1.01))` returns `false` when it should be `true`. 

There needs to be a full investigation into the number types used for query. From the SQL matching (and what is returned from SQL) to the java UnindexedMatcher. 

For now as part of #388 the warning will be disabled where required and it should be removed as part of this work if it is possible to do so.

tomblench:
QueryValidator.java explicitly prohibits the use of a float in queries:
```java
    private static boolean validateNotAFloat(Object value) {
        boolean valid = true;

        if (value instanceof Float) {
            valid = false;
            logger.log(Level.SEVERE, "Float value found in query: %f - Use Double instead.", value);
        }

        return valid;
    }
```

tomblench:
As stated `QueryValidator.java` prohibits use of float and we now have regression tests in place to codify this expectation, as done in #447

PR #423, opened 2016-12-05
Add some tests for numeric equality

tomblench:
See #389 

_What_: Add some tests for numeric equality using different combinations of int, float, double

_Why_: To prevent regressions and codify our expectations of the query API for numeric values


ricellis:
@tomblench needs a rebase

tomblench:
@ricellis see #447

Issue #424, opened 2016-12-05
Throw exceptions in Query/Index API


tomblench:

tomblench:
Done in #425

PR #425, opened 2016-12-06
Query throws exceptions on errors

tomblench:
See #424

_What_: Throw appropriate exceptions when error conditions are encountered in Query code.

Instances where `return null` signals an error condition have been replaced by throwing a `QueryException`. Where pre-conditions need to be checked, invoke the appropriate `Misc` checks resulting in `{IllegalArgument,IllegalState,NullPointer}Exception`s.

_Why_: Currently API clients can't distinguish between different error cases or between errors and where a null value is a valid result.

_Testing_: Modify tests which currently check for null on error condition to check for expected exceptions.

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/QueryValidator.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/QueryValidator.java
@@ -498,22 +498,30 @@ class QueryValidator {
                 //  - $in is the only operator that expects a List
                 if (operator.equals(NOT)) {
                     if (!(clauseOperand instanceof Map)) {
-                        throw new QueryException("clauseOperand not instance of Map");
+                        throw new QueryException(String.format(Locale.ENGLISH,
+                                "clauseOperand %s is not an instance of Map",
+                                clauseOperand));
                     }
                     validateClause((Map) clauseOperand);
                 } else if (operator.equals(IN)) {
                     if (!(clauseOperand instanceof List)) {
-                        throw new QueryException("clauseOperand not instance of List");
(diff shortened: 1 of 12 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/QueryValidator.java
```diff
@@ -607,46 +595,29 @@ private static boolean validateModArgument(Object modulus) {
...
     }
 
-    private static boolean validateTextSearchArgument(Object textSearch) {
-        boolean valid = true;
-
+    private static void validateTextSearchArgument(Object textSearch) throws QueryException {
         if (!(textSearch instanceof String)) {
-            valid = false;
-            logger.log(Level.SEVERE, "$search operator expects a String");
+            throw new QueryException("$search operator expects a String");
```
rhyshort: "expects a String" -> "requires a String operand"

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/query/QueryValidatorTest.java
```diff
@@ -373,50 +380,50 @@ public void checkForInvalidValues() {
         assertThat(QueryValidator.normaliseAndValidateQuery(query), is(notNullValue()));
```
rhyshort: Missed this assert.

rhyshort: As per slack, its fine to leave this here.

tomblench:
@rhyshort all issues should be addressed in 54761fd4c6c65859de123fbc8c345ea081ab582b

Issue #413, opened 2016-11-22
v2.0 API discussion points


tomblench:
Issue to track discussion points where we don't yet have a clear answer.

We should strike these off and/or open separate issues as and when we resolve what the next steps are for each.

* ~~should DocumentStore query/database be methods instead of fields~~
  we have resolved that the will be methods - this has been merged
* ~~naming proposal for Database methods (create, delete) instead of createDocumentFromRevision~~
  we have agreed on a set of names which are based on CRUD - this has been merged
* ~~`Database` is in documentstore package, but `Query` isn’t but they are peers~~
  we have decided to keep this as-is
* ~~double check API classes and decide if anything should not be public~~
  done, see https://github.com/cloudant/sync-android/issues/413#issuecomment-275451057
* ~~argument checks currently mostly throw `IllegalArgumentException` except `checkNotNull` which throws a `NullPointerException` should we always throw an `IllegalArgumentException`? Perhaps we should consider a checked `Exception` type?~~
  done, see https://github.com/cloudant/sync-android/issues/413#issuecomment-275680032

rhyshort:
With my arbitrary comparison 

```java
documentStore.database.create(revision);
documentStore.query.find(selector);

documentStore.database().create(revision);
documentStore.query().find(selector);
```

And after seeing our Mocks removed in #412, I think having `database` and `query` as methods (without the get prefix) is a nice compromise and allows us to do mocking without doing potential unsafe things to final fields via reflection. 

So my vote is for `.database()` and `.query()`.

ricellis:
Should discuss whether these should be public:

- [ ] `com.cloudant.sync.documentstore.DocumentBodyFactory#EMPTY` -> should be `DocumentBody` not `DocumentBodyImpl`
- [ ] `com.cloudant.sync.documentstore.Changes#Changes` (marked API private), but is in the public package
- [ ] `com.cloudant.sync.documentstore.Changes#getResults` returns a list of `InternalDocumentRevision`
- [ ] Naming of Id vs ID in many places
- [ ] `com.cloudant.sync.documentstore.DocumentRevision#isCurrent` - I think this is probably needed, but might be worth a conversation.
- [ ] `com.cloudant.sync.documentstore.encryption` package I think only `KeyProvider` and `SimpleKeyProvider` need be public?
- [ ] `com.cloudant.sync.query.IndexType#enumValue` is this needed for case insensitivity? All `enum` types have a `valueOf(String name)` method built-in.
[...]

PR #426, opened 2016-12-07
Obtain query and database through getters

tomblench:
Per discussion in #413 

- Change final fields `query` and `database` to getters with the same name. We are not using the javabeans `getX` convention for brevity.

- javadoc comments are provisional

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/DocumentStore.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/DocumentStore.java
@@ -48,12 +48,12 @@ import java.util.logging.Logger;
  *
  * <p>
  * Users can perform CRUD (create, read, update, delete) operations on JSON documents and attachments
- * by invoking methods on the {@link #database} member.
+ * by invoking methods through {@link #database()}.
  * </p>
  *
  * <p>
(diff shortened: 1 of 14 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/DocumentStore.java
```diff
@@ -182,6 +173,30 @@ public static DocumentStore getInstance(File location, KeyProvider provider) thr
         }
     }
 
+    /**
+     * Get a reference to the {@link Database} object.
+     *
```
ricellis: These comments need `<p>` tags if we want this paragraphing to appear in the generated doc.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/DocumentStore.java
```diff
@@ -182,6 +173,30 @@ public static DocumentStore getInstance(File location, KeyProvider provider) thr
...
+     */
+    public Database database() {
+        return database;
+    }
+
+    /**
+     * Get a reference to the {@link Query} object.
+     *
+     * Users can perform index and query operations on JSON documents by invoking methods on this
+     * object
```
ricellis: Again `<p>` for paragraphing.
Needs a `.` after object.

PR #427, opened 2016-12-07
Restore mocks in ReplicatorIdTest

tomblench:
Since merging #426 we can bring back the use of mocks in this test.


tomblench:
See #432 instead

PR #428, opened 2016-12-07
Simplify `Database` method names

tomblench:
As per #413 "naming proposal for Database methods (create, delete) instead of createDocumentFromRevision"

Should we also rename others: "getEventBus()" → "eventBus" and "getPath" → "getPath"?


rhyshort:
@tomblench I'm confused about `getPath` to `getPath` did you miss some characters?

rhyshort:
I'm not keen on `getAll` and `getAllIDs` but I can't really think of anything better all I have so far I've come up with `listDocuments` and `listIDs`.

tomblench:
Sorry, I meant getPath → path

tomblench:
I'm not a fan of `listDocuments` as "list" implies only getting some kind of primary key and not the whole document. I would be happier to change this to `get` and have the method signature and javadoc do the work of conveying to the user that it gets multiple documents based on certain criteria.

Not at all sure about the IDs one, maybe `getIds`?

tomblench:
Agreed

PR #429, opened 2016-12-08
Remove RuntimeException in ReplicatorBuilder

tomblench:
See #280 

## What 
- Remove unchecked exception when redacting URL for cookie interceptor
- Use `Misc.checkArgument` when checking default port
- Use `Misc.checkState` when checking both source and target are non-null

## Why
- As outlined in the parent issue, we are over-reliant on `RuntimeException`s which are often not appropriate

## Tests
- See new tests in `{Pull,Push}ReplicatorTest`


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
```diff
@@ -56,11 +58,13 @@ private URI addCookieInterceptorIfRequired(URI uri) {
...
             } else if ("https".equals(uriProtocol)) {
                 uriPort = 443;
-            } else {
-                throw new RuntimeException("Unknown protocol: "+uriProtocol);
             }
         }
 
+        Misc.checkArgument(uriPort >= 0,
+                String.format(Locale.ENGLISH,
+                        "Could not determine default port for protocol \"%s\"", uriProtocol));
```
rhyshort: It seems a bit odd that the IllegalArgumentException would be thrown when a port couldn't be determined. Unsupported protocol message would make more sense.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
```diff
@@ -70,18 +74,14 @@ private URI addCookieInterceptorIfRequired(URI uri) {
...
-            throw new RuntimeException(e);
-        }
+        //Remove user credentials from url
+        URI redacted = new URI(uriProtocol
+                + "://"
+                + uriHost
+                + ":"
+                + uriPort
+                + (uriPath != null ? uriPath : ""));
+        return redacted;
```
rhyshort: No need for "redacted" just return the `new URI`

PR #431, opened 2016-12-09
Throw QueryException instead of RuntimeException

rhyshort:
Throw QueryException instead of RuntimeException for listIndexes.

Part of #280

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/QueryImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/QueryImpl.java
@@ -16,6 +16,7 @@ package com.cloudant.sync.internal.query;
 
 import com.cloudant.sync.documentstore.Database;
 import com.cloudant.sync.documentstore.encryption.KeyProvider;
+import com.cloudant.sync.internal.documentstore.DatabaseImpl;
 import com.cloudant.sync.internal.documentstore.migrations.SchemaOnlyMigration;
 import com.cloudant.sync.query.FieldSort;
 import com.cloudant.sync.query.Index;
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/QueryImpl.java
```diff
@@ -137,15 +137,17 @@ public void close() {
...
      */
     @Override
-    public List<Index> listIndexes() {
+    public List<Index> listIndexes() throws QueryException {
         try {
             return dbQueue.submit(new ListIndexesCallable()).get();
         } catch (InterruptedException e) {
-            logger.log(Level.SEVERE,"Failed to list indexes",e);
-            throw new RuntimeException(e);
+            String msg = "Failed to list indexes";
```
emlaver: Thoughts on putting log messages in a Utils class?

rhyshort: I don't think we should, it's nice to have the the messages strings right next to the code where we log and throw. Admittedly the duplication is annoying, but it is something that can go away in future when we can use Java 7 language features.

tomblench: You could consider cribbing `DatastoreImpl.get` which I wrote to help us avoid this repeating pattern:

```java

    // helper to avoid having to catch ExecutionExceptions
    public static <T> T get(Future<T> future) throws ExecutionException {
        try {
            return future.get();
            logger.log(Level.SEVERE, "Re-throwing InterruptedException as ExecutionException");
            throw new ExecutionException(e);
        }
    }

```

PR #437, opened 2016-12-12
Fix sample app for v2 API

tomblench:
_What_: Fix sample app so it builds and works on new v2 API

_Why_: As well as being a pre-requisite for merging the feature branch to master, this is a useful exercise to see what the API looks like from a user's point of view

_Testing_: App builds and runs correctly

### DIFF
--- a/sample/todo-sync/src/com/cloudant/todo/TasksModel.java
+++ b/sample/todo-sync/src/com/cloudant/todo/TasksModel.java
@@ -43,14 +43,14 @@ import java.util.List;
 
 
 /**
- * <p>Handles dealing with the datastore and replication.</p>
+ * <p>Handles dealing with the DocumentStore and replication.</p>
  */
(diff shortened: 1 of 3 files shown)


sample/todo-sync/src/com/cloudant/todo/TasksModel.java
```diff
@@ -51,7 +52,7 @@
     private static final String DATASTORE_MANGER_DIR = "data";
```
brynh: For consistency with the new naming we should probably change `DATASTORE` to `DOCUMENT_STORE` for this and other occurrences of `Datastore` in this file. 

Also, we should probably fix the typo `MANGER`.

tomblench: Thanks for the feedback, I've fixed all the other incorrect use of datastore in comments etc.

PR #441, opened 2016-12-13
Enforce Event bus restrictions at register time.

rhyshort:
## What
Enforce event bus restrictions when a subscriber is being registered to the bus.

## Why 
Enforcing the event bus restrictions when registering means the `IllegalAccessException` thrown
when attempting to invoke a method on the subscriber should not occur during normal operation, so it becomes safe to throw a `RuntimeException` when an `IllegalAccessException` is caught.

## How
Make sure subscribing instances conform to the following restrictions when the register method is
called or throw an `IllegalArgumentException`.

- Subscriber class *must* be public
- Methods with the @Subscribe annotation *must* be public.
- Method with the @Subscribe annotation *must* have only one parameter.

Note: Previously the method and subscriber only needed to be visible to the EventBus class, this
has now been restricted to enforce public access for the subscription method and subscription class. This makes it easier to perform validation if

## Testing

New tests added to the EventBusTest which test exception cases.

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
@@ -14,10 +14,14 @@
 
 package com.cloudant.sync.event;
 
+import com.cloudant.sync.event.notifications.Notification;
+
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
@@ -46,7 +50,7 @@ public class EventBus {
      *
      * @param event to post to subscribers
      */
-    public void post(Object event) {
+    public void post(Notification event) {
         for (Map.Entry<Object, List<SubscriberMethod>> entry : listeners.entrySet()) {
             for (SubscriberMethod method : entry.getValue()) {
                 if (method.eventTypeToInvokeOn.isInstance(event)) {
@@ -74,11 +78,28 @@ public class EventBus {
      * {@link com.cloudant.sync.event.Subscribe}.
      * </p>
      *
+     * <p>
+     *     The following restrictions are placed on the subscriber instance.
+     * </p>
+     * <ul>
+     *    <li>The class of the subscriber <strong>must</strong> be public</li>
+     *    <li>Methods annotated with <code>@Subscribe</code> <strong>must</strong> be public</li>
+     *    <li>Methods annotated with <code>@Subscribe</code> <strong>must</strong> have exactly <strong>one</strong> parameter</li>
+     * </ul>
+     *
      * @param object the instance that will be notified when an event is posted
+     * @throws IllegalArgumentException if the subscribing class does not adhere to the restrictions listed.
+     *
      */
     public void register(Object object) {
 
         Class<?> listenerClass = object.getClass();
+
+        if(!Modifier.isPublic(listenerClass.getModifiers())){
+            throw new IllegalArgumentException(String.format(Locale.ENGLISH,
+                    "Event Subscriber class %s needs to be public",
+                    listenerClass.getName()));
+        }
 
         List<SubscriberMethod> methods = new ArrayList<SubscriberMethod>();
         // Do loop to traverse the class hierarchy looking for @Subscribe methods. Using
@@ -88,9 +109,28 @@ public class EventBus {
         do {
             for (Method m : listenerClass.getDeclaredMethods()) {
                 if (m.isAnnotationPresent(com.cloudant.sync.event.Subscribe.class)) {
+
+                    if (!Modifier.isPublic(m.getModifiers())){
+                        throw new IllegalArgumentException(String.format(Locale.ENGLISH,
+                                "Subscriber method %s#%s is required to be public",
+                                listenerClass.getName(), m.getName()));
+                    }
+
                     Class[] params = m.getParameterTypes();
                     if (params.length == 1) {
+                        if(!Notification.class.isAssignableFrom(params[0])){
+                            throw new IllegalArgumentException(String.format(Locale.ENGLISH,
+                                    "Subscriber method %s#%s parameter is required to be assignable %s",
+                                    listenerClass.getName(),
+                                    m.getName(),
+                                    Notification.class.getCanonicalName()));
+                        }
                         methods.add(new SubscriberMethod(m, params[0]));
+                    } else {
+                        throw new IllegalArgumentException(String.format(Locale.ENGLISH,
+                                "Subscriber method  %s#%s is required to have only 1 parameter",
+                                listenerClass.getName(),
+                                m.getName()));
                     }
                 }
             }
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/notifications/Notification.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/notifications/Notification.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2015 Cloudant, Inc. All rights reserved.
+ * Copyright © 2016 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
@@ -10,18 +10,13 @@
  * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
  * either express or implied. See the License for the specific language governing permissions
  * and limitations under the License.
+ *
  */
 
-package com.cloudant.sync.internal.replication;
+package com.cloudant.sync.event.notifications;
 
-import com.cloudant.sync.event.notifications.Notification;
-
-class ReplicationStrategyCompleted implements Notification {
-
-    protected ReplicationStrategyCompleted(ReplicationStrategy replicationStrategy) {
-        this.replicationStrategy = replicationStrategy;
-    }
-
-    protected final ReplicationStrategy replicationStrategy;
-    
+/**
+ * Marker interface for events.
+ */
+public interface Notification {
 }
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
@@ -18,38 +18,23 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
 
+import com.cloudant.sync.event.notifications.Notification;
+
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 import java.util.ArrayList;
 import java.util.List;
 
 public class EventBusTest {
 
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
     public EventBus bus;
     public TestBusSubscriber sub;
-
-    private static class TestEvent {
-
-    }
-
-    private static class ExtendedTestEvent extends TestEvent {
-
-    }
-
-    private static class TestBusSubscriber {
-
-        List<Object> eventsReceived = new ArrayList<Object>();
-
-        protected void onEvent(Object event) {
-            eventsReceived.add(event);
-        }
-
-        @Subscribe
-        public void onEvent(TestEvent event) {
-            onEvent((Object) event);
-        }
-    }
 
     @Before
     public void newBus() throws Exception {
@@ -84,7 +69,8 @@ public class EventBusTest {
         bus.register(sub);
 
         // Post an event (of type Object) - we are subscribed to TestEvent
-        Object e = new Object();
+        Notification e = new Notification() {
+        };
         bus.post(e);
 
         assertTrue("There should be no events", sub.eventsReceived.isEmpty());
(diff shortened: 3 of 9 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
```diff
@@ -88,9 +105,16 @@ public void register(Object object) {
         do {
             for (Method m : listenerClass.getDeclaredMethods()) {
                 if (m.isAnnotationPresent(com.cloudant.sync.event.Subscribe.class)) {
+
+                    if (!Modifier.isPublic(m.getModifiers())){
+                        throw new IllegalArgumentException("Subscriber method with @Subscribe is required to be public");
```
tomblench: Exception message should name the method which is at fault (fully qualified)

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
```diff
@@ -88,9 +105,16 @@ public void register(Object object) {
...
+                    }
+
                     Class[] params = m.getParameterTypes();
                     if (params.length == 1) {
                         methods.add(new SubscriberMethod(m, params[0]));
+                    } else {
+                        throw new IllegalArgumentException("Subscriber method is required to have only 1 parameter");
```
tomblench: As above

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
```diff
@@ -19,13 +19,18 @@
...
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 import java.util.ArrayList;
 import java.util.List;
 
 public class EventBusTest {
 
+    @Rule
+    public ExpectedException thrown= ExpectedException.none();
```
ricellis: spacing around `=`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
```diff
@@ -74,12 +75,28 @@ public void post(Object event) {
...
      *
+     * <p>
+     *     The following restrictions are placed on the subscriber instance.
+     *     <ul>
+     *         <li>The class of the subscriber <strong>must</strong> be public</li>
+     *         <li>Methods annotated with <code>@Subscribe</code> <strong>must</strong> be public</li>
+     *         <li>Methods annotated with <code>@Subscribe</code> <strong>must</strong> have exactly <strong>one</strong> parameter</li>
+     *     </ul>
+     *
+     * </p>
```
ricellis: This end tag is what is breaking the javadoc build.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
```diff
@@ -162,19 +167,178 @@ public void multipleEvents() throws Exception {
...
+
+    @Test
+    public  void privateSubscriber() throws Exception {
+        thrown.expect(IllegalArgumentException.class);
+        thrown.expectMessage("Event Subscriber class needs to be public");
+
+        bus.register(new PrivateSubscriber());
+
+        // Post an event
+        TestEvent e = new TestEvent();
```
tomblench: None of these error cases need to post events

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
```diff
@@ -162,19 +167,178 @@ public void multipleEvents() throws Exception {
...
+
+        @Subscribe
+        public void throwOnEvent(TestEvent event) throws Exception {
+            throw new Exception("Test subscriber exception");
+        }
+    }
+
+
+    @Test
+    public  void privateSubscriber() throws Exception {
```
tomblench: A comment per test like the existing ones in this package would be helpful

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
```diff
@@ -162,19 +167,178 @@ public void multipleEvents() throws Exception {
...
+
+    }
+
+    protected static class ProtectedSubScriber{
+    }
+
+    static class PackageProtectedSubscriber {
+    }
+
+    public static class PakageProtectedMethodSubscriber {
```
tomblench: Pakage → Package

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
ricellis: Should we consider doing additional checks here that the parameter is one of the notification classes. I believe they are all in the same package now so we could check that maybe?

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
```diff
@@ -162,19 +167,178 @@ public void multipleEvents() throws Exception {
...
+        thrown.expectMessage("Subscriber method is required to have only 1 parameter");
+
+        bus.register(new SubscriberMethod0Params());
+
+        // Post an event
+        TestEvent e = new TestEvent();
+        bus.post(e);
+    }
+
+    public static class SubscriberMethod2Params{
```
tomblench: Can we put all of these statics in one place? There are some existing ones at the top and one in the middle

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
```diff
@@ -162,19 +167,178 @@ public void multipleEvents() throws Exception {
...
+
+
+    @Test
+    public  void privateSubscriber() throws Exception {
+        thrown.expect(IllegalArgumentException.class);
+        thrown.expectMessage("Event Subscriber class needs to be public");
+
+        bus.register(new PrivateSubscriber());
+
+        // Post an event
```
ricellis: Remove, this isn't necessary when testing the exception.

ricellis: And for the other exception test cases.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
```diff
@@ -74,12 +75,28 @@ public void post(Object event) {
...
      * @param object the instance that will be notified when an event is posted
+     * @throws IllegalArgumentException if the subscribing class does not adhere to the restrictions listed.
+     *
      */
     public void register(Object object) {
 
         Class<?> listenerClass = object.getClass();
 
+        if(!Modifier.isPublic(listenerClass.getModifiers())){
+            throw new IllegalArgumentException("Event Subscriber class needs to be public");
```
tomblench: Exception message should name the class which is at fault (fully qualified)

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
ricellis: Possibly this leads back to the marker interface I suggested when we made these classes for 1.0. Then we can just do an assignable check.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
rhyshort: Done in af4d3d8

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
rhyshort: Done in af4d3d8

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
rhyshort: Done in af4d3d8

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
rhyshort: fixed in ffc25b7

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
rhyshort: removed in 98551a3

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
rhyshort: fixed in 7910d5d

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
rhyshort: removed in 98551a3

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
rhyshort: moved to end of file in 20d42da

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/event/EventBusTest.java
rhyshort: fixed in 3290b76

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
```diff
@@ -74,12 +76,29 @@ public void post(Object event) {
      * {@link com.cloudant.sync.event.Subscribe}.
      * </p>
      *
+     * <p>
```
tomblench: Small nitpick but the tags aren't indented properly which makes this hard to follow

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
```diff
@@ -88,9 +107,21 @@ public void register(Object object) {
         do {
             for (Method m : listenerClass.getDeclaredMethods()) {
                 if (m.isAnnotationPresent(com.cloudant.sync.event.Subscribe.class)) {
+
+                    if (!Modifier.isPublic(m.getModifiers())){
+                        throw new IllegalArgumentException(String.format(Locale.ENGLISH,
+                                "Subscriber method %s#%s is required to be public",
```
tomblench: Can the class be anonymous here as well?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
```diff
@@ -88,9 +107,21 @@ public void register(Object object) {
...
+                                listenerClass.getCanonicalName(), m.getName()));
+                    }
+
                     Class[] params = m.getParameterTypes();
                     if (params.length == 1) {
                         methods.add(new SubscriberMethod(m, params[0]));
+                    } else {
+                        throw new IllegalArgumentException(String.format(Locale.ENGLISH,
+                                "Subscriber method  %s#%s is required to have only 1 parameter",
```
tomblench: As above

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
rhyshort: No they can't. Anonymous will always be rejected by the first check because they cannot be made public.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
rhyshort: Fixed in c0df327

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/notifications/Notification.java
```diff
@@ -0,0 +1,22 @@
...
+ * and limitations under the License.
+ *
+ */
+
+package com.cloudant.sync.event.notifications;
+
+/**
+ * Marker interface for events.
+ */
+public interface Notification {
```
ricellis: is it weird to have `Notification`s instead of `Event`s on an `EventBus`? In either case the doc should be updated to say what type the parameter of a `@Subscribe` annotated method should be.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
```diff
@@ -74,12 +78,29 @@ public void post(Object event) {
...
+     *
      */
     public void register(Object object) {
 
         Class<?> listenerClass = object.getClass();
 
+        if(!Modifier.isPublic(listenerClass.getModifiers())){
+            throw new IllegalArgumentException(String.format(Locale.ENGLISH,
+                    "Event Subscriber class %s needs to be public",
+                    listenerClass.isAnonymousClass() ? "Anonymous" : listenerClass.getCanonicalName()));
```
ricellis: Why not just use `listenerClass.getName()` for these? It would avoid the need to do the anonymous check because an anonymous class will be returned with a `$n` reference which will make it easier for people to identify the offending class.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/notifications/Notification.java
rhyshort: I thought it made more sense to be `Notification` rather than event, because its in the `notifications` package.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/EventBus.java
rhyshort: done in 18c3fa7

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/event/notifications/Notification.java
ricellis: Well I feel like an `EventBus` should have `Event`s, but there are many other `Event` classes so it is probably less confusing to have a more unique name. Given that the package is notifications, probably best to stick with `Notification`.

ricellis:
@rhyshort I think the 1 remaining change on this is to add the `Notification` type to the `event.md` documentation.

rhyshort:
@ricellis done in 50f4317

PR #442, opened 2016-12-13
Throw IllegalArgumentException instead of RuntimeException.

rhyshort:
Throw IllegalArgumentException instead of RuntimeException in DatabaseUtils because this class makes more sense than a plain RuntimeException.


ricellis:
Whilst this is an improvement `Misc.checkArgument` throwing an `IllegalArgumentException` could still crash an application. In #280 we talked about some defensive checks to ensure that we didn't pass anything too short into this, are you confident that the internal callers of `makePlaceholders` are behaving sensibly?

rhyshort:
Ah forgot about that. From a quick check, there is at least one method which may misbehave. `read(final List<String> docIds)` doesn't check the length of the docIds passed in. Will update for these cases.

ricellis:
Yeah if you need to add additional checks at the API level don't forget to udpate documentation with e.g. `@throws IllegalArgumentException if the docIds list is null or empty`

rhyshort:
@ricellis I think I got the right in ff4c1c0

rhyshort:
@emlaver still +1?

emlaver:
Yes, +1

PR #445, opened 2016-12-14
Throw DocumentStoreException when revsDiff fails.

### LINKED ISSUE #444: DatabaseImpl#revsDiff returns null on error
DatabaseImpl#revsDiff returns null on error, this should throw a checked exception.

rhyshort:
Throw DocumentStoreException when revsDiff queue task throws an ExecutionException.

Fixes #444

### DIFF
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/documentstore/DatabaseImplRevsDiffTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/documentstore/DatabaseImplRevsDiffTest.java
@@ -31,7 +31,8 @@ public class DatabaseImplRevsDiffTest extends BasicDatastoreTestBase{
     public ExpectedException thrown = ExpectedException.none();
 
     @Test
-    public void revsDiff_emptyInput_returnEmpty() {
+    public void revsDiff_emptyInput_returnEmpty() throws Exception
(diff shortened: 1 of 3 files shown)


cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/documentstore/DatabaseImplRevsDiffTest.java
```diff
@@ -26,7 +26,7 @@
 public class DatabaseImplRevsDiffTest extends BasicDatastoreTestBase{
 
     @Test
-    public void revsDiff_emptyInput_returnEmpty() {
+    public void revsDiff_emptyInput_returnEmpty() throws Exception {
```
emlaver: Does this need to be `Exception` or can `DocumentStoreException` meet the requirement?

rhyshort: It could be `DocumentStoreException` but with it only being test code I didn't think it mattered too much.

emlaver: Ok, that's true, up to you then if you want to change it.

PR #447, opened 2016-12-15
Add some tests for numeric equality

tomblench:
See #389 

_What_: Add some tests for numeric equality using different combinations of int, float, double

_Why_: To prevent regressions and codify our expectations of the query API for numeric values

### DIFF
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/query/NumericQueryTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/query/NumericQueryTest.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright © 2016 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package com.cloudant.sync.internal.query;
+
+import com.cloudant.sync.documentstore.DocumentBodyFactory;
+import com.cloudant.sync.documentstore.DocumentRevision;
+import com.cloudant.sync.query.FieldSort;
+import com.cloudant.sync.query.QueryException;
+import com.cloudant.sync.query.QueryResult;
+
+import org.junit.After;
+import org.junit.Assert;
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/query/NumericQueryTest.java
```diff
@@ -0,0 +1,107 @@
...
+        ds.create(documentRevision);
+        List<FieldSort> fs = new ArrayList<FieldSort>();
+        fs.add(new FieldSort("integer"));
+        fs.add(new FieldSort("float"));
+        fs.add(new FieldSort("double"));
+        im.ensureIndexed(fs);
+    }
+
+    @After
+    public void after() {
```
ricellis: Any reason not to delete this method? It appears to be a no-op.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/query/NumericQueryTest.java
```diff
@@ -0,0 +1,107 @@
...
+    // query two doubles are equal to each other
+    @Test
+    public void testDoubleEqual() throws QueryException {
+        Map<String, Object> q = new HashMap<String, Object>();
+        q.put("double", defaultDouble);
+        QueryResult qr = im.find(q);
+        Assert.assertEquals(1, qr.size());
+    }
+
+}
```
ricellis: Add the newline.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/query/NumericQueryTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+package com.cloudant.sync.internal.query;
+
+import com.cloudant.sync.documentstore.DocumentBodyFactory;
+import com.cloudant.sync.documentstore.DocumentRevision;
+import com.cloudant.sync.query.FieldSort;
+import com.cloudant.sync.query.QueryException;
+import com.cloudant.sync.query.QueryResult;
+
+import org.junit.After;
```
brynh: We can remove this import.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/query/NumericQueryTest.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class NumericQueryTest extends AbstractQueryTestBase {
+
+    int defaultInt = 42;
+    float defaultFloat = 1.01f;
+    double defaultDouble = 1.01d;
```
brynh: I'd prefer these three default values to be `final` so it's clearer they don't change and I think they should probably be `private` and `static` too.

PR #450, opened 2016-12-19
Stopped randomly generating content for PullReplicationTerminationTest

ricellis:
*What*

Stopped randomly generating content for `PullReplicationTerminationTest`

*Why*

Some bytes are invalid UTF-8 and using randomly generated content appears to cause intermittent test failures.

*How*

Changed to use content of A-z,0-9 instead of random bytes.
Generated doc list once up front instead of once per test.

*Testing*

Updated `PullReplicationTerminationTest` as per above. All tests passing on Travis, Jenkins and in the matrix builds on Android where failures were appearing more regularly.

### DIFF
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/PullReplicationTerminationTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/replication/PullReplicationTerminationTest.java
@@ -20,12 +20,12 @@ import com.cloudant.http.HttpConnectionRequestInterceptor;
 
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


emlaver:
@ricellis FYI I've been monitoring the test database for python tests and I see a database named `pull_replication_termination_test14821662152488799080328274273` left behind.

ricellis:
Thanks @emlaver 
@rhyshort is this the database you were using for manual testing?

rhyshort:
I don't think that was me. I've been using local on 002 / Couch 2 on my box.

ricellis:
AFAICT if the test did issue the delete request it didn't reach the server. My changes didn't alter the DB teardown so I don't think it is related to this ticket. The build that made the request was deleted with the branch so I can't recover anything from the logs that might show a failure mode of test teardown, although it looks like we suppress the exceptions anyway.
I've deleted it manually and we'll look into it further if it happens again.

emlaver:
Ok, sounds good to me.

PR #451, opened 2016-12-21
New Credentials API for ReplicatorBuilder

### LINKED ISSUE #440: ReplicatorBuilder API for Passing Creds
One of the problems with the current API as spotted in CDTDatastore https://github.com/cloudant/CDTDatastore/issues/302 and resolved in https://github.com/cloudant/CDTDatastore/pull/306, is that is it difficult to pass usernames and passwords for authentication with the server if they contain special characters. 

Adding something similar to sync-android will make it easier for users to pass creds to the replicator.

I propose something like:

```java
Replicator replicator = ReplicatorBuilder.pull
    .from(SERVER_URI)
    .to(ds)
    .username("example") 
    .password("password")
    .build();
```

The provided credentials will be passed directly to the cookie interceptor to be used for authentication.

Ideally (if others agree this is a worth while addition) I think it should ship as part of 2.0.0.

rhyshort:
Added a new credentials API to the ReplicatorBuilder class, to make it
easier to pass credentials to the replicator. The new methods `username`
and `password` do not require escaping of URL special characters, which
greatly improves the usability of the replicator with complex passwords.

Resolves #440

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2015 IBM Corp. All rights reserved.
+ * Copyright © 2015, 2017 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
@@ -45,6 +45,9 @@ public abstract class ReplicatorBuilder<S, T, E> {
 
     private T target;
     private S source;
+    private String username;
+    private String password;
+
     private int id = ReplicatorImpl.NULL_ID;
     private List<HttpConnectionRequestInterceptor> requestInterceptors = new ArrayList
             <HttpConnectionRequestInterceptor>();
@@ -72,33 +75,39 @@ public abstract class ReplicatorBuilder<S, T, E> {
                     "Protocol %s not supported", uriProtocol));
         }
 
-        if (uri.getUserInfo() != null) {
+        if (uri.getUserInfo() != null && this.username == null && this.password == null) {
             String[] parts = uri.getUserInfo().split(":");
             if (parts.length == 2) {
+                this.username = parts[0];
+                this.password = parts[1];
+            }
+        }
 
-                String path = uri.getRawPath();
+        if(this.username == null && this.password == null){
+            return uri;
+        }
+
+
+
+        try {
+            String path = uriPath == null ? "" : uriPath;
+
+            if(path.length() > 0) {
                 int index = path.lastIndexOf("/");
                 if (index == path.length() - 1) {
                     // we need to go back one
                     path = path.substring(0, index);
                     index = path.lastIndexOf("/");
                 }
-
                 path = path.substring(0, index);
-
-                URI baseURI;
-
-                try {
-                    baseURI = new URI(uriProtocol, null, uriHost, uriPort, path, null, null);
-                } catch (URISyntaxException e) {
-                    throw new RuntimeException(e);
-                }
-
-
-                CookieInterceptor ci = new CookieInterceptor(parts[0], parts[1], baseURI.toString());
-                requestInterceptors.add(ci);
-                responseInterceptors.add(ci);
             }
+
+            URI baseURI = new URI(uriProtocol, null, uriHost, uriPort, path, null, null);
+            CookieInterceptor ci = new CookieInterceptor(username, password, baseURI.toString());
+            requestInterceptors.add(ci);
+            responseInterceptors.add(ci);
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
         }
 
         try {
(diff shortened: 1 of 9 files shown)


cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/mazha/CouchConfig.java
```diff
@@ -34,33 +39,98 @@
...
         this.responseInterceptors = responseInterceptors;
+        this.username = username;
+        this.password = password;
+
     }
 
-    public List<HttpConnectionRequestInterceptor> getRequestInterceptors() {
+
+    public List<HttpConnectionRequestInterceptor> getRequestInterceptors(boolean includeCookie) {
+        CookieInterceptor cookieInterceptor = buildCookieInterceptor();
```
ricellis: I'm not greatly in favour of a getter building an interceptor. I know this is test code, but it is pretty confusing to have an API with a setter and a getter, where the getter can do something other than return what was set in the setter.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/mazha/CouchConfig.java
```diff
@@ -34,33 +39,98 @@
...
+            List<HttpConnectionResponseInterceptor> requestInterceptors = new ArrayList
+                    <HttpConnectionResponseInterceptor>();
+            requestInterceptors.addAll(this.responseInterceptors);
+            requestInterceptors.add(cookieInterceptor);
+            return requestInterceptors;
+        }
+        return responseInterceptors;
+    }
+
+    private CookieInterceptor buildCookieInterceptor() {
```
ricellis: Perhaps I'm missing something that isn't immediately obvious, but following on from the above I don't see why this is needed here. You have changed the tests to use the `ReplicatorBuilder` with the `username()` and `password()` methods getting the necessary values from this config class (via `getUsername`, `getPassword`). The `ReplicatorBuilder` handles the building and addition of the `CookieInterceptor` so I think this test code can be greatly simplified by removing this method and the two new overload versions of `getResponseInterceptors`.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
```diff
@@ -372,6 +379,19 @@ public E addResponseInterceptors(List<HttpConnectionResponseInterceptor> interce
         return (E) this;
     }
 
+    public E username(String username) {
```
rhyshort: Oops, fixed in 85e8146 and 85e8146.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
```diff
@@ -372,6 +379,19 @@ public E addResponseInterceptors(List<HttpConnectionResponseInterceptor> interce
...
+        this.username = username;
```
rhyshort: I'm not sure, it doesn't exactly breaks anything if you pass in `null`. It could be useful in some circumstances to allow `null`, but I don't really have a strong opinion either way on this one.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/mazha/CouchConfig.java
rhyshort: Its to make the number of code changes vastly smaller than otherwise. Through out the test code we pull authentication information out of the config, and it is expected that the cookie interceptor will be present. This could be fixed just for the replicator tests be removing all instances of the cookie interceptor in the interceptor lists. But I thought it would be a better decision to not include it from the config, but still provide the authentication mechanism to any of the other tests.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
ricellis: I think it will help prevent potential NPEs down the line (e.g. when trying to encode in the `CookieInterceptor`). I can't think of a valid use case for passing `null` to the method (vs. say just omitting the call) and we are relying on `null` to tell us that username and password have not been set so we should probably "reserve" that state.

rhyshort: done in bece517

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
```diff
@@ -379,12 +379,22 @@ public E addResponseInterceptors(List<HttpConnectionResponseInterceptor> interce
...
+     * @return The current instance of {@link ReplicatorBuilder}
+     */
     public E username(String username) {
         this.username = username;
         //noinspection unchecked
         return (E) this;
     }
 
+    /**
+     * Sets the password to ise when authenticating with the server.
```
ricellis: typo
ise -> use

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/mazha/CouchConfig.java
ricellis: f9d6d74c5f898f40e9ae679176659aef9dca00d7 worked for me locally.
I feel like either I must be missing some nuance or you're over thinking it. Or perhaps it will fail in the CI and demonstrate to me why that approach doesn't work!

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
rhyshort: done in 6453ff2

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
```diff
@@ -72,33 +75,37 @@ private URI addCookieInterceptorIfRequired(URI uri) {
                     "Protocol %s not supported", uriProtocol));
         }
 
-        if (uri.getUserInfo() != null) {
+        if (uri.getUserInfo() != null && this.username == null && this.password == null) {
             String[] parts = uri.getUserInfo().split(":");
-            if (parts.length == 2) {
+            this.username = parts[0];
```
tomblench: Should probably still have the `if (parts.length == 2)` check - otherwise this will throw OOB exception for a URL of the format `http://user@host/abc`

rhyshort: done in ea0f183

PR #452, opened 2017-01-03
Extract remaining Database Callables.

rhyshort:
## What

Extract remaining Database Callables.

## How

- GetPublicIdentifierCallable extracted to its own class from DatabaseImpl.
- Collapse RevsDiff callable into revsDiff method, calling RevsDiffBatch callable after the list of revisions has been partitioned

## Testing 

All existing tests pass

## Issues

Part of #291

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
@@ -805,9 +805,25 @@ public class DatabaseImpl implements Database {
         Misc.checkNotNull(revisions, "Input revisions");
         Misc.checkArgument(!revisions.isEmpty(), "revisions cannot be empty");
 
-        // TODO hoist down queue.submit to just surround the call to RevsDiffBatchCallable
         try {
-            return get(queue.submit(new RevsDiffCallable(revisions)));
+
+            ValueListMap<String, String> missingRevs = new ValueListMap<String, String>();
+
+            // Break down by docId first to avoid potential rev ID clashes between doc IDs
+            for (Map.Entry<String, List<String>> entry : revisions.entrySet()) {
+                String docId = entry.getKey();
+                List<String> revs = entry.getValue();
+                // Partition into batches to avoid exceeding placeholder limit
+                // The doc ID will use one placeholder, so use limit - 1 for the number of
+                // revs for the remaining placeholders.
+                List<List<String>> batches = CollectionUtils.partition(revs,
+                        SQLITE_QUERY_PLACEHOLDERS_LIMIT - 1);
+
+                for (List<String> revsBatch : batches) {
+                    missingRevs.addValuesToKey(docId, get(queue.submit(new RevsDiffBatchCallable(docId, revsBatch))));
+                }
+            }
+            return missingRevs;
         } catch (ExecutionException e) {
             String message = "Failed to calculate difference in revisions";
             logger.log(Level.SEVERE, message, e);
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/callables/GetPublicIdentifierCallable.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    public String call(SQLDatabase db) throws Exception {
+        Cursor cursor = null;
+        try {
+            cursor = db.rawQuery("SELECT value FROM info WHERE key='publicUUID'", null);
+            if (cursor.moveToFirst()) {
+                return "touchdb_" + cursor.getString(0);
+            } else {
+                throw new IllegalStateException("Error querying PublicUUID, " +
+                        "it is probably because the sqlDatabase is not probably " +
```
emlaver: Change to `...sqlDatabase is not properly...`

rhyshort: fixed in 01086d9

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/callables/GetPublicIdentifierCallable.java
```diff
@@ -0,0 +1,43 @@
...
+
+package com.cloudant.sync.internal.documentstore.callables;
+
+import com.cloudant.sync.internal.sqlite.Cursor;
+import com.cloudant.sync.internal.sqlite.SQLCallable;
+import com.cloudant.sync.internal.sqlite.SQLDatabase;
+import com.cloudant.sync.internal.util.DatabaseUtils;
+
+/**
+ * Created by rhys on 22/12/2016.
```
emlaver: Replace auto-generated comment with class docs.

rhyshort: done in e7363ff

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
```diff
@@ -805,9 +805,25 @@ boolean isOpen() {
...
+                String docId = entry.getKey();
+                List<String> revs = entry.getValue();
+                // Partition into batches to avoid exceeding placeholder limit
+                // The doc ID will use one placeholder, so use limit - 1 for the number of
+                // revs for the remaining placeholders.
+                List<List<String>> batches = CollectionUtils.partition(revs,
+                        SQLITE_QUERY_PLACEHOLDERS_LIMIT - 1);
+
+                for (List<String> revsBatch : batches) {
+                    missingRevs.addValuesToKey(docId, get(queue.submit(new RevsDiffBatchCallable(docId, revsBatch))));
```
ricellis: This actually introduces a subtle behaviour change. Currently all the batches are retrieved from the same queue operation. After this change the batches could be interleaved between other operations on the queue that have changed the state of the document store.

The impact of the change could be minimised by submitting all the operations, then getting all the results, rather than submitting/getting each batch before moving onto the next one, but there would still be no guarantees that other operations didn't happen between batches.

I'm not actually sure if this presents any problems or not, I was trying to think of a few scenarios:

1. A new local revision is written between revs diff batch calculations that matches a server side revision (in the case of a document with many revisions such that more than 1 batch exists for the same doc ID) then it is possible we would record a revision as missing when it was present and subsequently attempt to add it, which I believe could cause a replication to fail.

1. A new local revision is deleted between revs diff batch calculations. I think this has no impact currently, but may introduce additional complexity if we later introduced a purge.

Thoughts?

rhyshort: For case 1, I think we can still get in the situation where the replication will error if a local doc with a matching id and rev gets written locally. Running the split on the queue does reduce the window for this to happen but could still easily happen if network latency is high enough. 

For case 2, since we only perform local revs diff when pulling, I don't think it is a problem right now, but with purge, I think that would come down to the exact use case, off the top of my head I can't think of a reason to purge a doc locally while it's still on the server and being actively pulled down.

We could submit all the calls at the same time and it should reduce the window for an operation to be interleaved with the calls for revs diff, but the exact gains would depend on the implementation of the executor service. It will need some work on the queue to enable submitting batches of tasks to be executed, but I think that is only a small amount of work.

ricellis: I wasn't sure what you meant for case 1, but I realise that you're referring to the situation can exist between replicator batches anyway, let alone the revs diff batches, right?

rhyshort: right.

PR #456, opened 2017-01-04
Wrapped stream operations with retries

### LINKED ISSUE #381: Attachments Pull Test Failure
Slightly different failure this time:

```
com.cloudant.sync.replication.AttachmentsPullTest > pullRevisionsWithBigAttachments[0] FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:86)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertTrue(Assert.java:52)
        at com.cloudant.sync.replication.AttachmentsPullTest.pull(AttachmentsPullTest.java:318)
        at com.cloudant.sync.replication.AttachmentsPullTest.pullRevisionsWithBigAttachments(AttachmentsPullTest.java:103)
```

ricellis:
*What*

Proposal for wrapping response stream processing in the retry handling. This is an experimental change for which I am soliciting feedback at the very least more tests need to be added before merging. It would be nice to hit this branch with the network reliability testing as well before any attempt to really merge it.

*Why*

Intermittent failures of the `AttachmentsPullTest` (see #381)  reveal that our HTTP retries currently only operate up to the point of reading response codes. After that time any exception that happens during the handling of the HTTP response `InputStream` will not be retried. To make replications in the library more resilient to network issues the stream handling could also be included in the retries.

I think there is an argument that this code should really be included in the cloudant-http module, but as it stands we currently have two levels of retries in sync-android and it is less invasive to change it at the `CouchClient` retry than the `HttpConnection` retry.

*How*

* Wrapped stream operations with retries
    * Added `InputStreamProcessor` to enable retry around stream operations.
    * Wrapped stream operations in the existing retry block.
    * Added `AttachmentPullProcessor` to combine remote stream read and attachment preparation.
    * Updated `PullStrategy` to use new `AttachmentPullProcessor`.
* Added comment around unusual _bulk_docs result
* Streamlined remote document updates
* Improved resilience of remote checkpoint write

*Testing*

Refactored `CouchClientBasicTest` and `AttachmentsPushTest` for method signature changes.
Existing tests pass.
Ran unreliable network testing, which also passes except for issue #479.
Suppressed `412` exceptions during replication test remote DB create (so retries don't fail the test, they are uniquely named so it should only happen on retry cases).
Tweaked `PullReplicationTerminationTest` to remove unnecessary code.
Improved resilience of reading writes in `AttachmentsPullTest`.

*Issues*

Fixes #381

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/AttachmentPullProcessor.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/AttachmentPullProcessor.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright © 2017 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package com.cloudant.sync.internal.replication;
+
+import com.cloudant.sync.documentstore.Attachment;
+import com.cloudant.sync.documentstore.AttachmentException;
+import com.cloudant.sync.documentstore.UnsavedStreamAttachment;
+import com.cloudant.sync.internal.documentstore.PreparedAttachment;
+import com.cloudant.sync.internal.mazha.CouchClient;
+
+import java.io.InputStream;
+
+public class AttachmentPullProcessor implements CouchClient
+        .InputStreamProcessor<PreparedAttachment> {
+
+    private final DatastoreWrapper datastoreWrapper;
+    private final String contentType;
+    private final Attachment.Encoding encoding;
+    private final long length;
+    private final long encodedLength;
+
+    AttachmentPullProcessor(DatastoreWrapper wrapper, String name, String contentType, String
+            encoding, long length, long encodedLength) {
+        this.datastoreWrapper = wrapper;
+        this.contentType = contentType;
+        this.encoding = Attachment.getEncodingFromString(encoding);
+        this.length = length;
+        this.encodedLength = encodedLength;
+    }
+
+    @Override
+    public PreparedAttachment processStream(InputStream stream) throws AttachmentException {
+        UnsavedStreamAttachment usa = new UnsavedStreamAttachment(stream, contentType, encoding);
+        PreparedAttachment attachment = datastoreWrapper.prepareAttachment(usa, length,
+                encodedLength);
+        return attachment;
+    }
+}
--- a/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/AttachmentsPullTest.java
+++ b/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/AttachmentsPullTest.java
@@ -17,8 +17,8 @@
 package com.cloudant.sync.internal.replication;
 
 import com.cloudant.common.RequireRunningCouchDB;
-import com.cloudant.sync.internal.mazha.Response;
 import com.cloudant.sync.documentstore.Attachment;
+import com.cloudant.sync.internal.mazha.Response;
 import com.cloudant.sync.util.TestUtils;
 
 import org.junit.Assert;
@@ -33,6 +33,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Created by tomblench on 26/03/2014.
@@ -180,8 +181,6 @@ public class AttachmentsPullTest extends ReplicationTestBase {
         bar._attachments = null;
 
         Response res = remoteDb.update(id, bar);
-        bar = remoteDb.get(BarWithAttachments.class, res.getId());
-
         rev = res.getRev();
     }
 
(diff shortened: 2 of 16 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/AttachmentPullProcessor.java
```diff
@@ -0,0 +1,46 @@
...
+            encoding, long length, long encodedLength) {
+        this.datastoreWrapper = wrapper;
+        this.name = name;
+        this.contentType = contentType;
+        this.encoding = Attachment.getEncodingFromString(encoding);
+        this.length = length;
+        this.encodedLength = encodedLength;
+    }
+
+    static PreparedAttachment pulLAttachmentWithRetry(CouchClient client, DatastoreWrapper
```
tomblench: This method could maybe live in `CouchClient`

tomblench: also `pulL` → `pull`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/AttachmentPullProcessor.java
```diff
@@ -0,0 +1,46 @@
...
+                                                      long length, long encodedLength) {
+        return client.processAttachmentStream(id, rev, name, true, new AttachmentPullProcessor
+                (wrapper, name, contentType, encoding, length, encodedLength));
+    }
+
+    @Override
+    public PreparedAttachment processStream(InputStream stream) throws AttachmentException {
+        UnsavedStreamAttachment usa = new UnsavedStreamAttachment(stream, name, contentType,
+                encoding);
+        PreparedAttachment attachment = datastoreWrapper.prepareAttachment(usa, length,
```
tomblench: This could return the `UnsavedStreamAttachment` and reduce the coupling on `DatastoreWrapper` which seems awkward

ricellis: Yes, the coupling is awkward, but as soon as we leave the scope of the `processStream` method then we lose the scope of the retry. So if we return the `UnsavedStreamAttachment` here we are back to where we were and the call to `prepareAttachment` (which is what actually consumes the response stream) will happen without a retry.

I think the alternative is another sub-type of `UnsavedStreamAttachment` that either stores the bytes in memory or in another temporary file on disk so they can be consumed separately. I felt that the memory consumption was likely undesirable and since the `PreparedAttachment` already places the file temporarily on disk it seemed wasteful to do that operation twice.

I guess there might be a contrivance where the same temporary file on disk is simultaneously used by the `UnsavedStreamAttachment` and the `PreparedAttachment`, but the awkwardness there was the use of the datastore's own attachments directory, which necessitates some knowledge of the datastore and so only moves the coupling slightly.

Happy to see an example if there is a better way that I'm missing.

ricellis: @tomblench I've had a go at improving this in 293446e - I moved the static method into the `CouchDB` interface (and parameterized it to instead accept an `InputStreamProcessor<T>`) with the implementation in the `CouchClientWrapper`.

This has these advantages:
1. Avoiding the need to cast the `CouchDB` in the `PullStrategy` to a `CouchClientWrapper` to get access to the `CouchClient` it can be used as a `CouchDB` as was the case previously.
1. De-coupling the `DatastoreWrapper` from the `CouchClient` in the `AttachmentPullProcessor`. This makes the `PullStrategy` the only place where the coupling between the `CouchDB` and `DatastoreWrapper` need exist (which is where the coupling was before).

I also fixed the typo :)
WDYT?

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -505,17 +523,10 @@ public Type getType() {
...
-            T returndoc = jsonHelper.fromJson(new InputStreamReader(is, Charset.forName("UTF-8"))
-                    , type);
-            logger.fine("getDocument returning " + returndoc);
-            return returndoc;
-        } finally {
-            closeQuietly(is);
-        }
+        HttpConnection connection = Http.GET(doc);
+        T returndoc = executeToJsonObjectWithRetry(connection, type);
+        logger.fine("getDocument returning " + returndoc);
```
tomblench: This is outside the scope of the change but I wonder whether these log lines should be guarded to avoid expensive `toString()` operations being done on large documents?

cloudant-sync-datastore-core/src/main/java/com/cloudant/mazha/CouchClient.java
```diff
@@ -527,16 +538,14 @@ public Type getType() {
...
-            return returndoc;
-        } finally {
-            closeQuietly(is);
-        }
+        Map<String, Object> queries = new HashMap<String, Object>();
+        queries.put("rev", rev);
+        final URI doc = this.uriHelper.documentUri(id, queries);
+        HttpConnection connection = Http.GET(doc);
+        T returndoc = executeToJsonObjectWithRetry(connection, type);
+        logger.fine("getDocument returning " + returndoc);
```
tomblench: As above

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/AttachmentsPullTest.java
```diff
@@ -274,8 +267,24 @@ private void updateRevisionAndAttachment() {
...
-        bar = remoteDb.get(Bar.class, res.getId());
-        rev = bar.getRevision();
+        rev = getLatestRevision(id, rev);
+    }
+
+    /**
+     * Utility for reading writes. When we are writing attachments it might take a little time for
+     * new revision to appear so we validate that the rev has incremented from what we currently
+     * know about before deciding that we have the latest rev.
+     * @param docId
```
brynh: Should probably document the params here, or remove the `@param` and `@return` annotations.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/AttachmentPullProcessor.java
```diff
@@ -0,0 +1,36 @@
+package com.cloudant.sync.internal.replication;
```
ricellis: Fixed in 547f630

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/AttachmentsPullTest.java
ricellis: Fixed in f950092

Issue #410, opened 2016-11-22
Update doc/replication.md for 2.0


tomblench:
- Fix code snippets (class names and import statements)
- Remove deprecation notices

tomblench:
Fixed in #457

PR #457, opened 2017-01-04
Update code samples in replication.md for new API

tomblench:
See #410 

Also add para pointing the user to the replication policy documentation.

### DIFF
--- a/doc/replication.md
+++ b/doc/replication.md
@@ -1,10 +1,8 @@
 # Replication with Cloudant Sync Android
 
-_This functionality is available in versions 0.3.3 and up._
-
-Replication is used to synchronise data between the local datastore and a
+Replication is used to synchronise data between the local document store and a
 remote database, either a CouchDB instance or a Cloudant database. Many
-datastores can replicate with the same remote database, meaning that
+document stores can replicate with the same remote database, meaning that
 cross-device synchronisation is achieved by setting up replications from each
 device to the remote database.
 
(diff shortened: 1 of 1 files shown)


doc/replication.md
```diff
@@ -68,65 +66,62 @@ devices. This web service needs to:
...
+synchronised. [Replication policies](replication-policies.md) allow
+for more complex "policies" to be executed, eg "only replicate when on
+Wifi" or "replicate every 30 minutes".
+
 First we create a simple listener that just sets a CountDownLatch when the
 replication finishes so we can wait for a replication to finish without
 needing to poll:
 
 ```java
-import com.cloudant.sync.event.Subscribe;
```
emlaver: Why were the `import` statements removed?

tomblench: I can be convinced to add them back but I think they just add noise, and user's IDEs will add them automatically.

I try to think of it this way - the code snipped is not giving a fully formed class which will compile on its own, so why do we need to provide imports?

emlaver: Ok, makes sense, keep them removed.

doc/replication.md
```diff
@@ -68,65 +66,62 @@ devices. This web service needs to:
...
 
 Replications are set up in code on a device. Use `ReplicatorBuilder` to create
 a pre-configured `Replicator` object. Each `Replicator`
 object can be assigned a `Listener` to receive message when replication
 completes or encounters an error.
 
+The replication examples below are 'one-shot' and will run once until
+completion, at which point the local and remote databases will be
+synchronised. [Replication policies](replication-policies.md) allow
+for more complex "policies" to be executed, eg "only replicate when on
```
ricellis: as per discussion on the other doc change yesterday, we should change `eg` to `for example`

doc/replication.md
```diff
@@ -68,65 +66,62 @@ devices. This web service needs to:
...
 
-Next we replicate a local datastore to a remote database:
+Next we replicate a local document store to a remote database:
 
 ```java
-import com.cloudant.sync.replication.ReplicatorBuilder;
-import com.cloudant.sync.replication.Replicator;
-
 // Username/password are supplied in the URL and can be Cloudant API keys
 URI uri = new URI("https://username:password@username.cloudant.com/my_database");
```
ricellis: I'm not sure if "username" is a known account, but this should probably be updated to "examples.cloudant.com" or something in the IANA documentation namespace (although `cloudant.com` probably provides a more clear example than `cloudant.example`).

ricellis: Note there are multiple instances.

tomblench:
@ricellis fc9cc7481a2f4b67428222d945144efcd92fef5c should address all outstanding changes.

Issue #407, opened 2016-11-22
Update doc/interceptors for 2.0


tomblench:

tomblench:
Done in #459

PR #459, opened 2017-01-05
Update interceptors documentation for new API

tomblench:
See #407

* Remove imports which add unnecessary noise to code example
* Change example to use basic auth (since we add cookie auth by default
  now)
* Datastore -> DocumentStore

### DIFF
--- a/doc/interceptors.md
+++ b/doc/interceptors.md
@@ -15,22 +15,19 @@ your class:
 - To modify the outgoing request, `HTTPConnectionRequestInterceptor`.
 - To examine the incoming response, `HTTPConnectionResponseInterceptor`.
 
-For an example of how to implement a interceptor, see  the `CookieInterceptor` class.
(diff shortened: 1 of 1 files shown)


doc/interceptors.md
```diff
@@ -15,22 +15,19 @@ your class:
...
 
-For example, this is how to add an instance of `CookieInterceptor` to a pull replication:
-
-```java
-import com.cloudant.sync.replication.Replicator
-import com.cloudant.sync.replication.ReplicatorBuilder
-import com.cloudant.sync.datastore.Datastore
-import com.cloudant.sync.http.CookieFilter
+One of the built-in interceptors is `BasicAuthInterceptor` which can be used to
+auhenticate with servers where cookie authentication is not supported.
```
brynh: Typo `auhenticate`

tomblench:
@brynh 0796a40 should address your feedback

PR #467, opened 2017-01-09
Update code and prose for new API

tomblench:
Update instances of Datastore to DocumentStore and fix code snippets

### DIFF
--- a/doc/encryption.md
+++ b/doc/encryption.md
@@ -1,6 +1,6 @@
-# Datastore encryption
+# DocumentStore encryption
 
-Android’s datastore now supports encryption of all data inside your database 
+Android’s document store now supports encryption of all data inside your database 
 using 256-bit AES: JSON documents, Query indexes and attachments.
 
 JSON documents and Query indexes are stored in SQLite databases. We use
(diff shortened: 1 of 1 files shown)


doc/encryption.md
```diff
@@ -106,18 +105,16 @@ protected void onCreate(Bundle savedInstanceState) {
...
- 
-    Datastore ds = null;
+    DocumentStore ds = null;
     try {
-        ds = manager.openDatastore("my_datastore", keyProvider);
-    } catch (DatastoreNotCreatedException e) {
+        ds = DocumentStore.getInstance(new File(path, "my_documentstore"), keyProvider);
+    } catch (DocumentStoreNotOpenedException e) {
         e.printStackTrace();
     }
```
emlaver: Update exception at line 144 (https://github.com/cloudant/sync-android/pull/467/files?diff=split#diff-0ea0055410d32eaef0440d8985d51bbdR144) to include `com.cloudant.sync.documentstore`?

doc/encryption.md
```diff
@@ -57,10 +57,9 @@ Once you've included the binaries in your app's build, you need to perform some
     byte[] key = "testAKeyPasswordtestAKeyPassword".getBytes();
```
brynh: Typo `retreiving` at [line 55](https://github.com/cloudant/sync-android/pull/467/files?diff=split#diff-0ea0055410d32eaef0440d8985d51bbdR55)

doc/encryption.md
tomblench: @emlaver I thought about the exception - the real fix is to regenerate the stack trace so that it has all of the correct class names, line numbers etc, but I think that's out of scope and I'm not even sure how to trigger this exception anyway

PR #473, opened 2017-01-10
Make operations throw correct SQLException

### LINKED ISSUE #470: Incorrect SQLException handling on Android
There are three classes named `SQLException`:
- `java.sql.SQLException` (checked exception)
- `android.database.SQLException` (unchecked exception)
- `net.sqlcipher.SQLException` (unchecked exception)

The abstract class `SQLDatabase` declares that some methods throw `SQLException` (referring to `java.sql.SQLException`), however, the concrete classes `AndroidSQLite` and `AndroidSQLCipherSQLite` throw the unchecked exceptions `android.database.SQLException` and `net.sqlcipher.SQLException` respectively, which could lead to crashes at runtime for exceptions that on initial inspection appear to be handled.

To fix this, we should catch `android.database.SQLException` and `net.sqlcipher.SQLException` and wrap them in a `java.sql.SQLException` so they behave as expected.

brynh:
Wrap `android.database.SQLException` and `net.sqlcipher.SQLException` in `java.sql.SQLException`.

Fixes #470

### DIFF
--- a/cloudant-sync-datastore-android-encryption/src/main/java/com/cloudant/sync/internal/sqlite/android/AndroidSQLCipherSQLite.java
+++ b/cloudant-sync-datastore-android-encryption/src/main/java/com/cloudant/sync/internal/sqlite/android/AndroidSQLCipherSQLite.java
@@ -33,6 +33,7 @@ import com.cloudant.sync.internal.util.Misc;
 
 import net.sqlcipher.database.SQLiteConstraintException;
 import net.sqlcipher.database.SQLiteDatabase;
+import net.sqlcipher.database.SQLiteException;
 
 import java.io.File;
 
@@ -64,7 +65,12 @@ public class AndroidSQLCipherSQLite extends SQLDatabase {
 
     @Override
     public void compactDatabase() {
-        database.execSQL("VACUUM");
+        try {
+            database.execSQL("VACUUM");
+        } catch (net.sqlcipher.SQLException e) {
+            String error = "Fatal error running 'VACUUM', the database is probably malfunctioning.";
+            throw new IllegalStateException(error);
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-android-encryption/src/main/java/com/cloudant/sync/internal/sqlite/android/AndroidSQLCipherSQLite.java
```diff
@@ -101,19 +102,27 @@ public void setTransactionSuccessful() {
     }
 
     @Override
-    public void execSQL(String sql)  {
+    public void execSQL(String sql) throws java.sql.SQLException {
```
rhyshort: Missed compact on line 68

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/internal/sqlite/android/AndroidSQLite.java
```diff
@@ -86,15 +86,23 @@ public void setTransactionSuccessful() {
     }
 
     @Override
-    public void execSQL(String sql) throws SQLException {
+    public void execSQL(String sql) throws java.sql.SQLException {
```
rhyshort: compact missed here too.

cloudant-sync-datastore-android-encryption/src/main/java/com/cloudant/sync/internal/sqlite/android/AndroidSQLCipherSQLite.java
brynh: Good spot. Fixed in 4272ecb.

I've handled it the same way as the java se version in [SQLiteWrapper.java](https://github.com/cloudant/sync-android/blob/master/cloudant-sync-datastore-javase/src/main/java/com/cloudant/sync/sqlite/sqlite4java/SQLiteWrapper.java#L104) so they are consistent.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/internal/sqlite/android/AndroidSQLite.java
brynh: Fixed in 4272ecb.

PR #474, opened 2017-01-11
Move constants for Callables from DatabaseImpl

### LINKED ISSUE #355: Refactor - move DatastoreImpl constants
Once all of the SQL work is moved into `Callable` classes, remove the SQL static String constants from `DatastoreImpl`. They should either be declared in the `Callable`s where they are used, or grouped together in their own class.

See #291 and #342

rhyshort:
Move constants for Database callables to a `CallableSQLConstants class`

Resolves #355

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
@@ -101,46 +101,6 @@ public class DatabaseImpl implements Database {
 
     private static final Logger logger = Logger.getLogger(DatabaseImpl.class.getCanonicalName());
 
-    // for building up the strings below
-    public static final String METADATA_COLS = "docs.docid, docs.doc_id, revid, sequence, " +
-            "current, deleted, parent";
-
-    public static final String FULL_DOCUMENT_COLS = METADATA_COLS + ", json";
-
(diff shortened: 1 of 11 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
```diff
@@ -101,41 +101,6 @@
...
-    // get metadata (everything except json) for a given revision and doc ID† (see below)
-    public static final String GET_METADATA_GIVEN_REVISION =
-            "SELECT " + METADATA_COLS + " " + GIVEN_REVISION_CLAUSES;
-
-    public static final String GET_DOC_NUMERIC_ID = "SELECT doc_id from docs WHERE docid=?";
-
-    public static final String SQL_CHANGE_IDS_SINCE_LIMIT = "SELECT doc_id, max(sequence) FROM " +
-            "revs WHERE sequence > ? AND sequence <= ? GROUP BY doc_id ";
-
     // † N.B. whilst there should only ever be a single result bugs have resulted in duplicate
```
brynh: This note should be moved to the new `CallableSQLConstants` class.

rhyshort: Ah great spot, moved in 9d0c2cc

PR #475, opened 2017-01-12
Remove batch limit for replications.

### LINKED ISSUE #462: ReplicationCompleted event is misleading
The `ReplicationCompleted` event can be misleading in that it can either mean all the data has been replicated, or it can mean that we've hit the limit on the size of replication due to the combination of `changeLimitPerBatch` and `batchLimitPerRun` in the `PushStrategy` or `PullStrategy` classes.

I can't find any documentation to describe this behaviour and as a user I think I'd assume that if I received a `ReplicationCompleted` event it meant that all the data had been replicated.

At the very least, we should document this behaviour, but given that `PushStrategy` and `PullStrategy` are private, we probably should instead do something more obvious, like:
* have a flag in the `ReplicationCompleted` object to indicate whether we've hit the limits or replicated all the changes;
* have a separate event for the case where replication has stopped because we've hit the limits;
* remove the limits so that we only get `ReplicationCompleted` in the case we've successfully replicated all the changes.

rhyshort:
## What
Remove the limit of the number of batches that can run for replications.

## Why
Currently replications will be marked as completed even if they did not replicate all changes, this is due to the limit on the number of batches. Removing this limit ensures that a replication will replicate until there are no more changes to be replicated from the database.

## How

- Remove `batchLimtPerRun` configuration option
- Change `for` loop in {Pull,Push}Strategy to `while` and use cancel as the condition.

## Issues

Resolves #462

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
@@ -44,6 +44,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Level;
@@ -92,8 +93,6 @@ public class PullStrategy implements ReplicationStrategy {
     private boolean useBulkGet = false;
 
     public int changeLimitPerBatch = 1000;
-
-    public int batchLimitPerRun = 100;
 
     public int insertBatchSize = 100;
 
@@ -223,12 +222,8 @@ public class PullStrategy implements ReplicationStrategy {
         }
 
         this.state.documentCounter = 0;
-        for (this.state.batchCounter = 1; this.state.batchCounter < this.batchLimitPerRun; this
-                .state.batchCounter++) {
-
-            if (this.state.cancel) {
-                return;
-            }
+        while(!this.state.cancel) {
+            this.state.batchCounter++;
 
             String msg = String.format(
                     "Batch %s started (completed %s changes so far)",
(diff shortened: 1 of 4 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
```diff
@@ -223,12 +221,8 @@ private void replicate()
         }
 
         this.state.documentCounter = 0;
-        for (this.state.batchCounter = 1; this.state.batchCounter < this.batchLimitPerRun; this
-                .state.batchCounter++) {
-
-            if (this.state.cancel) {
-                return;
-            }
+        while(!this.state.cancel) {
```
brynh: This change to the `while` loop means that if the replication was cancelled, we'd get the logging after the loop indicating that the pull completed, which could prove misleading when debugging from logs.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PushStrategy.java
```diff
@@ -206,12 +204,9 @@ private void replicate()
         }
 
         this.state.documentCounter = 0;
-        for (this.state.batchCounter = 1; this.state.batchCounter < this.batchLimitPerRun; this
-                .state.batchCounter++) {
+        while (!this.state.cancel) {
```
brynh: This change to the `while` loop means that if the replication was cancelled, we'd get the logging after the loop indicating that the pull completed, which could prove misleading when debugging from logs.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
rhyshort: done in ffcc25b

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PushStrategy.java
rhyshort: done in ffcc25b

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
```diff
@@ -273,12 +268,21 @@ private void replicate()
...
         long deltaTime = endTime - startTime;
-        String msg = String.format(
-                "Pull completed in %sms (%s total changes processed)",
-                deltaTime,
-                this.state.documentCounter
-        );
+        String msg;
+        if (this.state.cancel) {
+            msg = String.format(Locale.ENGLISH,
+                    "Pull canceled after %sms (%s changes processed",
```
tomblench: Missing right paren and this message differs from the other cancellation message

rhyshort: fixed in c363ac9

PR #476, opened 2017-01-16
Restore isTextSearchEnabled method

tomblench:
_What_

Restore `isTextSearchEnabled` method.

This had previously been removed from the implementation and not added
to the interface.

_Why_

This is public API and mentioned in the query markdown doc

_Testing_

See `validateTextSearchIsAvailable`

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/Query.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/Query.java
@@ -164,8 +164,10 @@ public interface Query {
             throws QueryException;
 
     /**
-     * Returns whether the underlying SQLite instance supports Full Text Search (FTS)
-     * @return whether the underlying SQLite instance supports Full Text Search (FTS)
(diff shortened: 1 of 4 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/query/Query.java
```diff
@@ -163,5 +163,11 @@ QueryResult find(Map<String, Object> query,
                      List<FieldSort> sortDocument)
             throws QueryException;
 
+    /**
+     * Returns whether the underlying SQLite instance supports Full Text Search (FTS)
+     * @return whether the underlying SQLite instance supports Full Text Search (FTS)
```
emlaver: What about `@return true if underlying SQLite instance supports FTS; false otherwise.`?

tomblench: Good suggestion - done in 361f27f

PR #480, opened 2017-01-17
Add ability to have multiple PeriodicReplicationServices

brynh:
This PR fixes the following issues:
- The abstract class `PeriodicReplicationService` currently stores information in `SharedPreferences` under fixed key names. This means that if an app implements multiple concrete `PeriodicReplicationService`s they will interfere with each other. To fix this, we now store the values in `SharedPreferences` using keys prefixed with the name of the concrete class implementing the `PeriodicReplicationService`.
- After a reboot, the elapsed time since boot at which the next should occur isn't always updated to reflect the fact the device rebooted. This could lead to replications not being invoked at the expected times.
- Rather than storing the time at which the next replication should be triggered, we now store the time at which the last replication occurred. This feels more logical and means we don't have to recalculate the value stored in `SharedPreferences` after bind/unbind and the only time the value will need adjusting is after a reboot.

### Testing

The existing tests have been updated in accordance with these changes. They now verify that the keys used in `SharedPreferences` are prefixed with the name of the concrete implementation of `PeriodicReplicationService` and verify that the elapsed time since boot is always updated after the service has been notified of a reboot.

### DIFF
--- a/cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
+++ b/cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
@@ -35,6 +35,8 @@ import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.closeTo;
 import static org.mockito.Matchers.anyInt;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Mockito.mock;
@@ -47,6 +49,16 @@ public class ReplicationServiceTest extends ServiceTestCase<TestReplicationServi
 
     private static final long ALARM_TOLERANCE_MS = 500;
     private static final int DEFAULT_WAIT_SECONDS = 5;
+    private static final int MILLIS_IN_SECOND = 1000;
+
+    /** An arbitrary number of replication intervals that's greater than one. */
+    private static final int ARBITRARY_NUMBER_OF_INTERVALS = 10;
+
+    /** An arbitrary number of replication intervals that's between zero and one. */
+    private static final float ARBITRARY_FRACTIONAL_INTERVAL = 0.5f;
+
+    private static final String PREFERENCE_CLASS_NAME = "com.cloudant.android.TestReplicationService";
+
     private Context mMockContext;
     private SharedPreferences mMockPreferences;
     private SharedPreferences.Editor mMockPreferencesEditor;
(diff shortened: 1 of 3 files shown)


cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
```diff
@@ -682,4 +695,22 @@ public void testSetReplicatorsMultipleInvocations() {
         }
     }
 
+    private void checkClockTime(long expectedTime, long actualTime) {
+        assertTrue("Last alarm elapsed time not within " + ALARM_TOLERANCE_MS + "ms of expected time",
+                Math.abs(expectedTime - actualTime) < ALARM_TOLERANCE_MS);
+    }
+
+    private void checkElapsedTime(long expectedTime, long actualTime) {
+        assertTrue("Last alarm clock time not within " + ALARM_TOLERANCE_MS + "ms of expected time",
```
tomblench: More idiomatic would be to use a Hamcrest matcher like [closeTo](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#closeTo(double,%20double))

cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
```diff
@@ -105,6 +106,9 @@ public void testOnStartCommandRebootImmediateAlarm() {
         service.onCreate();
         final CountDownLatch latch = new CountDownLatch(1);
 
+        long timeReturned = System.currentTimeMillis() - ((1000 * service.getUnboundIntervalInSeconds()) * 10);
```
tomblench: What does the `*10` do in this context?

cloudant-sync-datastore-android/src/test/java/com/cloudant/sync/replication/ReplicationServiceTest.java
```diff
@@ -169,17 +175,24 @@ public void operationStarted(int operationId) {
         service.setReplicators(mMockReplicators);
         try {
             assertTrue("The countdown should reach zero", latch.await(DEFAULT_WAIT_SECONDS, TimeUnit.SECONDS));
-            verify(mMockPreferencesEditor, never()).putLong(captorPrefKeys.capture(), captorPrefValues.capture());
+            verify(mMockPreferencesEditor, times(2)).putLong(captorPrefKeys.capture(), captorPrefValues.capture());
+            List<String> prefsKeys = captorPrefKeys.getAllValues();
+            List<Long> prefsValues = captorPrefValues.getAllValues();
+            long expectedElapsedTime = SystemClock.elapsedRealtime() - (1000 * service.getUnboundIntervalInSeconds() / 2);
```
tomblench: Although you've written some clear comments for each test I'm struggling to understand what these magic numbers like `/ 2` are for. Perhaps pulling them out into constants with comments might help.

brynh:
@tomblench I've tried to address your comments in 8bfe34e. Hopefully it makes the tests more easily readable.

Issue #402, opened 2016-11-22
v2.0 API migration guide


tomblench:
Provide a markdown document with simple 'recipes' to transition from 1.x to 2.0

In particular this should note the difference between the old and new directory locations when going from the old "DatastoreManager directory + Datastore name" scheme

tomblench:
Done in #481

PR #481, opened 2017-01-18
Version 2.0 API Migration Guide

tomblench:
See #402

### DIFF
--- a/doc/migration.md
+++ b/doc/migration.md
@@ -1,70 +1,80 @@
-Datastore -> DocumentStore
+# Datastore -> DocumentStore
 
-* The DatastoreManager class has been removed - TODO example of how to get an instance vs previously
-* "CRUD" (create, read, update, delete) functionality has been migrated to the new `Database` class. To ...
-  * `getDocument` has been renamed `read`
-  * `getAllDocuments` has been renamed `read`
-  * `getDocumentsWithIds` has been renamed `read`
-  * `containsDocument` has been renamed `contains`
-  * `getAllDocumentIds` has been renamed `getIds`
-  * `getConflictedDocumentIds` has been renamed `getConflictedIds`
-  * `resolveConflictsForDocument` has been renamed `resolveConflicts`
-  * `createDocumentFromRevision` has been renamed `create`
-  * `updateDocumentFromRevision` has been renamed `update`
-  * `deleteDocumentFromRevision` has been renamed `delete`
-  * `deleteDocument` has been renamed `delete`
+The DatastoreManager class has been removed - TODO example of how to get an instance vs previously
 
+"CRUD" (create, read, update, delete) functionality has been migrated to the new `Database` class. To ...
 
-* The `throws` clauses on `Database` methods are different to those on their counterparts in `Datastore`. Code which calls these methods may have to be adjusted to catch different exceptions.
+* `getDocument` has been renamed `read`
+* `getAllDocuments` has been renamed `read`
+* `getDocumentsWithIds` has been renamed `read`
+* `containsDocument` has been renamed `contains`
+* `getAllDocumentIds` has been renamed `getIds`
+* `getConflictedDocumentIds` has been renamed `getConflictedIds`
+* `resolveConflictsForDocument` has been renamed `resolveConflicts`
+* `createDocumentFromRevision` has been renamed `create`
+* `updateDocumentFromRevision` has been renamed `update`
+* `deleteDocumentFromRevision` has been renamed `delete`
+* `deleteDocument` has been renamed `delete`
 
-Notifications/Events - namespaces
+The `throws` clauses on `Database` methods are different to those on their counterparts in `Datastore`. Code which calls these methods may have to be adjusted to catch different exceptions.
+
+# Notifications/Events - namespaces
 com.cloudant.sync.notifications -> com.cloudant.sync.event.notifications
 
+# IndexManager -> Query
 
+The `ensureIndexed` methods have been replaced by `createJsonIndex` and `createTextIndex`
 
-IndexManager -> Query
+The first argument to createJsonIndex and createTextIndex is of type `List<FieldSort>`.
 
-* The `ensureIndexed` methods have been replaced by `createJsonIndex` and `createTextIndex`
-* The first argument to createJsonIndex and createTextIndex is of type `List<FieldSort>`.
-  To specify the fields "name" and "age", replace instances of
-  ```java
-  Arrays.<Object>asList("name", "age")
-  ```
-  with
-  ```java
-  Arrays.<FieldSort>asList(new FieldSort("name"), new FieldSort("age"))
-  ```
-* The second argument to `createJsonIndex` and `createTextIndex` is the index name. This is required; to automatically generate the index name, replace instances of
-  `ensureIndexed(fields)`
-  with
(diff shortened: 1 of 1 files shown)


doc/migration.md
```diff
@@ -0,0 +1,158 @@
+# Package changes
+
+Classes in the library have been re-organised to simplify the packaging.
+
+All classes are now in the `com.cloudant.sync` package:
+
+* All private API classes are in the `com.cloudant.sync.internal`
+  package. API users are discouraged from using these classes as fields,
```
ricellis: Is `discouraged` a strong enough word for this now that the segregation is so clear?
`should not` or maybe even `must not`
>API users **must not** directly use the classes in these packages as ... .

Maybe consider something along the lines of:
>Directly calling any classes or methods in these packages is unsupported.

doc/migration.md
```diff
@@ -0,0 +1,158 @@
...
+* `getConflictedDocumentIds` has been renamed `getConflictedIds`
+* `resolveConflictsForDocument` has been renamed `resolveConflicts`
+* `createDocumentFromRevision` has been renamed `create`
+* `updateDocumentFromRevision` has been renamed `update`
+* `deleteDocumentFromRevision` has been renamed `delete`
+* `deleteDocument` has been renamed `delete`
+
+The `throws` clauses on `Database` methods are different to those on
+their counterparts in `Datastore`. Code which calls these methods may
+have to be adjusted to catch different exceptions.
```
ricellis: It's probably worth another section on `Exceptions` that could be linked from here and query that says something along the lines of how we've made an effort to use checked exceptions instead of runtime exceptions to reduce the likelihood of unexpected application crashes. Maybe worth saying that runtime exceptions may still be thrown in some circumstances and these have been documented where possible.

doc/migration.md
```diff
@@ -0,0 +1,158 @@
...
+
+# Notifications and Events packages
+
+Events which were previously in the `com.cloudant.sync.notifications`
+package have moved to the `com.cloudant.sync.event.notifications`
+package.
```
ricellis: Weren't some of the classes renamed too?

doc/migration.md
```diff
@@ -0,0 +1,158 @@
...
+        new FieldSort("age")),
+    "textIndex",
+    new Tokenizer("porter")).indexName;
+```
+
+`createJsonIndex` and `createTextIndex` now return the full `Index`
+object. To obtain the index name, use the `indexName` property of the
+returned index.
+
+`updateAllIndexes` has been renamed `updateAllIndexes`.
```
ricellis: not much of a rename ;)

doc/migration.md
```diff
@@ -0,0 +1,158 @@
...
+
+Methods on the `Query` interface throw checked exceptions rather than
+throwing `null` to indicate an error condition. To check for error
```
brynh: `throwing` -> `returning`

ricellis:
I guess some of this will be covered in the changes log, but we might want to mention these here as they are behaviour changes:
Entire paths being created as per https://github.com/cloudant/sync-android/issues/294
No more replication limit as per https://github.com/cloudant/sync-android/issues/462

tomblench:
@brynh @ricellis 5a0f09d should address all of these

doc/migration.md
```diff
@@ -0,0 +1,214 @@
+# Package changes
+
+Classes in the library have been re-organised to simplify the packaging.
+
+All classes are now in the `com.cloudant.sync` package:
+
+* All private API classes are in the `com.cloudant.sync.internal`
+  package. API users should not use these classes as fields, method
+  signatures, and implementation details may be subject to
+  change. Directly using these classes are calling their methods in
```
brynh: Maybe change 
> Directly using these classes are calling their methods in these packages is not supported.

to 
>Directly using the classes in these packages or calling their methods is not supported.

doc/migration.md
```diff
@@ -0,0 +1,214 @@
...
+File path = getApplicationContext().getDir("datastores");
+DocumentStore ds = DocumentStore.getInstance(new File(path, "my_datastore"));
+// read a doc
+DocumentRevision dr = ds.database().read("my-document-id");
+```
+
+The `getInstance` method will try to create all necessary
+sub-directories in order to construct the path represented by the
+`File` argument. This differs from the behaviour of the 1.x versions
+of the library would would only attempt to create one level of
```
brynh: `would would` -> `that would`

doc/migration.md
```diff
@@ -0,0 +1,214 @@
...
+equivalent index if it exists, regardless of the requested index
+name. See the javadoc for these methods for more details.
+
+The `close` method has been removed. The native resources used by the
+indexes database are released when the owning `DocumentStore` has
+`close` called on it.
+
+# Changes to the Replicator and `ReplicatorBuilder`
+
+The `batchLimitPerRun` property has been removed the Pull and Push
```
brynh: `removed` -> `removed from`

doc/migration.md
```diff
@@ -0,0 +1,214 @@
...
+
+This differs from the behaviour of the 1.x versions of the library
+which were much more likely to throw runtime exceptions which the
+developer could not anticipate.
+
+Situations for which runtime (unchecked) exceptions may be thrown include:
+
+* Incorrect or `null` arguments passed to an API method: in these
+  cases `NullPointerException` or `IllegalArgumentException` can be
+  thrown. The developer can defend against these by ensure that
```
brynh: `ensure` -> `ensuring`

doc/migration.md
```diff
@@ -0,0 +1,214 @@
...
+  correct and nun-`null` arguments are passed to these methods. The
```
brynh: `nun` -> `non`

doc/migration.md
brynh: We should probably have a link to the section on Exceptions here.

tomblench:
@brynh thanks for spotting all those silly typos. 045b700 should address these.

doc/migration.md
```diff
@@ -0,0 +1,230 @@
...
+`File` argument. This differs from the behaviour of the 1.x versions
+of the library which would only attempt to create one level of
+directories.
+
+As the above example shows, "CRUD" (create, read, update, delete)
+functionality has been migrated to the new `Database` class. Obtain an
+instance of the `Database` class managed by the `DocumentStore` by calling
+the `database()` getter method.
+
+The method names on the `Database` methods are different to their
```
brynh: `methods` -> `class`

doc/migration.md
```diff
@@ -0,0 +1,239 @@
...
+
+`createJsonIndex` and `createTextIndex` now return the full `Index`
+object. To obtain the index name, use the `indexName` property of the
+returned index.
+
+`updateAllIndexes` has been renamed `refreshAllIndexes`.
+
+Methods on the `Query` interface throw checked exceptions rather than
+returning `null` to indicate an error condition. To check for error
+conditions, enclose your existing code in `try/catch` block instead of
```
brynh: `in` -> `in a`

doc/migration.md
tomblench: Done in ef21345

doc/migration.md
tomblench: Done in ef21345

PR #483, opened 2017-01-19
Replication timer fixes

brynh:
### What

- Stopping and starting periodic replication currently occurs in an inconsistent way. Sometimes when an explicit call to start periodic replications is made after an explicit stop, the replications will occur immediately and sometimes the replications will continue the previous schedule.
- It is not obvious how to change the schedule of periodic replications in progress and one might assume that by changing the values returned by PeriodicReplicationService, that would be sufficient. This PR improves and documents the process.

### How

Store a new value in `SharedPreferences` indicating whether the service was explicitly stopped, by sending an intent containing `PeriodicReplicationService.COMMAND_STOP_PERIODIC_REPLICATION`, or implicitly stopped (e.g. by rebooting the device).

This allows us to implement the following rules.

- After an explicit stop of the `PeriodicReplicationService`, any subsequent start of it will cause a replication to begin immediately.
- After an implicit stop (e.g. device reboot), any subsequent start of the `PeriodicReplicationService` will cause the replications to continue on the previous schedule (as if the device had not been rebooted) unless we have missed a replication that was due when the device was off (in which case the first replication will occur immediately and subsequent replications will occur at the intervals specified by the `PerioidicReplicationService` from the current time).

To support a use-case where an app allows the frequency of replication to be adjusted (e.g. by some user preference), we now add a new `PeriodicReplicationService.COMMAND_RESET_REPLICATION_TIMERS` command that can be sent in an `Intent` to the replication service to indicate that the replication timers should be re-evaluated.

The Replication Policy documentation has been updated to describe this behaviour.

### Testing

New tests have been added to test this functionality.

### DIFF
--- a/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
+++ b/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
@@ -66,11 +66,50 @@ public abstract class PeriodicReplicationService<T extends PeriodicReplicationRe
     private static final String PERIODIC_REPLICATION_ENABLED_SUFFIX
         = ".periodicReplicationsActive";
 
+    /* We store a flag indicating whether periodic replications were explicitly stopped in
+     * SharedPreferences using a key with this suffix. We have to store the flag persistently as
+     * the service may be stopped and started by the operating system. */
+    private static final String EXPLICITLY_STOPPED_SUFFIX = ".explicitlyStopped";
+
     private static final long MILLISECONDS_IN_SECOND = 1000L;
 
+    /**
+     * To start periodic replications, this value should be passed to this service in an Intent
+     * using an Extra with the {@link #EXTRA_COMMAND} key.
+     * @see
+     * <a href="http://github.com/cloudant/sync-android/blob/master/doc/replication-policies.md#controlling-the-replication-service">
+     *     Replication Policy User Guide</a> for more details.
+     */
     public static final int COMMAND_START_PERIODIC_REPLICATION = 2;
+
+    /**
+     * To stop periodic replications, this value should be passed to this service in an Intent
+     * using an Extra with the {@link #EXTRA_COMMAND} key.
+     * @see
+     * <a href="http://github.com/cloudant/sync-android/blob/master/doc/replication-policies.md#controlling-the-replication-service">
+     *     Replication Policy User Guide</a> for more details.
+     */
     public static final int COMMAND_STOP_PERIODIC_REPLICATION = 3;
+
+    /**
+     * When the device is rebooted, this value should be passed to this service in an Intent
+     * using an Extra with the {@link #EXTRA_COMMAND} key so that replications can be setup again
+     * following a reboot.
+     * @see
+     * <a href="http://github.com/cloudant/sync-android/blob/master/doc/replication-policies.md#controlling-the-replication-service">
+     *     Replication Policy User Guide</a> for more details.
+     */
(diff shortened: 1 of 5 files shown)


cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
```diff
@@ -66,11 +66,17 @@
...
+     * SharedPreferences using a key with this suffix. We have to store the flag persistently as
+     * the service may be stopped and started by the operating system. */
+    private static final String EXPLICITLY_STOPPED_SUFFIX = ".explicitlyStopped";
+
     private static final long MILLISECONDS_IN_SECOND = 1000L;
 
     public static final int COMMAND_START_PERIODIC_REPLICATION = 2;
     public static final int COMMAND_STOP_PERIODIC_REPLICATION = 3;
     public static final int COMMAND_DEVICE_REBOOTED = 4;
+    public static final int COMMAND_RESET_REPLICATION_TIMERS = 5;
```
emlaver: Is a comment needed here? Also, how did you come to the value of 5?

cloudant-sync-datastore-android/src/test/java/com/cloudant/android/TestReplicationService.java
```diff
@@ -25,6 +25,8 @@
     private static final String DATASTORE_MANGER_DIR = "data";
     private static final String TAG = "TestReplicationService";
 
+    private int mUnboundIntervalSeconds = 120;
```
emlaver: Although this value already existed before this PR, is adding a comment worthwhile? is it documented elsewhere?

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
brynh: Added comment in 3751a3a.  The value 5 is chosen because it's one more than the current highest of the `COMMAND_*` values currently in use. See my additional comment also about this, which I've made separately so it doesn't get folded by GitHub.

cloudant-sync-datastore-android/src/test/java/com/cloudant/android/TestReplicationService.java
brynh: Fixed in 3751a3a

brynh:
@emlaver I've addressed your comments in 3751a3a, but it also got me thinking we should reserve some IDs for the `COMMAND_*` commands for possible future extension, just in case anyone tries to extend `ReplicationService` or `PeriodicReplicationService` for some reason.  I've therefore added code to throw a `RuntimeException` in the service if a user uses a number between our current maximum in use (5) and 99 so that we can safely use those numbers ourselves without interfering with any extension users have made. I've also added a note to the Replication Policy document to describe this. I've manually tested this to make sure it works as expected.

emlaver:
@brynh Ok great and I agree with your decision to reserve the IDS for `COMMAND_*`.

doc/replication-policies.md
```diff
@@ -105,6 +114,9 @@ If you want to start replications from anywhere other than a `PeriodicReplicatio
 [`WakeLock`](http://developer.android.com/reference/android/os/PowerManager.WakeLock.html)
 management you require yourself.
 
+If you need to create a custom child of `ReplicationService` and you add additional commands passed using the `ReplicationService.EXTRA_COMMAND`
```
tomblench: subclass instead of child?

PR #485, opened 2017-01-20
Add `length` property to `Attachment`

tomblench:
This field is needed if users want to easily copy attachment contents into a byte buffer, for instance
```java
        Attachment att = retrieved.getAttachments().get("cute_cat.jpg");
        InputStream is = att.getInputStream();

        // get all bytes in memory if required
        byte[] data = new byte[(int) att.length];
        is.read(data);
```
(taken from crud.md)

Defaults to -1 if unknown

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/Attachment.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/Attachment.java
@@ -37,10 +37,11 @@ import java.io.InputStream;
  */
 public abstract class Attachment implements Comparable<Attachment>{
 
-    public Attachment(String name, String type, Encoding encoding) {
+    public Attachment(String name, String type, Encoding encoding, long length) {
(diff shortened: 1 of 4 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/Attachment.java
```diff
@@ -37,10 +37,11 @@
  */
 public abstract class Attachment implements Comparable<Attachment>{
 
-    public Attachment(String name, String type, Encoding encoding) {
+    public Attachment(String name, String type, Encoding encoding, long length) {
```
ricellis: Given that this class is `abstract` should this constructor really be `protected` rather than `public`?

Since we only document `public` visibility I guess the question really is, should we be documenting this constructor so that people have guidance if they are constructing their own subclasses or do we consider our subclasses sufficient in which case we should mark this protected and omit it from the docs?

PR #487, opened 2017-01-20
Change getConflictedIds signature

tomblench:
Change signature to Iterable so that the result can be used by the for/each syntax


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/Database.java
```diff
@@ -233,7 +233,7 @@ DocumentRevision read(String documentId, String revisionId) throws
      * @see <a target="_blank" href="http://wiki.apache.org/couchdb/Replication_and_conflicts">Replication and conflicts</a>
      * @throws DocumentStoreException if there was an error reading from the database.
      */
-    Iterator<String> getConflictedIds() throws DocumentStoreException;
+    Iterable<String> getConflictedIds() throws DocumentStoreException;
```
brynh: We could now remove the import for `Iterator` from this class (and from `DatabaseImpl.java`)

Issue #468, opened 2017-01-09
Query { "$not" : { "$size" : 5} } does not work


rhyshort:
Using `$not` does not work with `$size`. It is incorrectly translated to a SQL Query, when it needs to be translated into the java matcher. 

```
Caused by: java.sql.SQLException: com.almworks.sqlite4java.SQLiteException: [1] DB[3] prepare() SELECT _id FROM "_t_cloudant_sync_query_index_basic" WHERE _id NOT IN (SELECT _id FROM "_t_cloudant_sync_query_index_basic" WHERE "age" null ?) [near "null": syntax error]
	at com.cloudant.sync.internal.sqlite.sqlite4java.SQLiteWrapper.rawQuery(SQLiteWrapper.java:260)
	at com.cloudant.sync.internal.sqlite.sqlite4java.SQLiteWrapper.rawQuery(SQLiteWrapper.java:37)
	at com.cloudant.sync.internal.query.QueryExecutor.executeQueryTree(QueryExecutor.java:243)
```

tomblench:
Fixed in #489

emlaver:
@tomblench @brynh @ricellis Would it be beneficial to add more tests similar to the ones added for [#352](https://github.com/cloudant/CDTDatastore/issues/352) in the CDT library? Tests are in `CDTQQueryExecutorTests.m` and `CDTQQuerySqlTranslatorTests.m`: https://github.com/cloudant/CDTDatastore/pull/354/files

brynh:
@emlaver I looked at the changes made in #489 and I don't see anything that's obviously wrong, but I'd prefer some additional tests as IMO it's the safest way to be sure it works, and although I _think_ #489 looks good additional tests would help to confirm that and guard against regressions.

PR #489, opened 2017-01-24
Allow queries combining $not and $size

tomblench:
_What_ 
Fix buggy handling of queries like `{ "$not" : { "$size" : 5} }`. see #468.

_How_
Change `isOperatorFoundInClause` to find negated operators

_Testing_
Add test `canFindDocumentsUsingNOTAndSIZE`


ricellis:
@tomblench I think this needs a `CHANGES.md` entry for the fix though.

PR #490, opened 2017-01-25
Remove `name` property from attachments

tomblench:
_What_

Remove `name` property from the `Attachment` class.

Because this involves changing some internal data structures, ensure that attachment keys in json metadata and attachment entries in multiparts are written in correct order (see https://issues.apache.org/jira/browse/COUCHDB-1521).

During testing, I realised that the existing implementation of `MultipartAttachmentWriter.makeInputStream` could exhaust the number of open filehandles by opening lots of streams simultaneously (this was observed on Android). So, this was re-written.

_Why_

The canonical source for attachment names is the key in the attachments map - having a second place to store this value is confusing and inconsistent.

_Testing_

Existing tests pass, added `attachmentOrderingTest`.


ricellis:
@tomblench the build seems to have failed with a [`1802`](https://www.sqlite.org/rescode.html#ioerr_fstat) error. Haven't looked into whether it is related to the change.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/Attachment.java
```diff
@@ -128,9 +112,6 @@ public boolean equals(Object o) {
 
         Attachment that = (Attachment) o;
 
-        if (name != null ? !name.equals(that.name) : that.name != null) {
-            return false;
-        }
```
ricellis: This method is now basically saying that `Attachment`s are equal if they have the same type and encoding. That's a pretty loose equality! A few possibilities spring to mind although I'm sure there are others...
1. Use default `equals` & `hashcode` in this class, and override in subclasses where a more useful check can be made.
1. Leave these here and override in subclasses, calling `super` to get additional part of the comparisons. The downside of this is that we need to remember.
1. Add `abstract` methods that subclasses are forced to implement that is used by this `equals` and `hashcode` to provides the additional parts whilst declaring this `equals` and `hashcode` `final`. This is rigorous about ensuring correct implementations, but maybe a bit weird.
1. Decide that we aren't going to compare `Attachment` equality or use them as keys (since we tend to key of the `String` name so we just use the default `Object.equals` & `hashcode`.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/UnsavedFileAttachment.java
```diff
@@ -33,12 +33,12 @@
...
-        super(file.getName(), type, Encoding.Plain, file.length());
+        super(type, Encoding.Plain, file.length());
         this.file = file;
     }
 
     public UnsavedFileAttachment(File file, String type, Encoding encoding) {
-        super(file.getName(), type, encoding, file.length());
+        super(type, encoding, file.length());
         this.file = file;
     }
```
ricellis: As per previous comment possibly need some `equals` `hashcode` overrides.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/UnsavedStreamAttachment.java
```diff
@@ -30,13 +30,13 @@
...
+        super(type, Encoding.Plain, -1);
         this.stream = stream;
     }
 
-    public UnsavedStreamAttachment(InputStream stream, String name, String type, Encoding encoding) {
-        super(name, type, encoding, -1);
+    public UnsavedStreamAttachment(InputStream stream, String type, Encoding encoding) {
+        super(type, encoding, -1);
         this.stream = stream;
     }
```
ricellis: As per previous possibly need some `equals` `hashcode` overrides.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/AttachmentManager.java
```diff
@@ -238,12 +239,12 @@ protected static PreparedAttachment prepareAttachment(String attachmentsDir,
      * @param attachments Attachments to search.
      * @return List of attachments which already exist in the attachment store, or an empty list if none.
      */
-    public static List<SavedAttachment> findExistingAttachments(
-            Collection<? extends Attachment> attachments) {
-        ArrayList<SavedAttachment> list = new ArrayList<SavedAttachment>();
-        for (Attachment a : attachments) {
+    public static Map<String, SavedAttachment> findExistingAttachments(
+            Map<String, ? extends Attachment> attachments) {
+        Map<String, SavedAttachment> list = new HashMap<String, SavedAttachment>();
```
ricellis: The field name `list` is a bit confusing now, since this isn't a `List` any more. I'm not massively concerned about this because the method is concise and so it is clear what is going on, but it might be better named `existingAttachments` or `attachments`.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/AttachmentManager.java
```diff
@@ -255,11 +256,11 @@ protected static PreparedAttachment prepareAttachment(String attachmentsDir,
      * @param attachments Attachments to search.
      * @return List of attachments which need adding to the attachment store, or an empty list if none.
      */
-    public static List<Attachment> findNewAttachments(Collection<? extends Attachment> attachments) {
-        ArrayList<Attachment> list = new ArrayList<Attachment>();
-        for (Attachment a : attachments) {
+    public static Map<String, Attachment> findNewAttachments(Map<String, ? extends Attachment> attachments) {
+        Map<String, Attachment> list = new HashMap<String, Attachment>();
```
ricellis: `newAttachments`?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/AttachmentManager.java
```diff
@@ -275,14 +276,14 @@ protected static PreparedAttachment prepareAttachment(String attachmentsDir,
...
+    public static Map<String, PreparedAttachment> prepareAttachments(String attachmentsDir,
                                                                  AttachmentStreamFactory attachmentStreamFactory,
-                                                                 List<Attachment> attachments)
+                                                                 Map<String, Attachment> attachments)
         throws AttachmentException {
-        ArrayList<PreparedAttachment> list = new ArrayList<PreparedAttachment>();
-        for (Attachment a : attachments) {
-            PreparedAttachment pa = AttachmentManager.prepareAttachment(attachmentsDir, attachmentStreamFactory, a);
-            list.add(pa);
+        Map<String, PreparedAttachment> list = new HashMap<String, PreparedAttachment>();
```
ricellis: `preparedAttachments`?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/Attachment.java
tomblench: I've gone for 4 in 667bdaf - we don't need to compare `Attachment`s for equality.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/UnsavedFileAttachment.java
tomblench: As above, kept with default implementations

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/UnsavedStreamAttachment.java
tomblench: As above, kept with default implementations

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/AttachmentManager.java
tomblench: Fixed names in 667bdaf

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/AttachmentManager.java
tomblench: Fixed names in 64e3595

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/AttachmentManager.java
tomblench: Fixed names in 64e3595

tomblench:
@ricellis PR feedback implemented in 64e3595 - see log message in commit.

ricellis:
Changes look good, but need to look into the test failure.

PR #494, opened 2017-01-30
Removed enumValue method from API

ricellis:
*What*

Removed `enumValue` method from `IndexType` API.

*Why*

The intent of the API is that users use the enum constants, not strings. The `enumValue` is
used internally to convert from string to enum, but need not be exposed as API.

*How*

Removed `IndexType.enumValue()` method.
Replaced single call site with `IndexType.valueOf(value.toUpperCase(Local.ENGLISH))`

*Testing*

No new tests, existing tests pass.

*Issues*

Part of #413

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/callables/ListIndexesCallable.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/callables/ListIndexesCallable.java
@@ -47,9 +47,7 @@ public class ListIndexesCallable implements SQLCallable<List<Index>> {
             while (cursorIndexNames.moveToNext()) {
                 String indexName = cursorIndexNames.getString(0);
                 String sqlIndexes = String.format("SELECT index_type, field_name, index_settings " +
-                        "FROM %s " +
-                                "WHERE index_name = ?",
-                        QueryImpl.INDEX_METADATA_TABLE_NAME);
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/query/callables/ListIndexesCallable.java
```diff
@@ -37,14 +38,16 @@
...
+        String sqlIndexNames = String.format("SELECT DISTINCT index_name FROM %s", QueryImpl
+                .INDEX_METADATA_TABLE_NAME);
         Cursor cursorIndexNames = null;
         ArrayList<Index> indexes = new ArrayList<Index>();
         try {
             cursorIndexNames = db.rawQuery(sqlIndexNames, new String[]{});
             while (cursorIndexNames.moveToNext()) {
                 String indexName = cursorIndexNames.getString(0);
-                String sqlIndexes = String.format("SELECT index_type, field_name, index_settings FROM %s " +
+                String sqlIndexes = String.format("SELECT index_type, field_name, index_settings " +
```
tomblench: String has become awkwardly concatenated from reformatting

ricellis: Fixed in 457a633

PR #495, opened 2017-01-30
Refactored Changes to not expose internals

ricellis:
*What*

Refactored `Changes` to not expose internals as API.

*How*

Made a `Changes` interface and `ChangesImpl` so as to avoid exposing the constructor.
Removed `getIds()` and `size()` methods from `Changes` as they duplicated stuff that could be easily done via the `getResults()` list.
Stopped exposing `InternalDocumentRevision` on `getResults()` and associated internal API changes that result from that change.
Updated migration doc.

*Testing*

Refactored tests to match new API. Existing tests pass.

*Issues*

Part of #413

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/Changes.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/Changes.java
@@ -1,8 +1,6 @@
 /*
  * Copyright © 2017 IBM Corp. All rights reserved.
  *
- * Copyright © 2013 Cloudant, Inc. All rights reserved.
- *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
  *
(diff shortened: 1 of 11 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/Changes.java
```diff
@@ -1,64 +1,8 @@
...
-     *
-     * @api_private
-     */
-    public Changes(long lastSequence, List<InternalDocumentRevision> results) {
-        Misc.checkNotNull(results, "Changes results");
-        this.lastSequence = lastSequence;
-        this.results = results;
-    }
-
+public interface Changes {
```
tomblench: Needs copyright header

doc/migration.md
```diff
@@ -242,3 +242,25 @@ Situations for which runtime (unchecked) exceptions may be thrown include:
...
+Before:
+```java
+getChanges().size();
+getChanges().getIds();
+```
+After:
+```java
+getChanges().getResults().size();
+// Get a list of IDs using a collector
+getChanges().getResults().stream().map(DocumentRevision::getId).collect(Collectors.toList());
```
tomblench: We should specifically point out that this is a java 1.8 (?) feature

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/Changes.java
ricellis: Fixed in 74ea1f87f85c4ffd726882f9d3ee3e7e1187a056

doc/migration.md
ricellis: Added in bd4e832c435c3fedc8c9de572ee26cf93c2dbd57

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/ChangesImpl.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+
+/**
+ * <p>The {@code Changes} object describes a list of changes to the {@link Database}.</p>
+ *
+ * <p>The object contains a list of the changes between a given sequence number
+ * (passed to the {@link Database#changes(long, int)} method) and
+ * the {@link Changes#lastSequence} field of the object.</p>
+ *
+ * @api_public
```
brynh: `@api_private`

ricellis: Good spot. I've removed the tags in d5bca4738a84c73bee40eba66af565b22b1c09fa as per the approach we'll take in #499.

PR #498, opened 2017-01-31
413 replication

ricellis:
*What*

Replication policies API changes for 2.0.0

*Why*

The `ReplicationPolicyManager.ReplicationListener` class was `public` (although marked API private) so that the `EventBus` could invoke it, but it is not API so moved into a class of its own and into the `internal` package.

There were two nearly identical listener classes in the replication policies API. `ReplicationService.ReplicationCompleteListener` and `ReplicationPolicyManager.ReplicationsCompletedListener`.

*How*

Extracted `ReplicationListener` class.
Removed `ReplciationCompleteListener` and replaced usages with `ReplicationsCompletedListener`
Moved `ReplicationService.SimpleReplicationCompleteListener` to
`ReplicationManager.SimpleReplicationsCompletedListener`.

Updated migration documentation.

*Testing*

No test changes needed, Existing tests pass.

*Issues*

Part of #413

### DIFF
--- a/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
+++ b/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
@@ -26,7 +26,6 @@ import android.os.IBinder;
 import android.os.Looper;
 import android.os.Message;
 import android.support.v4.content.WakefulBroadcastReceiver;
-import android.util.Log;
 
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -45,7 +44,9 @@ public abstract class ReplicationService extends Service
 
     public static final String EXTRA_INTENT = "intent";
 
-    /** The key used to identify the command being sent to the service. */
+    /**
+     * The key used to identify the command being sent to the service.
+     */
     public static final String EXTRA_COMMAND = "command";
 
     public static final int COMMAND_NONE = -1;
@@ -53,18 +54,20 @@ public abstract class ReplicationService extends Service
     /**
      * To start replications, this value should be passed to this service in an Intent using an
      * Extra with the {@link #EXTRA_COMMAND} key.
+     *
      * @see
      * <a href="http://github.com/cloudant/sync-android/blob/master/doc/replication-policies.md#controlling-the-replication-service">
-     *     Replication Policy User Guide</a> for more details.
+     * Replication Policy User Guide</a> for more details.
      */
     public static final int COMMAND_START_REPLICATION = 0;
 
     /**
      * To stop replications, this value should be passed to this service in an Intent using an
      * Extra with the {@link #EXTRA_COMMAND} key.
+     *
      * @see
      * <a href="http://github.com/cloudant/sync-android/blob/master/doc/replication-policies.md#controlling-the-replication-service">
-     *     Replication Policy User Guide</a> for more details.
+     * Replication Policy User Guide</a> for more details.
      */
     public static final int COMMAND_STOP_REPLICATION = 1;
 
@@ -77,11 +80,13 @@ public abstract class ReplicationService extends Service
     private List<Message> mCommandQueue = new ArrayList<Message>();
 
     /**
-     * Stores the set of {@link ReplicationCompleteListener}s listening for replication complete
+     * Stores the set of {@link ReplicationPolicyManager.ReplicationsCompletedListener}s
+     * listening for replication complete
      * events. Note that all modifications or iterations over mListeners should be protected by
      * synchronization on the mListeners object.
      */
-    private final Set<ReplicationCompleteListener> mListeners = new HashSet<ReplicationCompleteListener>();
+    private final Set<ReplicationPolicyManager.ReplicationsCompletedListener> mListeners = new
+            HashSet<ReplicationPolicyManager.ReplicationsCompletedListener>();
 
     // It's safest to assume we could be transferring a large amount of data in a
     // replication, so we want a high performance WiFi connection even though it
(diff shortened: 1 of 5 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/ReplicationListener.java
```diff
@@ -0,0 +1,75 @@
...
+import com.cloudant.sync.event.notifications.ReplicationErrored;
+import com.cloudant.sync.replication.ReplicationPolicyManager;
+import com.cloudant.sync.replication.Replicator;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * This class is not intended as API, it is public for EventBus access only.
+ * API consumers should not call these methods. See #setReplicationsCompletedListener for more
```
brynh: There are no comments on `setReplicationsCompletedListener` in this class to see. Maybe we should just remove this sentence, as even the comments on `ReplicationPolicyManager#setReplicationsCompletedListener` aren't that useful in aiding understanding about replication lifecycle events.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
```diff
@@ -155,10 +121,11 @@ public ServiceHandler(Looper looper) {
...
             try {
                 if (msg.arg2 > INTERNALLY_RESERVED_COMMAND_MAX_USED && msg.arg2 <=
-                    INTERNALLY_RESERVED_COMMAND_MAX) {
+                        INTERNALLY_RESERVED_COMMAND_MAX) {
                     throw new RuntimeException("ReplicationService received an EXTRA_COMMAND " +
-                        "using an id in the range reserved for internal use. Custom commands must" +
-                        " use an id above " + INTERNALLY_RESERVED_COMMAND_MAX);
+                            "using an id in the range reserved for internal use. Custom commands " +
+                            "must" +
+                            " use an id above " + INTERNALLY_RESERVED_COMMAND_MAX);
```
brynh: I'd prefer to combine this with the previous line.

doc/replication-policies.md
```diff
@@ -337,7 +337,7 @@ private ServiceConnection mConnection = new ServiceConnection() {
     @Override
     public void onServiceConnected(ComponentName name, IBinder service) {
         mReplicationService = ((ReplicationService.LocalBinder) service).getService();
-        mReplicationService.addListener(new ReplicationService.SimpleReplicationCompleteListener() {
+        mReplicationService.addListener(new ReplicationPolicyManager.SimpleReplicationCompleteListener() {
```
brynh: `SimpleReplicationsCompletedListener`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/ReplicationListener.java
```diff
@@ -0,0 +1,75 @@
...
+    public void add(Replicator replicator) {
+        synchronized (replicatorsInProgress) {
+            replicatorsInProgress.add(replicator);
+        }
+    }
+
+    public void remove(Replicator replicator) {
+        synchronized (replicatorsInProgress) {
+            if (replicatorsInProgress.remove(replicator)) {
+                replicator.getEventBus().unregister(this);
```
tomblench: As discussed there might be some subtle issues around eventBus registration and de-registration

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/ReplicationListener.java
ricellis: Fixed in d06de6bd6c02a55e24d09d038c08bf5388376be1 with some description about what the class is actually for.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/ReplicationListener.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+public class ReplicationListener {
+
+    private final Set<Replicator> replicatorsInProgress = new HashSet<Replicator>();
+    private ReplicationPolicyManager.ReplicationsCompletedListener mReplicationsCompletedListener
+            = null;
+
+    public void setReplicationsCompletedListener(ReplicationPolicyManager
+                                                         .ReplicationsCompletedListener
+                                                         mReplicationsCompletedListener) {
```
ricellis: Fixed in 07dfc26aec223d5ccfaf9ca96a4021c591fe2ef7

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/ReplicationService.java
ricellis: Fixed in 730a8d8e5098908a89fad52cc0db104b30e67896

doc/replication-policies.md
ricellis: Fixed in bb36ff48797c771d3d5f234ba0759468c5816548

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/ReplicationListener.java
ricellis: I think we got a bit muddled by the different listeners in play here. I'm confident this is actually ok because:
1. An instance of `this` `ReplicationListener` is registered on each `Replicator`'s own `EventBus` in the loop in `ReplicationPolicyManager.startReplications()`.
1. The `unregister` here in `remove` is the counterpart to that registration for when `stopReplications`  has been called or the replication complete/errored events have fired.

Unlike the replicator's `EventBus` for each `ReplicationPolicyManager` there can only be a single `ReplicationsCompletedListener` set. So I don't think there is any issue with multiple registrations. That said there is currently nothing to stop somebody changing the `ReplicationsCompletedListener` at any given point (or setting it to `null`), but the instance of `ReplicationListener` would not change and it has `null` checks so I don't think this would actually break anything.

However, I do think this would be more clear if the registration/unregistration were done symmetrically. Also, looking at it further I don't think the `inProgress` check is necessary as calling start twice on the same replication is a no-op (unless it already finished, but if that was the case it would have fallen out of the `inProgress` set anyway). The tests fail if I change that because they assert the number of times start/stop were called, but we can still avoid the multiple calls without the `inProgress` method just by returning the set addition/removal result `boolean`.

@tomblench @brynh see 055a8bb69fc43ce931d3759e47f1ebcb11962520

brynh: @ricellis I think the changes you've made do, indeed, make it clearer. LGTM

PR #504, opened 2017-02-01
Fix a few typos, links and other minor errors.

brynh:
Fix broken link in `CHANGES.md`

Add full stop in `CrudSamples.java`

In the example in `replication.md`, Listener needs to be `public` so the event handler can invoke the listener.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
   there are a large number of breaking changes to package names,
   classes and methods. API users will need to make changes to their
   existing code in order to use this version of the library. Consult
-  the [migration guide](doc/migration.md) for a comprehensive list of
+  the [migration guide](https://github.com/cloudant/sync-android/blob/2.0.0/doc/migration.md) for a comprehensive list of
   changes and suggested strategies to migrate your code.
 
 - [BREAKING CHANGE] The `name` field has been removed from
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -117,7 +117,7 @@
   `MutableDocumentRevision` classes have been removed, in order to
   simplify use of API methods. All code using this library will need
   to be updated to use the `DocumentRevision` class. See the updated
-  [CRUD guide](https://github.com/cloudant/sync-android/blob/master/doc/crud.md)
+  [CRUD guide](https://github.com/cloudant/sync-android/blob/master/doc/CrudSamples.java)
```
ricellis: What about changing this link to point to the 1.0.0 tag? That is better than being broken, but not as strange as pointing a 1.0.0 change at a 2.0.0 document.
https://github.com/cloudant/sync-android/blob/1.0.0/doc/crud.md

brynh:
As agreed in our conversation in Slack, I've updated the links in `CHANGES.md` to point to fixed versions that were the versions released when the entry in `CHANGES.md` was written. The only minor issue with this is that until we actually release a version, the links for the latest changes will be broken - e.g. the changes for 2.0.0 link to where the documents will be when 2.0.0 is released, but they're not there yet.

ricellis:
I think that is reasonable the 2.0.0 changes are unreleased at this point in time.
I'm still +1

PR #505, opened 2017-02-01
416 create with history

### LINKED ISSUE #416: Provide API for 'forceInsert()'
In our project, we have the requirement to get documents from Server, and insert them to our local DB with same `revision` received from Server.

Use of `Replicator` and `filter functions`, is not suitable in our case.  Till now, we were using `forceInsert() ` method of `DatastoreImpl`. But later in the forum we got to know that, `DatastoreImpl` is not a public class to be used and `forceInsert()` was not supposed to be consumed by end user.

But for our project implementation, use of` forceInsert()` is very much crucial. So, we are requesting you to provide an **implementation** to access the same functionality of `forceInsert()`.

ricellis:
*What*

Provide an API for `forceInsert`/`new_edits=false`

Reviewers please feel free to suggest better names.

*Why*

Some advanced users wish to interact with the `DocumentStore` outside the scope of the typical use case APIs of CRUD, query, and replication.

*How*

* Enabling work:
    * Removed deprecated `forceInsert` method and refactored remaining calls to use `ForceInsertItem`.
    * Added `DocumentNotFoundException` to `database()#update(...)` as this case was missing.
    * Added possibility of deleting document by `update`
        * Since a document `delete` creates and empty revision provide a way of deleting via `update`.
        * Adds `DocumentRevision#setDeleted()` method, which is necessary for creating a deleted revision with history.
    * Fixed potential NPE for force inserting deleted revisions - A deleted revision should have an empty body, but if it was `null` the attempt to read the body bytes would NPE.
* New API:
    * Added package for advanced APIs `com.cloudant.sync.documentstore.advanced` with package documentation and usage warnings.
    * Added `advanced.Database` interface for advanced database methods.
    * Added `public advanced.Database advanced()` to `DocumentStore` to access the `advanced.Database` API.
    * Added `void createWithHistory(DocumentRevision documentRevision, int revisionsStart, List<String> revisionsIDs)` and implementation.
* Updated `CHANGES.md`

*Testing*

*Re-enabled `ForceInsertTest`
* Updated `ForceInsertTest` to use non-deprecated `forceInsert` method.
* Added test for `DocumentNotFoundException` during `update`.
* Added `MutableDocumentDeleteTest.deleteViaRevisionUpdate()`
* Added tests for `createWithHistory`
    * Extracted `DocumentStoreTestBase`.
    * Slimmed `DatastoreTestBase` to not create `DocumentStore`.
    * Renamed setup methods to avoid override.
    * Added `AdvancedAPITest` superclass.
    * Added `CreateWithHistoryTest`.

*Issues*

Fixes #416

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
@@ -1073,7 +1073,8 @@ public class DatabaseImpl implements Database {
 
     @Override
     public DocumentRevision update(final DocumentRevision rev)
-            throws AttachmentException, DocumentStoreException, ConflictException {
+            throws AttachmentException, DocumentNotFoundException, ConflictException,
+            DocumentStoreException  {
 
         Misc.checkNotNull(rev, "DocumentRevision");
         Misc.checkState(isOpen(), "Datastore is closed");
@@ -1112,6 +1113,8 @@ public class DatabaseImpl implements Database {
             throwCauseAs(e, InvalidDocumentException.class);
             // conflictexception if rev ID is not winning rev
             throwCauseAs(e, ConflictException.class);
+            // not found if tried to update something that doesn't exist
+            throwCauseAs(e, DocumentNotFoundException.class);
             String message = "Failed to update document";
             logger.log(Level.SEVERE, message, e);
             throw new DocumentStoreException(message, e.getCause());
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
@@ -557,20 +557,6 @@ public class DatabaseImpl implements Database {
     }
 
     /**
-     * This method has been deprecated and should not be used.
-     * @see #forceInsert(List)
-     */
-    @Deprecated
-    public void forceInsert(final InternalDocumentRevision rev,
-                            final List<String> revisionHistory,
-                            final Map<String, Object> attachments,
-                            final Map<String[], Map<String, PreparedAttachment>> preparedAttachments,
-                            final boolean pullAttachmentsInline) throws DocumentException {
-        forceInsert(Collections.singletonList(new ForceInsertItem(rev, revisionHistory,
-                attachments, preparedAttachments, pullAttachmentsInline)));
-    }
-
-    /**
      * <p>
      * Inserts one or more revisions of a document into the database. For efficiency, this is
      * performed as one database transaction.
@@ -674,22 +660,21 @@ public class DatabaseImpl implements Database {
      * <p>Equivalent to:</p>
      *
      * <code>
-     *    forceInsert(rev, Arrays.asList(revisionHistory), null, null, false);
+     * forceInsert(rev, Arrays.asList(revisionHistory), null, null, false);
      * </code>
      *
-     * @param rev A {@code DocumentRevision} containing the information for a revision
-     *            from a remote datastore.
+     * @param rev             A {@code DocumentRevision} containing the information for a revision
+     *                        from a remote datastore.
      * @param revisionHistory The history of the revision being inserted,
      *                        including the rev ID of {@code rev}. This list
      *                        needs to be sorted in ascending order
-     *
-     * @see DatabaseImpl#forceInsert(InternalDocumentRevision, List, Map, Map, boolean)
      * @throws DocumentException if there was an error inserting the revision into the database
      */
     public void forceInsert(InternalDocumentRevision rev, String... revisionHistory) throws
             DocumentException {
         Misc.checkState(this.isOpen(), "Database is closed");
-        this.forceInsert(rev, Arrays.asList(revisionHistory), null, null, false);
+        this.forceInsert(Collections.singletonList(new ForceInsertItem(rev, Arrays.asList
+                (revisionHistory), null, null, false)));
     }
 
     private boolean checkRevisionIsInCorrectOrder(List<String> revisionHistory) {
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
@@ -19,21 +19,27 @@
 
 package com.cloudant.sync.internal.documentstore;
 
-import com.cloudant.sync.documentstore.DocumentRevision;
-import com.cloudant.sync.internal.common.CouchUtils;
-import com.cloudant.sync.internal.common.ValueListMap;
 import com.cloudant.sync.documentstore.Attachment;
 import com.cloudant.sync.documentstore.AttachmentException;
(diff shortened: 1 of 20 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
```diff
@@ -1191,4 +1186,43 @@ public DocumentRevision delete(final DocumentRevision rev) throws
...
+                .setDeleted(revision.isDeleted())
+                .build();
+
+        Map<String[], Map<String, PreparedAttachment>> preparedAttachments = Collections
+                .singletonMap
+                        (new String[]{revision.getId(), revision.getRevision()}, AttachmentManager
+                                .prepareAttachments(attachmentsDir, attachmentStreamFactory,
+                                        attachments));
+
+        Map<String, Object> attachmentsMetadata = new HashMap<String, Object>();
```
tomblench: I don't think this does what you think it does. The `revision.getAttachments()` will return a `Map<String, Attachment>` but what we need is the metadata in the same format returned from the raw json, something like
```
"_attachments":{
  "attachmenta305ba51-4a93-4ef7-8b99-b64b3e633243" {
    "follows":true,"length":86,"content_type":"image/jpeg"}...
```

ricellis: Since the caller has to initialize the `DocumentRevision` that is passed to `createWithHistory` the idea was that the attachment metadata setting that you describe would be performed by the caller since we don't do any direct JSON parsing as part of this.

I tried to indicate this intent in the [javadoc example](https://github.com/cloudant/sync-android/blob/14ed6536a7312e9426f9330f2bd355a9ae755349/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/advanced/XDatabase.java#L60)

Perhaps that needs to be elaborated further.

ricellis: Also the [test](https://github.com/cloudant/sync-android/blob/14ed6536a7312e9426f9330f2bd355a9ae755349/cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/advanced/CreateWithHistoryTest.java#L339) shows how I expected it to work, but that doesn't mean it is right :)

ricellis: Clarified the code commenting around this in daac2ed0e1a399008a9d35f1f6938837275ec0c3.
Will follow with additional testing of existing attachments and more example doc of how to handle attachments.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/documentstore/advanced/XDatabase.java
```diff
@@ -0,0 +1,105 @@
...
+ * <P>
+ * \u26A0 Interacting with these methods is not required for the typical use cases. Use with extreme
+ * caution.
+ * </P>
+ */
+public interface XDatabase {
+
+    /**
+     * <P>
+     * \u26A0 Creates a new revision in the database with the specified revision history.
```
ricellis: Fixed in 26f3cba5e7805eb67e7bd8e63e61eeea7cfab91e

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
ricellis: Updated the documentation with more explicit instructions for attachments in 5b6f87db490219631383fbbbc49987d92bb7e297.
Added new attachments tests in 66092c6d007431627cb36cf77e0f2a6030d32355 and while I was in the test class added some additional test doc and improved some assertions in b92d86bb86be54857fb34b1b6347b4a0236355ca.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/advanced/CreateWithHistoryTest.java
```diff
@@ -0,0 +1,793 @@
...
+                .contains(TestDoc.ID));
+        // Assert that the specified revisions are present
+        for (DocumentRevision testDoc : testDocRevisions) {
+            Assert.assertTrue("The revision should be present in the document store",
+                    database.contains(TestDoc.ID, testDoc.getRevision()));
+        }
+    }
+
+    /**
+     * Asserts that the onl revisions present for a document are the ones expected.
```
brynh: `onl`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
```diff
@@ -1191,4 +1159,43 @@ public DocumentRevision delete(final DocumentRevision rev) throws
         }
     }
 
+    @Override
+    public void createWithHistory(DocumentRevision revision, int revisionsStart, List<String>
+            revisionsIDs) throws DocumentException {
+
+        Map<String, Attachment> attachments = revision.getAttachments();
```
tomblench: Check `revision` not null and `revisionsStart > 0?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
```diff
@@ -1191,4 +1159,43 @@ public DocumentRevision delete(final DocumentRevision rev) throws
...
+                                .prepareAttachments(attachmentsDir, attachmentStreamFactory,
+                                        attachments));
+
+        // Couch _revisions.ids are newest -> oldest without a generation, so we need to manipulate
+        // them for forceInsert to an ascending order list with generational prefix
+        Map<String, Object> revisions = new HashMap<String, Object>(2);
+        revisions.put("start", revisionsStart);
+        revisions.put("ids", Collections.list(Collections.enumeration(revisionsIDs)));
+        // Parse into a descending order list of rev IDs
+        List<String> revIDs = CouchUtils.parseCouchDBRevisionHistory(Collections.<String, Object>
```
tomblench: Looks like you are the only caller of this method - before your changes it wasn't being used. Note also that when doing pull replication and force inserting, `DatastoreWrapper.bulkInsert` uses the similar `DocumentRevsUtils.createRevisionIdHistory`.

My suggestion is that we should try and share the logic between these code paths and avoid this call site having to put values into a map just to satisfy another method (for instance, it would be better to have some sort of `List<String> getHistory(String start, List<Integer> ids)`)

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
ricellis: Checks and new tests added in 6a1c93789ead8c220b7e8696a559c33345de9952

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/DatabaseImpl.java
ricellis: Made a common path in 01e9e91eedeb8cb75fbaa9e5c74296442397a47a

PR #507, opened 2017-02-01
Add migration instructions for multiple PeriodicReplicationReceivers

brynh:
Add migration instructions for the special case of having had a `PeriodicReplicationReceiver` when using sync-android < 2.0.0 and now upgrading to >= 2.0.0 and having multiple `PeriodicReplicationReceiver`s. 

Note that if users never used replication policies before sync-android 2.0.0, they don't need to do anything. If users only have one replication policy, they don't need to do anything.

In order to provide a simple way of handling this, `PeriodicReplicationService#upgradePreferences` has been made `protected`. This seemed the easiest way of making this special case relatively easy to deal with.

I wasn't sure whether the example code adds anything to the clarity. Any thoughts?


ricellis:
>Note that if users never used replication policies before sync-android 2.0.0, they don't need to do anything.

I think it is worth calling this out in the migration guide. I don't think it is clear from the proposed doc that if you are only starting to use replication policies in 2.0.0 that you don't need to do the empty override (even though it says _"There is one use-case that requires special handling"_).

brynh:
@ricellis Suggested clarification added in cbca7ae.

PR #509, opened 2017-02-02
javadoc v2 fixes

emlaver:
Javadoc updates and fixes during v2 sync-android shakedown.

PR #510, opened 2017-02-02
Fix sample app and update for renames

brynh:
1. Fixed crash on startup of the todo-sync sample app (first commit).
1. Various updates to comments and docs for things that have been renamed in v2.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,7 +32,7 @@
   `doc/crud.md`) has been migrated to a
   [java source file](https://github.com/cloudant/sync-android/blob/2.0.0/doc/CrudSamples.java).
 
-- [NEW] `databaseWithAdvancedAPIs()` getter on `DocumentStore` for specialist advanced use cases.
+- [NEW] `advanced()` getter on `DocumentStore` for specialist advanced use cases.
   Adds support for creating specific document revisions with history.
 
 - [FIXED] Issue with double encoding of restricted URL characters in credentials when using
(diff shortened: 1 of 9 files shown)


doc/conflicts.md
```diff
@@ -10,7 +10,7 @@ MVCC data-model is used to do this. This page describes how it works.
 
 ## Replication
 
-The Sync datastore participates in master-less replication with
+The Sync document store participates in master-less replication with
```
emlaver: I've used `DocumentStore` in the docs/comments for my PR (#509) but I'm not sure if it's better than `document store`.  @brynh @ricellis @tomblench thoughts?

brynh: I think I just used `document store` partly because I was thinking in more abstract terms and I was probably influenced by the fact it was `datastore` and not `Datastore`. I think actually it makes sense to use `DocumentStore` as we are talking about the specific entity, so unless anyone disagrees I'll change it here so it's consistent with what you've done.

brynh: Fixed in 8ed5b18

CHANGES.md
```diff
@@ -32,7 +32,7 @@
   `doc/crud.md`) has been migrated to a
   [java source file](https://github.com/cloudant/sync-android/blob/2.0.0/doc/CrudSamples.java).
 
-- [NEW] `databaseWithAdvancedAPIs()` getter on `DocumentStore` for specialist advanced use cases.
+- [NEW] `advanced()` getter on `DocumentStore` for specialist advanced use cases.
```
ricellis: Thanks, sorry I missed this one!

PR #511, opened 2017-02-06
Extracted and renamed `ReplicationsCompletedListener`

ricellis:
*What*

Extracted and renamed `ReplicationsCompletedListener`

*Why*

As part of the v2 API changes two very similar interfaces were combined. However, the interface is shared by both the `ReplicationPolicyManager` and the Android service approaches to replication policies, so it could confusing to have it as an inner class of the `ReplicationPolicyManager`. As such it is moved into its own class file in the package `com.cloudant.sync.replication`.
However, in that package it would be confusing to have a `ReplicationsCompletedListener` that didn't work with all replications, so the name is changed to indicate it is for use with replication policies: `PolicyReplicationsCompletedListener`.
Similarly the `SimpleReplicationsCompletedListener` can be an inner class of the `PolicyReplicationsCompletedListener` interface instead of the `ReplicationPolicyManager`.

*How*

Moved `ReplicationPolicyManager.ReplicationsCompletedListener` ->
`PolicyReplicationsCompletedListener`.
Moved `ReplicationPolicyManager.SimpleReplicationsCompletedListener` ->
`PolicyReplicationsCompletedListener.SimpleListener`.
Updated docs to match.

*Testing*
Existing tests pass.

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PolicyReplicationsCompletedListener.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PolicyReplicationsCompletedListener.java
@@ -1,3 +1,17 @@
+/*
+ * Copyright © 2017 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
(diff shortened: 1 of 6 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PolicyReplicationsCompletedListener.java
```diff
@@ -0,0 +1,41 @@
+package com.cloudant.sync.replication;
```
brynh: Should probably have a copyright header.

ricellis: Yes, it should, annoyingly the IDE seems to miss that step when extracting classes instead of creating them from new - must be a tick box somewhere I need to find!

ricellis: Fixed in b69c805

PR #518, opened 2017-03-21
Named logcat files uniquely

ricellis:
*What*

Named logcat files uniquely

*How*

Now that there are multiple matrix test environments it helps to have uniquely named `logcat` files so that they all get collected and it is easier to identify which one is for which test run.

Added a `TEST_ENV_NAME` for a prefix, reverting to a UUID if there is no `TEST_ENV_NAME` set.

*Testing*

Log files uniquely named in test build, test failures related to intermittent Android device disconnection.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,7 +35,7 @@ def runTests(testEnv, isAndroid) {
                     junit '**/build/**/*.xml'
                     if (isAndroid) {
                         // Collect the device log
-                        archiveArtifacts artifacts: '**/build/**/*.log'
+                        archiveArtifacts artifacts: '**/build/**/*logcat.log'
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -60,10 +60,10 @@ stage('Build') {
 
 stage('QA') {
     // Define the matrix environments
-    def CLOUDANT_ENV = ['DB_HTTP=https', 'DB_HOST=clientlibs-test.cloudant.com', 'DB_PORT=443', 'DB_IGNORE_COMPACTION=true', 'CREDS_ID=clientlibs-test']
-    def COUCH1_6_ENV = ['DB_HTTP=http', 'DB_HOST=cloudantsync002.bristol.uk.ibm.com', 'DB_PORT=5984', 'DB_IGNORE_COMPACTION=false', 'CREDS_ID=couchdb']
-    def COUCH2_0_ENV = ['DB_HTTP=http', 'DB_HOST=cloudantsync002.bristol.uk.ibm.com', 'DB_PORT=5985', 'DB_IGNORE_COMPACTION=true', 'CREDS_ID=couchdb']
-    def CLOUDANT_LOCAL_ENV = ['DB_HTTP=http', 'DB_HOST=cloudantsync002.bristol.uk.ibm.com', 'DB_PORT=8081', 'DB_IGNORE_COMPACTION=true', 'CREDS_ID=couchdb']
+    def CLOUDANT_ENV = ['TEST_ENV_NAME=Cloudant_Test','DB_HTTP=https', 'DB_HOST=clientlibs-test.cloudant.com', 'DB_PORT=443', 'DB_IGNORE_COMPACTION=true', 'CREDS_ID=clientlibs-test']
```
tomblench: You could change these arrays to a hash of arrays where the key takes the place of `TEST_ENV_NAME`

Jenkinsfile
```diff
@@ -35,7 +35,7 @@ def runTests(testEnv, isAndroid) {
                     junit '**/build/**/*.xml'
                     if (isAndroid) {
                         // Collect the device log
-                        archiveArtifacts artifacts: '**/build/**/*.log'
+                        archiveArtifacts artifacts: '**/build/**/*logcat.log'
```
tomblench: Do we have a clean step somewhere which removes old logs?

ricellis: Yes :)
https://github.com/cloudant/sync-android/blob/b7a084a6cb980bc74428ca11abc4f89b67e6d52e/Jenkinsfile#L21

Jenkinsfile
ricellis: Well the axes is already has a key of name, but I couldn't find a way to access that value from inside the scope of the parallel task.
At the moment the env var means the value gets passed from the Jenkins env through to the gradle env. I think if we use a hash then we'll need an extra step to extract the key and set it as an env var anyway, unless I'm misunderstanding what you mean.

Issue #521, opened 2017-03-30
Calling getAttachments() on a deleted revision throws ClassCastException


brynh:
In sync-android v2.0.0, when calling `getAttachments()` on a deleted revision, a ClassCastException is thrown:
> `java.lang.ClassCastException: java.util.HashMap cannot be cast to com.cloudant.sync.internal.common.ChangeNotifyingMap`

Although one shouldn't be able to access attachments on a deleted revision, it would be much nicer if this returned `null` or an empty `Map` or a more appropriate exception.

A simple example demonstrating this issue is:
```java
File path = getApplicationContext().getDir("documentstores", Context.MODE_PRIVATE);

try {
    DocumentStore ds = DocumentStore.getInstance(new File(path, "my_document_store"));

    DocumentRevision revision = new DocumentRevision();
    Map<String, Object> body = new HashMap<String, Object>();
    body.put("animal", "cat");
    revision.setBody(DocumentBodyFactory.create(body));

    DocumentRevision saved = ds.database().create(revision);
    Map<String, Attachment> attachments = saved.getAttachments(); // <<< This works.

    DocumentRevision deleted = ds.database().delete(saved);
    attachments = deleted.getAttachments(); // <<< This throws a ClassCastException.

} catch (Exception e) {
    e.printStackTrace();
}
```

Relevant stack trace:
```console
java.lang.ClassCastException: java.util.HashMap cannot be cast to com.cloudant.sync.internal.common.ChangeNotifyingMap
at com.cloudant.sync.internal.documentstore.InternalDocumentRevision.getAttachments(InternalDocumentRevision.java:214)
at com.cloudant.sync.internal.documentstore.InternalDocumentRevision.getAttachments(InternalDocumentRevision.java:32)
...
```

ricellis:
Fixed by #522

PR #522, opened 2017-04-04
Fix deleted revision bug

tomblench:
See #521 

_What_
Fetching attachments on a deleted revision would cause an exception to be thrown.

_How_
Fix bug by setting attachments to an empty map in `DeleteDocumentCallable`.

_Testing_
Add test `getAttachmentsOnDeletedRevision`

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/callables/DeleteDocumentCallable.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/callables/DeleteDocumentCallable.java
@@ -14,10 +14,10 @@
 
 package com.cloudant.sync.internal.documentstore.callables;
 
+import com.cloudant.sync.documentstore.Attachment;
 import com.cloudant.sync.documentstore.ConflictException;
 import com.cloudant.sync.documentstore.DocumentStoreException;
 import com.cloudant.sync.internal.common.CouchUtils;
-import com.cloudant.sync.internal.documentstore.DatabaseImpl;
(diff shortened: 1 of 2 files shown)


cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/documentstore/AttachmentTest.java
```diff
@@ -411,4 +414,23 @@ public void attachmentOrderingTest() throws Exception {
         Matcher m = p.matcher(regex);
         Assert.assertTrue("Should match pattern", m.matches());
     }
+
+    // test for regression on fetching attachments after deleting revision
+    @Test
+    public void getAttachmentsOnDeletedRevision() throws Exception {
```
ricellis: Might be worth having a similar test that has some attachments before the deletion too, but I don't think it is essential as this already covers the `CCE` problem.

tomblench: Changed test so there are two variants - one with an initial attachment and one without - in a0fdf011

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/documentstore/callables/DeleteDocumentCallable.java
```diff
@@ -30,6 +31,8 @@
 import com.cloudant.sync.internal.util.Misc;
 
 import java.sql.SQLException;
+import java.util.Collections;
+import java.util.HashMap;
```
ricellis: This import may not be needed now.

PR #524, opened 2017-04-06
Fix 513

### LINKED ISSUE #513: WifiPeriodicReplicationReceiver sometimes incorrectly schedules replications
`WifiPeriodicReplicationReceiver` doesn't correctly schedule periodic replications after WiFi connection has been lost.

WiFi can be lost for a number of reasons, including:

1. User moves out of WiFi coverage;
1. Device turns off WiFi radio to save power (this can't happen during a replication as we hold a WiFi lock, but can happen between replications);
1. User turns off WiFi (including enabling airplane mode).

Because of case 2, we should never cancel the scheduled alarms, as even though we might not currently have WiFi, we might be able to get a WiFi connection if we wake the device up, so at the scheduled times, we should always see if we can get WiFi.

To fix this:

- `WifiPeriodicReplicationReceiver` should never send `PeriodicReplicationService.COMMAND_STOP_PERIODIC_REPLICATION` or `PeriodicReplicationService.COMMAND_START_PERIODIC_REPLICATION`. These should only be used by the user to control whether the replication policy is active.
- When loss of WiFi is detected, `WifiPeriodicReplicationReceiver` should only stop any in-progress replication by sending `ReplicationService.COMMAND_STOP_REPLICATION`. 
- When `WifiPeriodicReplicationReceiver` receives an alarm, it should only pass it on to the superclass if we're connected to WiFi.

It would also make sense to enhance `WifiPeriodicReplicationReceiver` so that when we connect to WiFi, we check if there was a replication in progress when we previously lost WiFi, and if so, we restart a replication immediately.

brynh:
Fix the issues raised in #513.

Additionally, ensure wake locks are released only after we've finished processing commands sent to the replication service. Previously wake locks were being released too early (after the command had been identified but before the processing of it had been completed). This would mean for a replication, the device could go back to sleep while a replication is in progress.

Add guidance to the replication policy guide on when to use replication policies and when to use Android's `JobScheduler`. Having attempted to integrate `JobScheduler` into replication policies it became clear that doing so, while retaining `JobScheduler`'s flexibility wasn't easy and there are also options developers can go with such as using the Firebase `JobDispatcher`. Furthermore, at some point we will probably support only Android API level 21 and above, at which point Replication Policies can be completely removed. It was therefore decided that giving guidance on when and how to use of  `JobScheduler` provided more flexibility than trying to build it into Replication Policies.

### Testing

Various tests have been added to check issues raised in #513.

Extensive manual testing of a version of the TodoSync app modified to use the WifiPeriodicReplicationReceiver has been made where instrumentation was added to log events in the replication policies to file. These files have then been regularly reviewed to ensure that the replication policy is behaving as expected.

Checks on battery stats have been made to ensure that the replication policies are not holding onto wake locks and draining the battery excessively.

The examples in the guidance for the `JobScheduler` have been manually tested to ensure they work as expected.

Fixes #513

### DIFF
--- a/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
+++ b/cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
@@ -47,7 +47,7 @@ public abstract class PeriodicReplicationService<T extends PeriodicReplicationRe
      * So that multiple PeriodicReplicationService instances can be used in one application without
      * interfering with each other, the preferences in this file are stored using keys prefixed with
      * the name of the concrete class implementing the PeriodicReplicationService. */
-    private static final String PREFERENCES_FILE_NAME = "com.cloudant.preferences";
+    static final String PREFERENCES_FILE_NAME = "com.cloudant.preferences";
 
     /* We store the elapsed time since booting at which the last alarm occurred in SharedPreferences
      * using a key with this suffix. This is used to set the initial alarm time when periodic
@@ -68,6 +68,14 @@ public abstract class PeriodicReplicationService<T extends PeriodicReplicationRe
      * SharedPreferences using a key with this suffix. We have to store the flag persistently as
      * the service may be stopped and started by the operating system. */
     private static final String EXPLICITLY_STOPPED_SUFFIX = ".explicitlyStopped";
+
+    /* We store a flag indicating whether there are replications pending in
+     * SharedPreferences using a key with this suffix. Replications may be pending because they
+     * are currently in progress and have not yet completed, or becasue a previous scheduled
+     * replication didn't take place because the conditions for replication were not met. We have
+     * to store the flag persistently as the service may be stopped and started by the operating
+     * system. */
+    private static final String REPLICATIONS_PENDING_SUFFIX = ".replicationsPending";
 
     private static final long MILLISECONDS_IN_SECOND = 1000L;
 
@@ -156,19 +164,29 @@ public abstract class PeriodicReplicationService<T extends PeriodicReplicationRe
 
         @Override
         public void handleMessage(Message msg) {
+            Intent intent = msg.getData().getParcelable(EXTRA_INTENT);
+
             switch (msg.arg2) {
                 case COMMAND_START_PERIODIC_REPLICATION:
                     startPeriodicReplication();
+                    releaseWakeLock(intent);
+                    stopSelf(msg.arg1);
                     break;
                 case COMMAND_STOP_PERIODIC_REPLICATION:
                     stopPeriodicReplication();
                     setExplicitlyStopped(true);
+                    releaseWakeLock(intent);
+                    stopSelf(msg.arg1);
                     break;
                 case COMMAND_DEVICE_REBOOTED:
                     resetAlarmDueTimesOnReboot();
+                    releaseWakeLock(intent);
+                    stopSelf(msg.arg1);
                     break;
                 case COMMAND_RESET_REPLICATION_TIMERS:
                     restartPeriodicReplications();
+                    releaseWakeLock(intent);
+                    stopSelf(msg.arg1);
                     break;
                 default:
                     // Do nothing
@@ -228,6 +246,7 @@ public abstract class PeriodicReplicationService<T extends PeriodicReplicationRe
     protected void startReplications() {
         super.startReplications();
         setLastAlarmTime(0);
+        setReplicationsPending(this, getClass(), true);
     }
 
     /** Start periodic replications. */
@@ -344,6 +363,14 @@ public abstract class PeriodicReplicationService<T extends PeriodicReplicationRe
             false);
     }
 
+    public static boolean isPeriodicReplicationEnabled(Context context,
+                                                        Class <? extends
+                                                            PeriodicReplicationService> prsClass) {
+        SharedPreferences prefs = context.getSharedPreferences(PREFERENCES_FILE_NAME, Context
+            .MODE_PRIVATE);
+        return prefs.getBoolean(constructKey(prsClass, PERIODIC_REPLICATION_ENABLED_SUFFIX), false);
+    }
+
     /**
      * Set a flag in SharedPreferences to indicate whether periodic replications were explicitly
      * stopped.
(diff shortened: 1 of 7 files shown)


cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
```diff
@@ -417,8 +444,51 @@ private int getIntervalInSeconds() {
...
+
+    /**
+     * Sets whether there are replications pending. This may be because replications are
+     * currently in progress and have not yet completed, or because a previous scheduled
+     * replication didn't take place because the conditions for replication were not met.
+     * @param context
+     * @param prsClass
+     * @param pending true if there is a replication pending, or false otherwise.
+     */
+    public static void setReplicationsPending(Context context, Class<? extends
```
tomblench: It's my understanding that the `replicationsPending` state as stored in shared preferences will persist across reboots. What's the impact of restarting the device when there are still replications pending, is it a problem or will they automatically resume and eventually set `replicationsPending` to `false`?

doc/replication-policies.md
```diff
@@ -385,3 +406,263 @@ protected void onStop() {
...
+```
+
+## Mixing JobScheduler and Replication Policies
+
+If you want to use a combination of Replication Policies and `JobScheduler`, follow the [Replication Policy](#android-replication-policies)
+guidance above and everywhere you send a message to your `ReplicationService`, make a conditional call dependent on API
+level to start/cancel/restart the `JobScheduler` with your `JobService` with any necessary changes to the configuration
+of your `JobService`.
+
+To make this easier, we can put the code to start and cancel our JobScheduler tasks in methods. You may wish to make them
```
brynh: I've changed it and gone with slightly different wording in 48f5b55.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
brynh: Your understanding is correct, and indeed, there was a small bug (fixed in 48f5b55) that meant pending replications wouldn't always be resumed immediately we reconnected to WiFi following a reboot because the persistent `wasOnWifi` flag stored in shared preferences wasn't being reset following reboot. Under this circumstance, replications sometimes wouldn't resume until the next scheduled period. However, I think under most circumstances when the device turned off, we would probably have got an event indicating we were disconnecting from WiFi as the device shutdown, so I think it's likely this bug would only affect a device crash or forced shutdown, and everything would have got back to normal following the next scheduled replication.

Thanks for raising this issue and making me go back to look at it.

I've fixed this by setting the `wasOnWifi` flag to `false` when we get notification of the reboot. This will be safe as we'll always get the notification of the reboot before we get notification that the device has connected to WiFi following the reboot.

Now, with this fix, if the device reboots when there are still replications pending:

1. `WifiPeriodicReplicationReceiver` will receive an Intent with action `Intent.ACTION_BOOT_COMPLETED` and we'll make sure the previous state of the WiFi is set so that it shows we weren't on WiFi.
1. Then WiFi will either be available or not available:
    * If WiFi is available when the device restarts, `WifiPeriodicReplicationReceiver` will then receive an Intent with action `ConnectivityManager.CONNECTIVITY_ACTION`. It will then go into the `else if (isConnectedToWifi)` branch and check `replicationsPending` and immediately resume the replication.
    * If WiFi was not available when the device restarted, when we next connect to WiFi, the same will happen. If there are any `ALARM_ACTION` intents received before the `ConnectivityManager.CONNECTIVITY_ACTION` indicating we're back on WiFi, the `replicationsPending` will be re-set to true (although it will already be true anyway).
1. When the replications complete, the `allReplicationsCompleted()` callback in `PeriodicReplicationService` will set `replicationsPending` to `false`.

doc/replication-policies.md
```diff
@@ -14,6 +14,27 @@ intervals.
...
 
+If your Android app is targeting Android 5.0 (API level 21) or above it is recommended that you use Android's [`JobScheduler`](https://developer.android.com/reference/android/app/job/JobScheduler.html)
+rather than the Replication Policies described in this section. Using the `JobScheduler` you can set various conditions
+for replication much like with Replication Policies and they make even more efficient use of device resources than Replication Policies.
+See the section [Replication using the JobScheduler](#replication-using-the-jobscheduler) for examples and advice regarding using
+the `JobScheduler`.
+
+If your app is targeting Android 7.0 (API level 24) or above, the `WifiPeriodicReplicationReceiver` will not work correctly because
+apps declaring a `BroadcastReceiver` in their manifest no longer receive the `android.net.conn.CONNECTIVITY_CHANGE` event. In
+this case, you should consider using `JobScheduler`.
```
ricellis: I think we should say `must` rather than `should consider` as it sounds like there isn't another choice.

cloudant-sync-datastore-android/src/main/java/com/cloudant/sync/replication/PeriodicReplicationService.java
```diff
@@ -156,19 +164,29 @@ public ServiceHandler(Looper looper) {
 
         @Override
         public void handleMessage(Message msg) {
+            Intent intent = msg.getData().getParcelable(EXTRA_INTENT);
+
             switch (msg.arg2) {
                 case COMMAND_START_PERIODIC_REPLICATION:
                     startPeriodicReplication();
+                    releaseWakeLock(intent);
```
brynh: On the face of it, that sounds like a nice idea, but we don't want to call `releaseWakeLock` or `stopSelf` in the `default` case, as the processing gets passed on to the `ReplicationService.ServiceHandler` in the superclass. There, when we handle a `COMMAND_START_REPLICATION` message, we don't want to release the wake lock or stop the service when we've processed the message in the `handleMessage()` method.  We need to wait until the replications have actually completed and the `allReplicationsCompleted()` callback is executed before we release the wake lock and stop the service.

I think even if we move the call to `super.handleMessage(msg)` to the `default`, using a `finally` block is problematic as if we did likewise in the superclass we'd end up attempting to release the wake lock and stop the service twice for the cases handled by the superclass.

I can't think of a clean way of doing it using `finally`.

doc/replication-policies.md
brynh: Fixed in df7b4a7

PR #548, opened 2017-10-04
Target cloudant-http 2.11.0

tomblench:
This is anticipated to be the next release version which will include IAM interceptors.

~~This branch will only build cleanly and be mergeable once the next version of `cloudant-http` (anticipated to be 2.10.0) is released.~~

Now targetting 2.11.0


tomblench:
@ricellis OK added CHANGES

PR #554, opened 2017-11-09
552 add proguard in docs

### LINKED ISSUE #552: Proguard on Android
I implemented sync-android on my Android app following the sample todo app. 
My app is written in **Kotlin**. Not sure if this is important but stating it here anyway.
Everything worked fine until I enabled Proguard. Since there was no mention of Proguard in the README or docs, I followed the Proguard config file from the sample app as below:

```
-keepattributes Signature,*Annotation*,EnclosingMethod
-keepattributes InnerClasses
-keepattributes SourceFile,LineNumberTable
-keepnames class com.fasterxml.jackson.** { *; }
-keep class org.w3c.dom.bootstrap.DOMImplementationRegistry
-keep class javax.inject.** { *; }
-keep class com.cloudant.** { *; }
-dontwarn com.cloudant.http.**
-dontwarn com.fasterxml.jackson.**
```

However, I realized that the completion event was not being fired. After much debugging, I figured out that the following methods (from [TaskModel.java](https://github.com/cloudant/sync-android/blob/master/sample/todo-sync/src/com/cloudant/todo/TasksModel.java)) were being removed:
```
    /** javadoc comment */
    @Subscribe
    public void complete(ReplicationCompleted rc) {
        // handler code..
    }

    /** javadoc comment */
    @Subscribe
    public void error(ReplicationErrored re) {
       // handler code..
    }
```

So, I went ahead and set a rule to keep all the public methods etc within my class:
```
-keep class com.azri.myapp.service.DataModel {
  public *;
}
```

And fixed!

My issue now is:

1. Did I miss something else that was obvious that I shouldn't have to set a different proguard rule?
2. If I did the right things, maybe this should be mentioned explicitly in the README or docs? This is usually done for a lot of libraries I've seen.

Thanks! 😄

emlaver:
## What

Add details from #552 into Events.md documentation.

## How

-  Updated events.md with Rich's note about ProGuard
- Added note in `todo-sync` README about additional `keep` in `proguard.cfg`

## Issues

fixes #552

### DIFF
--- a/sample/todo-sync/README.md
+++ b/sample/todo-sync/README.md
@@ -88,3 +88,6 @@ Once installed, the default settings entered in the XML file above should appear
 To test your connection, add a couple of tasks and hit "Upload (Push)" from the menu in the top right. You should see the JSON documents appear in your Cloudant database. Changes to the documents in the Cloudant database will be replicated back to the device when you tap "Download (Pull)".
 
 If you see "Replication Error" rather than "Replication Complete" as a popup message, check the logs using `adb` to see what the exception was.
+
+**Note:** If you require ProGuard and your app has a different namespace, you'll need an additional `keep` in `proguard.cfg` to preserve the `@Subscribe` methods that are reflectively invoked.
+
--- a/doc/events.md
+++ b/doc/events.md
@@ -62,9 +62,11 @@ For `DocumentStore` events there is a similar superclass: `DocumentStoreModified
 
 Whether to use these generic events or one method per event type is simply a matter of style.
 
-Note that subscribing methods are invoked synchronously. This means that it is important to
+Note that subscribing methods are:
+- invoked synchronously. This means that it is important to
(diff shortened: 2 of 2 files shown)


sample/todo-sync/README.md
```diff
@@ -88,3 +88,6 @@ Once installed, the default settings entered in the XML file above should appear
 To test your connection, add a couple of tasks and hit "Upload (Push)" from the menu in the top right. You should see the JSON documents appear in your Cloudant database. Changes to the documents in the Cloudant database will be replicated back to the device when you tap "Download (Pull)".
 
 If you see "Replication Error" rather than "Replication Complete" as a popup message, check the logs using `adb` to see what the exception was.
+
+**Note:** If you require Proguard and your app has a different namespace, you'll need an additional `keep` in `proguard.cfg` to preserve the `@Subscribe` methods that are reflectively invoked.
```
ricellis: Slight mis-casing here should be `ProGuard`.

PR #562, opened 2017-12-18
Performance: use efficient sub-query to find leafs

tomblench:
## What

Make replication of large and/or conflicted data sets faster by improving performance of `PickWinningRevisionCallable`, which is called for every new revision pulled.

## How

Analysis with `jvisualvm` and sqlite `explain query plan` revealed that the sub-query `(SELECT DISTINCT parent FROM revs WHERE parent NOT NULL)` forced a table scan because it could not use indexes to make this sub-query.

Changing this and other similar sub-queries to be as specific as possible by discriminating on revision id or numeric document id allows the query to succeed without a scan, without affecting correctness.

## Testing

Existing tests pass. Manual testing with large database (largedb1g on clientlibs-test account) shows a dramatic speedup. Previously each batch would take an exponentially longer time to insert due to the slow-down in picking the winning revision.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+# Unreleased
+- [IMPROVED] Improved efficiency of sub-query when picking winning
+  revisions. This improves performance when inserting revisions,
+  including during pull replication.
+
 # 2.1.0 (2017-12-04)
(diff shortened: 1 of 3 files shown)


ricellis:
and copyright updates

Issue #560, opened 2017-12-14
Is there any plan to include selectors for filtering _changes in the replication process?


jjrodrig:
We are working on a mobile application that uses Cloudant Sync. So far, it has been very successful. Thanks for your work. 

We are using the filtered replication mechanism based on a filter function for the Pull replication process. The filter function in CouchDB/Cloudant is know to be quite inefficient as it has process the complete database in order to obtain the result. 

CouchDB 2.0/Cloudant NoSQL DB support _selectors in the filtering of the database changes. Selector can be supported over Mango indexes and it can be much more efficient than the filter functions. 

Is there any plan to include support for this kind of filtering in Cloudant Sync?

Thanks,
Juanjo

- sync-android version(s) that are affected by this issue.
- sync-android artifact you are using
    - [X] cloudant-sync-datastore-android
    - [ ] cloudant-sync-datastore-android-encryption
    - [ ] cloudant-sync-datastore-javase
- Android or Java version (including vendor and platform).
- A small code sample that demonstrates the issue.

ricellis:
Right now we don't, but if people want to +1 this issue we can gauge demand for it.

PR #566, opened 2018-01-11
Add support for mango selectors in Pull replicator

jjrodrig:

- [x] You have added tests for any code changes

## What

 Support for mango selectors is added to the Pull Replicator. This allows to use a more efficent mechanism for changes filtering in CouchDB 2.x

## How

I've found many performance problems with filtered replication using JavaScript filters. This can be improved by using mango selectors. 

I've created a PullSelector class equivalent to PullFilter that can be used in the ReplicatorBuilder. 

## Testing

I've rewrite every pull replication filter test with the selector mechanism

## Issues

This a proposal for the enhancement #560


tomblench:
@jjrodrig thanks for your contribution. We are seeing some test failures because we test against some versions of CouchDB which do not support the mango selector feature. I'll suggest a work-around shortly.

tomblench:
@jjrodrig you may want to cherry-pick [this commit](https://github.com/tomblench/sync-android/commit/23c4476ac053933415a0452e3fdc37730ed6ee33) and add the new category to your tests which require Cloudant (as opposed to CouchDB 1.x).

This means that your tests will be skipped in Travis but will still run in our private Jenkins pipeline.

jjrodrig:
Thanks @tomblench, it is done. I've excluded RequireCloudant category alsy during test stage

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -311,12 +311,39 @@ public ChangesResult changes(String filterName, Map<String, String> filterParame
         return this.changes(options);
     }
 
+    public ChangesResult changes(String selector, Object since, Integer limit) {
+        Map<String, Object> options = getDefaultChangeFeedOptions();
```
tomblench: There's a lot of copy and paste between this and the existing constructor, maybe hive this all off to a utility function which builds the options?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/CouchClientWrapper.java
```diff
@@ -140,6 +141,14 @@ public ChangesResult changes(PullFilter filter, Object lastSequence, int limit)
         }
     }
 
+    public ChangesResult changes(PullSelector selector, Object lastSequence, int limit) {
```
tomblench: Should be marked `@Override`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullSelector.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright © 2015 IBM Corp. All rights reserved.
```
tomblench: Copyright should be 2018

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
```diff
@@ -100,18 +102,23 @@
...
                 responseInterceptors));
         this.targetDb = new DatastoreWrapper((DatabaseImpl) target);
         String replicatorName;
-        if (filter == null) {
-            replicatorName = String.format("%s <-- %s ", target.getPath(), source);
-        } else {
+        if (filter != null) {
             replicatorName = String.format("%s <-- %s (%s)", target.getPath(), source,
                     filter.getName());
+        } else if (selector != null) {
```
tomblench: The current logic means that if a filter and a selector are both set, the filter takes precedence. I don't think this makes sense and it's not obvious to the user.

Maybe it would be better to throw an exception either here or in the builder if both are set.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullSelector.java
```diff
@@ -0,0 +1,56 @@
...
+ * feed.</p>
+ *
+ * <p>Use this constructor:</p>
+ *
+ * <pre>
+ * PullSelector selector = new PullSelector("{...a selector...}")
+ * </pre>
+ *
+ * @see
+ * <a target="_blank" href="http://docs.couchdb.org/en/latest/replication/intro.html#controlling-which-documents-to-replicate">Controlling documents replicated</a>
```
tomblench: Link to the bluemix documentation where possible, in this case https://console.bluemix.net/docs/services/Cloudant/api/replication.html#selector-field

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullSelector.java
```diff
@@ -0,0 +1,56 @@
...
+ * @see
+ * <a target="_blank" href="http://docs.couchdb.org/en/latest/replication/replicator.html#selectorobj">Selector Objects</a>
```
tomblench: As above

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -311,12 +311,39 @@ public ChangesResult changes(String filterName, Map<String, String> filterParame
         return this.changes(options);
     }
 
+    public ChangesResult changes(String selector, Object since, Integer limit) {
```
tomblench: I think these two constructors are a bit confusing now, because initially there was one constructor with a filter name and parameters which could both be optional (which was probably a bad design in the first place...)

How about we have 3 constructors:
- No filter/No selector
- Filter
- Selector

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/PullStrategySelectorTest.java
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright © 2017 IBM Corp. All rights reserved.
```
tomblench: Please use the standard copyright which is just © 2018 IBM Corp. for new files.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
jjrodrig: Yes, it is really confusing, I'll implement the 3 methods:

ChangesResult changes(Object lastSequence, int limit);
ChangesResult changes(PullFilter filter, Object lastSequence, int limit);
ChangesResult changes(PullSelector selector, Object lastSequence, int limit);

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
jjrodrig: Ok, I'll do it in the builder

jjrodrig:
Thanks, I pushed the changes.

tomblench:
@jjrodrig as a general comment, it's very difficult to read your diff because your IDE is set to auto-format every file you edit which changes the whitespace and creates lots of noise in the diff.

jjrodrig:
Sorry, I think that I have configured the formatting as specified in the contributing guide. Is there something I miss?

I'll check it in order to be closer with the previous code format.

tomblench:
OK, don't worry it's not a big deal. There are a few comments where whitespace is significant though, like [this](https://github.com/cloudant/sync-android/pull/566/commits/a7ca3fc361532e85acb9011ce73f7f5672c2362f#diff-32c4df8ca0c70c2193dd0d4e3e0bd182L686), because it's easier to read the example json in the comment if it is correctly indented.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullSelector.java
```diff
@@ -0,0 +1,56 @@
+/*
```
tomblench: I don't think this class is really necessary because all it does is wrap a `String`. I think you should remove it and just use `String` directly where this is currently used.

jjrodrig:
Thanks @tomblench

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/ReplicatorBuilder.java
```diff
@@ -301,10 +310,30 @@ public Pull filter(PullFilter pullPullFilter) {
             return this;
         }
 
+        /**
+         * <p>Provides the mango query selector for filter criteria to be used
+         * when a pull replication calls the source database's {@code _changes} feed.
+         * </p>
+         *
```
ricellis: An additional comment should be added in here to point out that this method is only valid for replications against CouchDB 2.x APIs since the `_selector` is not available for `_changes` in CouchDB 1.x.

PR #576, opened 2018-03-26
Force tls1.2 on old android

ricellis:

- [x] You have added tests for any code changes - manual test

## What

Added a custom `SSLSocketFactory` for enabling TLS1.2 on Android API < 20.

## How

Android API >= 20 already has TLS1.2 enabled by default. API level 19 has it available but not enabled. On older versions of Android it is possible for users to workaround this, for example by configuring the `HttpsUrlConnection.setDefaultSocketFactory` or similar. However, we can check the API version and set the socket factory for each `HttpConnection` automatically so that older versions of Android work without any user required changes.

* Added Android version and TLS support checks as well as a TLSv1.2 only `SSLSocketFactory` to enable on older versions.
* Moved default interceptors into `CouchClient` from `ReplicatorBuilder`.

## Testing

* Updated `HttpTest` that called `HttpConnection` directly to consume default interceptors.
* Refactored `HttpTest` slightly to simplify interceptor addition.
* Updated replication tests for new assertions, due to interceptor move to `CouchClient` and interceptor "sizes".

Existing tests pass.
Manually tested on API 19 emulator against TLS1.2 only Cloudant.
CI tests run against API >= 20, so don't regularly check this. We could consider an extra branch of testing using the emulator at API level 19.

## Issues

N/A

### DIFF
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
@@ -25,6 +25,8 @@ import com.cloudant.http.Http;
 import com.cloudant.http.HttpConnection;
 import com.cloudant.http.HttpConnectionRequestInterceptor;
 import com.cloudant.http.HttpConnectionResponseInterceptor;
+import com.cloudant.http.internal.interceptors.SSLCustomizerInterceptor;
+import com.cloudant.http.internal.interceptors.UserAgentInterceptor;
 import com.cloudant.sync.internal.common.RetriableTask;
 import com.cloudant.sync.internal.documentstore.DocumentRevsList;
 import com.cloudant.sync.internal.documentstore.MultipartAttachmentWriter;
@@ -39,11 +41,16 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.lang.reflect.Type;
+import java.net.InetAddress;
+import java.net.Socket;
 import java.net.URI;
 import java.nio.charset.Charset;
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -52,7 +59,24 @@ import java.util.concurrent.Callable;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocket;
+import javax.net.ssl.SSLSocketFactory;
+
 public class CouchClient {
+
+    public static final List<HttpConnectionRequestInterceptor> DEFAULT_REQUEST_INTERCEPTORS;
+
+     // Set up the defaults
+     static {
+         HttpConnectionRequestInterceptor ua_interceptor =
+                 new UserAgentInterceptor(CouchClient.class.getClassLoader(),
+                         "META-INF/com.cloudant.sync.client.properties");
+         HttpConnectionRequestInterceptor tlsInterceptor = checkAndGetTlsInterceptor();
+         DEFAULT_REQUEST_INTERCEPTORS = (tlsInterceptor != null) ?
+                 Arrays.asList(ua_interceptor, tlsInterceptor) :
+                 Collections.singletonList(ua_interceptor);
+     }
 
     private CouchURIHelper uriHelper;
     private List<HttpConnectionRequestInterceptor> requestInterceptors;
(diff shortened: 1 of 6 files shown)


cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -52,8 +60,28 @@
...
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocket;
+import javax.net.ssl.SSLSocketFactory;
+
 public class CouchClient {
 
+    private static final List<HttpConnectionRequestInterceptor> DEFAULT_REQUEST_INTERCEPTORS =
+            new ArrayList<HttpConnectionRequestInterceptor>();
+
+    private static final UserAgentInterceptor USER_AGENT_INTERCEPTOR =
```
tomblench: Can `USER_AGENT_INTERCEPTOR` and `TLS_INTERCEPTOR` go into the `static` block so they don't pollute the top-level fields?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -75,6 +105,145 @@ public CouchClient(URI rootUri,
...
+
+        private Socket tlsOnlySocket(Socket s) {
+            if (s instanceof SSLSocket) {
+                ((SSLSocket) s).setEnabledProtocols(TLSv12Only);
+            }
+            return s;
+        }
+    }
+
+    private static class Special extends SSLCustomizerInterceptor {
```
tomblench: Is this class only used for debugging?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -75,6 +105,145 @@ public CouchClient(URI rootUri,
...
+                        // the default, but there isn't really a way to check this, equals is not
+                        // overridden on the SSLSocketFactory and there is no route back to the
+                        // SSLContext that generated the SSLSocketFactory.
+                        SSLContext sc = SSLContext.getInstance(tlsv1_2);
+                        sc.init(null, null, null);
+                        // We add this interceptor, but it could be overridden by a later user
+                        // provided interceptor, so this shouldn't change anyone's existing
+                        // customizations.
+                        SSLSocketFactory s = sc.getSocketFactory();
+                        return new Special(s);
```
tomblench: Can you remove the `Special` class (see below), and just use `SSLCustomizerInterceptor` directly?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -75,6 +105,145 @@ public CouchClient(URI rootUri,
...
+            } catch (KeyManagementException e) {
+                throw new RuntimeException(e);
+            }
+        }
+        return null;
+    }
+
+    private static final class TlsOnlySslSocketFactory extends SSLSocketFactory {
+
+        private static final String[] TLSv12Only = new String[]{"TLSv1.2"};
```
tomblench: This string also occurs in `checkAndGetTlsInterceptor`, maybe have a static const somewhere?

cloudant-sync-datastore-core/src/test/java/com/cloudant/http/HttpTest.java
```diff
@@ -40,6 +43,36 @@
...
+        conn.setRequestBody(new HttpConnection.InputStreamGenerator() {
+            @Override
+            public InputStream getInputStream() {
+                return bis;
+            }
+        });
+        // This test invokes HttpConnection directly rather than via the CouchClient, so we need to
+        // force the addition of some default interceptors, which we obtain reflectively for the
+        // purposes of these tests.
+        Field interceptors = CouchClient.class.getDeclaredField("DEFAULT_REQUEST_INTERCEPTORS");
```
tomblench: Since `CouchClient` is internal, I don't mind the field `DEFAULT_REQUEST_INTERCEPTORS` being public if we are going to use it in a few other places, even if that's in test code - but it's up to you.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
ricellis: Yes, done in 72fef696319034370794c5137617a11e5c9bea81

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
ricellis: Yes, I intended to remove it, sorry. Done in 5b23adcf577db53b8ed66e4ab6b8ade7c8e4ff4a

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
ricellis: Yes, as above, done in 5b23adcf577db53b8ed66e4ab6b8ade7c8e4ff4a

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
ricellis: Done in 9d877ca53d472c12993fada59740b3c77880b226

cloudant-sync-datastore-core/src/test/java/com/cloudant/http/HttpTest.java
ricellis: Findbugs flagged it as malicious because it was `public` and a mutable list, but now that I've changed the initialization the list is no longer mutable, so happy to make it public again.
Done in ae17649d66c5827add7df0d5d7b75c2d13a5d11d

PR #581, opened 2018-06-19
Template updates

ricellis:

- [x] You have added tests for any code changes

## What

Updated issue and PR templates and contributing file.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,25 +4,15 @@ Please [read these guidelines](http://ibm.biz/cdt-issue-guide) before opening an
 
 ## Bug Description
 
-### 1. Steps to reproduce and a small code sample that demonstrates the issue
+### 1. Steps to reproduce and the simplest code sample possible to demonstrate the issue
 <!--
-EXAMPLE:
-1. Initialize client
-2. List DBs
-
-cc = new CloudantClient(...)
-dbNames = cc.allDbs()
+Outline the steps you take to make the problem happen.
+Provide the simplest code sample you can, in context, that reproduces the issue.
 -->
 
 ### 2. What you expected to happen
-<!--
-Expected to get a list of all databases
--->
 
 ### 3. What actually happened
-<!--
-Got a 400 bad request
--->
 
 ## Environment details
 <!--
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,9 +1,38 @@
-Please include the following information in your ticket.
+Please [read these guidelines](http://ibm.biz/cdt-issue-guide) before opening an issue.
 
-- sync-android version(s) that are affected by this issue.
-- sync-android artifact you are using
+<!-- Issues will be CLOSED IMMEDIATELY if the following template is not completed. -->
+
+## Bug Description
+
+### 1. Steps to reproduce and a small code sample that demonstrates the issue
```
tomblench: How about something like https://gist.github.com/tomblench/909e5ea69428d1da2cb4c8c40b60b5a5 ?

It's a contrived example, and not actually a bug, but at least it's clearer because it is a concrete example from this library

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,24 +1,86 @@
...
+- [ ] Added tests for code changes _or_ test/build only changes
+- [ ] Updated the change log file (`CHANGES.md`|`CHANGELOG.md`) _or_ test/build only changes
+- [ ] Completed the PR template below:
 
-## What
+## Description
 
-What was changed e.g.
->Corrected NPE when trying to read settings from a migrated index.
+<!--
```
tomblench: Do you still want the user to put a summary as well as the link to the issue? I personally would find this useful, and issue descriptions change over time and often don't capture the underlying bug (ie the GH issue is framed in "when I click x, y happens", but the description here should be "an NPE gets thrown because foo was set to baz")

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,24 +1,86 @@
...
+Be brief: which component(s) of the code base does the fix focus on.
+
+A place to note whether the part of the code base that is being worked is
+particularly sensitive.
+-->
+
+## Schema & API Changes
+
+<!--
```
tomblench: This may be per-library specific, but we need an awareness of what constitutes "API" here. For us it would be something like "anything in com.cloudant.sync but not in com.cloudant.sync.internal". And not test code, which should be obvious.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,24 +1,86 @@
...
+    - test y
+    - test z
+
+OR
+
+- Modified existing tests because ...
+
+OR
+
+- N/A build or packaging only changes
```
tomblench: changes which can only be manually tested? eg specific device etc

PR #582, opened 2018-06-27
Support for doc_ids filter in Pull filtered replication.

### LINKED ISSUE #578: Add support for for _doc_ids filter in Pull replicator
Thanks for your hard work,
Can you please add support for for _doc_ids filter in Pull replicator ?
Ref : http://docs.couchdb.org/en/2.1.1/api/database/changes.html#doc-ids

Thanks
Youcef HILEM

jjrodrig:
## Checklist

- [X] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](../blob/master/DCO1.1.txt)
- [X] Added tests for code changes _or_ test/build only changes
- [X] Updated the change log file (`CHANGES.md`|`CHANGELOG.md`) _or_ test/build only changes
- [X] Completed the PR template below:

## Description

Please, consider this PR that fixes #578

It allow to specify  a list of documents IDs in the Filtered Pull replication builder. The replication will occur only for the set of document IDs provided.

The document IDs are provided as a _doc_ids filter in the _changes API. 
https://console.bluemix.net/docs/services/Cloudant/api/database.html#get-changes

## Approach

The change affects to Pull Replicator (Replicator Factory, PullStrategy,...)

## Schema & API Changes

A new method is added to the ReplicatorBuilder
```
ReplicatorBuilder.pull().
                from(this.couchConfig.getRootUri()).
                to(this.documentStore)
                .docIds(docIds)
```
New method docIds accept a List of Ids represented as Strings

## Security and Privacy
 "No change"

## Testing
- Added new tests:
 at PullReplicatorTest.java
    - testPullReplicationCreatedSuccessfullyWithDocIds()
    - testPullReplicationSelectorAndDocIdIncompatible()
    - testPullReplicationFilterAndDocIdIncompatible()
at PullStrategyDocIdTest.java
   - pull_filterDocIdsFromAnimalDb_twoDocShouldBePulled()
   - pull_filterSelectorMammalFromAnimalDbUsingParameterizedFilter_eightDocShouldBePulled()
at ReplicatorIdTest.java
  - pullWithDocIdsNotEqual()
  - pullWithDifferentDocIdsNotEqual()

## Monitoring and Logging
- "No change"


tomblench:
@jjrodrig sorry for the late response and thanks for your contributions so far.

We've realised that there is an easier way to do this: sub-class `PullFilter`, in this new class set `name` to `_doc_ids`, and set `parameters` to `{"doc_ids": [blah...]}`.

Thinking about it, this is how `selector` should have been done as well. We are keen to deprecate the existing method of using `selector` and use the approach outlined above. This would make the API much cleaner since `PullFilter`s or their sub-classes can be used for all three use cases (doc_ids, selector, or a custom filter).

jjrodrig:
Thanks @tomblench,  we are using this library in different mobile projects. It is great to have the opportunity  to contribute.

I see your proposal of sub-classing PullFilter. Let me rework this PR with this approach.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -452,12 +453,17 @@ public ChangesResult changes(Object since, Integer limit) {
     public ChangesResult changes(PullFilter filter, Object since, Integer limit) {
         Map<String, Object> options = getParametrizedChangeFeedOptions(since, limit);
         if (filter != null) {
-            String filterName = filter.getName();
-            Map filterParameters = filter.getParameters();
-            if (filterName != null) {
-                options.put("filter", filterName);
-                if (filterParameters != null) {
-                    options.putAll(filterParameters);
+            if (filter instanceof PullDocIdsFilter) {
```
tomblench: I don't think it's a good idea to make a special case for this. We can `POST` the options in the body for all `changes` requests

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullFilter.java
```diff
@@ -148,4 +146,10 @@ public String toQueryString() {
             return Misc.join("&", queries);
         }
     }
+
+    protected void setParameters(Map<String, String> parameters) {
+        Map<String, String> internalParams = new HashMap<String, String>();
```
tomblench: Since `parameters` isn't exposed with a getter, we don't need to make it unmodifiable any more. This can be a straightforward setter method.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullDocIdsFilter.java
```diff
@@ -0,0 +1,42 @@
...
+    public static final String DOC_IDS_FILTER_NAME = "_doc_ids";
+
+    public PullDocIdsFilter(List<String> docIds) {
+        super(DOC_IDS_FILTER_NAME);
+
+        Misc.checkState(docIds != null && !docIds.isEmpty(),
+                "Doc Ids should not be empty in a Doc Ids Filter");
+
+        List<String> internalDocIds = new ArrayList<String>(docIds);
+        Collections.sort(internalDocIds);
```
tomblench: Is there any reason for sorting doc ids?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullDocIdsFilter.java
```diff
@@ -0,0 +1,42 @@
...
+
+        Map<String, Object> docIdsParameter = new HashMap();
+        docIdsParameter.put("doc_ids", internalDocIds);
+
+        Map<String, String> parameters = new HashMap();
+        parameters.put("body", JSONUtils.serializeAsString(docIdsParameter));
```
tomblench: I don't really like this idea of unpacking and packing things via the "body" entry in the map. We can just use the map directly when doing the POST and get rid of this extra level of indirection.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -452,26 +453,26 @@ public ChangesResult changes(Object since, Integer limit) {
...
+               String bodyParams = filter.getParameters().get("body");
```
tomblench: See my comment elsewhere, I think it's unnecessary to use this "body" key.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullDocIdsFilter.java
jjrodrig: As we need to build a constant replicationId the sorting allows to keep the replication id constant even if the aplication provides de ids in different order. 

I was not sure if it could be convenient to remove duplicated Ids from the list.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullDocIdsFilter.java
tomblench: Actually looking at the original definition this was probably done to avoid encoding issues when sending the query string. Since we are now using the JSON encoder to send these parameters as in a POST request, <String,Object> is the right choice. I wonder if we can keep a <String,String> setter so that we don't break backwards compatibility?

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
jjrodrig: Ok, checking again the API doc, I see that POST could be used also for standard filters in the _changes endpoint

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullFilter.java
jjrodrig: I'm not sure of this. If the application keeps a reference to the parameter map those parameters can be modified without our control.

jjrodrig: There is a getter over parameters in PullFilter. I think we require the unmodifiable list

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullFilter.java
```diff
@@ -96,9 +96,11 @@ public PullFilter(String filterName) {
      */
     public PullFilter(String filterName, Map<String, String> parameters) {
```
tomblench: This constructor needs to be deprecated and a new constructor `PullFilter(String filterName, Map<String, String> parameters)` needs to be added to allow users to create their own `PullFilter`s with complex arguments in the `POST` body

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullDocIdsFilter.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @see
+ * <a target="_blank" href="https://console.bluemix.net/docs/services/Cloudant/api/database.html#get-changes">See doc_ids filtering</a>
+ */
+public class PullDocIdsFilter extends PullFilter {
+
+    private static final String DOC_IDS_FILTER_NAME = "_doc_ids";
+
+    public PullDocIdsFilter(List<String> docIds) {
+        super(DOC_IDS_FILTER_NAME);
```
tomblench: Instead of calling `super`, just set the filter name directly and then set the parameters. It's not as "pure" but it means you can make `parameters` final in `PullFilter`.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullDocIdsFilter.java
```diff
@@ -0,0 +1,40 @@
...
+        Misc.checkState(docIds != null && !docIds.isEmpty(),
+                "Doc Ids should not be empty in a Doc Ids Filter");
+
+        List<String> internalDocIds = new ArrayList<String>(docIds);
+        Collections.sort(internalDocIds);
+
+        Map<String, Object> parameters = new HashMap<String, Object>();
+        parameters.put("doc_ids", internalDocIds);
+
+        setParameters(parameters);
```
tomblench: OK I can see the value in sorting the doc ids, but there is no need to call the `setParameters` helper. You are creating a `HashMap` temporarily here and then throwing it away when you call `setParameters`. Instead just do

```java
Map<String, Object> internalParameters = new HashMap<String, Object>();
internalParameters.put("doc_ids", internalDocIds);
parameters = Collections.unmodifiableMap(internalParams);
```

tomblench: Note that you will need to make `name` and `parameters` protected in `PullFilter` for the above and for my comment relating to setting `name` directly in the `PullDocIdsFilter` - I think it's a worthwhile compromise to get the `final` correctness right.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullFilter.java
jjrodrig: @tomblench What do you think about change the signature using generics? If we define the constructor as ```PullFilter(String filterName, Map<String, ? extends Object> parameters)``` the API will be backward compatible whereas the internal parameters storage could be changed to a ```Map<String,Object>```

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullDocIdsFilter.java
jjrodrig: It is also required to add a protected default constructor for PullFilter to avoid the use of ```super```

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullFilter.java
tomblench: I was surprised that this works but it looks a bit hacky. I would prefer the original approach of a deprecated constructor and a new one as it makes it clear that we have made a change to the API here.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullDocIdsFilter.java
tomblench: Good point, I had to try this myself to understand what's going on here. OK we can leave as-is.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullFilter.java
jjrodrig: The problem is that we can not deprecate the current constructor ```PullFilter(String filterName, Map<String, String> parameters)``` and create a new one ```PullFilter(String filterName, Map<String, Object> parameters)``` as due to type erasure both constructors will have the same signature. 

The only option is to generalise the current one.

It's true that it looks a bit hacky, but I dont see other mechanism that allows a smooth API evolution in this case. Using Map<String, ?> is a way to generalise the current definition.

tomblench: Good point. It looks like you can just change this to `PullFilter(String filterName, Map<String, Object> parameters) ` and it will not break existing code (binary or source). My understanding is that this works because of the way type erasure is done.

jjrodrig: If we define the constructor as Map<String,Object> we will break the existing source code. We will break the compilation of the existing projects. 

As an example, we get this error compiling the tests:
```
ReplicatorIdTest.java:160: error: incompatible types: Map<String,String> cannot be converted to Map<String,Object>
```
I only see the use of generics as a solution for this

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullDocIdsFilter.java
jjrodrig: It is not clear for me if you want to keep the ```final``` modifier for ```name``` and ```parameters```. In this case we can not set the value for these members from the constructor of a subclass.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullFilter.java
tomblench: OK, go with the generics

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullDocIdsFilter.java
tomblench: OK, yes I can see why we can't keep parameters `final`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullFilter.java
```diff
@@ -96,9 +96,11 @@ public PullFilter(String filterName) {
      */
```
tomblench: javadoc is wrong (and was probably wrong before):
- `"Constructs a filter object for a function that requires no parameters"`: obviously not true
- `@param parameters Any parameters required for the function. Can be {@code null}`: can't be null as we will NPE
- `Integer values should be added as String objects`: this is no longer true since we don't use a query string. I wonder if it's a breaking change?

tomblench:
@jjrodrig turns out I was wrong (again). We can only POST the keys for the doc ids filter - all other filters have to use the current method of GET with key/value pairs on the query string. I need to rethink this and put out a proposal of my own before we can move forward.

Thanks for your efforts so far and sorry this has taken so long!

jjrodrig:
@tomblench I checked CouhDB code and only selectors and doc_ids accepts content in the POST request body. The classic filters require to encode the params in the query string but accepts a POST request with an empty body. 

CouchDB doc states 

> POST /{db}/_changes
> "Requests the database changes feed in the same way as GET /{db}/_changes does, but is widely used with ?filter=_doc_ids ..."

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -480,6 +483,15 @@ public ChangesResult changes(final Map<String, Object> options) {
         return executeToJsonObjectWithRetry(connection, ChangesResult.class);
     }
 
+    private ChangesResult changesRequestWithPost(String filter, String body, Object since, Integer limit) {
```
tomblench: For consistency with `changes(final Map<String, Object> options)`, this should have the signature `changesRequestWithPost(String filter, String body, final Map<String, Object> options` and we should hoist the call to `getParametrizedChangeFeedOptions` into the caller.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
```diff
@@ -465,13 +465,16 @@ public ChangesResult changes(PullFilter filter, Object since, Integer limit) {
...
+
+    public ChangesResult changes(List<String> docIds, Object since, Integer limit) {
+        Misc.checkState((docIds != null && !docIds.isEmpty()), null);
+        Map<String, Object> docIdsMap = new HashMap<String, Object>();
+        docIdsMap.put("doc_ids", docIds);
+        String docsIdsDoc = JSONUtils.serializeAsString(docIdsMap);
+        return changesRequestWithPost("_doc_ids", docsIdsDoc, since, limit);
     }
 
     public ChangesResult changes(final Map<String, Object> options) {
```
tomblench: This version of `changes` should be made private and have its name changed to something like `changesRequestWithGet` to be consistent with its new counterpart `changesRequestWithPost`

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/mazha/CouchClient.java
tomblench: @jjrodrig actually i just realised it can be `changesRequestWithPost(String body, final Map<String, Object> options)` and the caller can put `filter` into the `options`

jjrodrig:
@tomblench no problem, it is a logical maturing process. It is better to be sure that we implement the most adequate solution.

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/replication/PullFilter.java
```diff
@@ -86,7 +85,7 @@ public PullFilter(String filterName) {
      *                   and the name of the filter function, separated by a slash. For example,
      *                   {@code filterDoc/filterFunctionName}
      *
-     * @param parameters Any parameters required for the function. Can be {@code null}. The contents
+     * @param parameters Any parameters required for the function. Can't be {@code null}. The contents
```
tomblench: `Must not be null` would be better here.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/PullStrategyDocIdTest.java
```diff
@@ -0,0 +1,78 @@
...
+        PullStrategy replicator = super.getPullStrategy(docIds);
+
+        Assert.assertEquals(0, datastore.getDocumentCount());
+
+        AnimalDb.populate(remoteDb.couchClient);
+        this.pull(replicator, 2);
+
+        Assert.assertEquals(2, datastore.getDocumentCount());
+        String[] birds = {"snipe", "kookaburra"};
+        for (String bird : birds) {
```
emlaver: What about removing the `birds` object and just iterating over `docIds`?
```
for (String bird : docIds) { ... }
```

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/PullStrategyDocIdTest.java
```diff
@@ -0,0 +1,78 @@
...
+                "panda", "zebra");
+        PullStrategy replicator = super.getPullStrategy(docIds);
+
+        Assert.assertEquals(0, datastore.getDocumentCount());
+
+        AnimalDb.populate(remoteDb.couchClient);
+        this.pull(replicator, 8);
+
+        Assert.assertEquals(8, datastore.getDocumentCount());
+        String[] mammals = {"aardvark", "badger", "elephant", "giraffe", "lemur", "llama",
```
emlaver: See comment above.

tomblench:
+1, I will squash and merge

viskp:
Hi Team,

I need to run 'in' clause on a database from android, so need the code for that. i'm all ready doing it for "lessthan"as 


`
Map<String, Object> query = new HashMap<>();
query.put(key, new HashMap<String, Object>() {

                        {
                            put("$eq", "Test");
                        }
                    });`
I need same for the "$in" clause. please help me out

emlaver:
@viskp Can you please open a new issue? There are guidelines (.e.g. steps to reproduce with sample code) when you open a issue that will assist our team on investigating your inquiry. Thanks.

PR #600, opened 2020-01-17
Fix #599 - Record checkpoint on empty _changes result

### LINKED ISSUE #599: Record checkpoint when _changes result is empty in pull replications
Please [read these guidelines](http://ibm.biz/cdt-issue-guide) before opening an issue.

## Bug Description

 Filtered pull replications does not update checkpoint if the result of _changes is empty. 

 If the remote database has changes that do not match with the filter the last seq is not recorded in the local checkpoint and the same set of changes are reprocessed in each synchronization cycle until a new change match the filtering criteria. 

When the remote database has frequent changes but very few match the filtering criteria this issue can derive in a performacen problem. 
 
### 1. Steps to reproduce and the simplest code sample possible to demonstrate the issue

 - Create a remote database with some docs
 - Configure a filtered pull based on a selector that evaluate to false (e.g. a non existent doc id)
 - Run the pull replication 
 - Check that the checkpoint is not generated in the local documentstore
 - Run again the pull replication (the since param is not incluided in the _changes request)
  
### 2. What you expected to happen

 The lastseq should be stored in the local checkpoint even if the changes result is empty

### 3. What actually happened

 The lastseq is not stored whe changes are empty

## Environment details
Version 2.4.0

- sync-android artifact you are using:
    - [X] cloudant-sync-datastore-android
    - [ ] cloudant-sync-datastore-android-encryption
    - [X] cloudant-sync-datastore-javase

jjrodrig:
## Checklist

- [X] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](../blob/master/DCO1.1.txt)
- [X] Added tests for code changes _or_ test/build only changes
- [X] Updated the change log file (`CHANGES.md`|`CHANGELOG.md`) _or_ test/build only changes
- [X] Completed the PR template below:

## Description

Filtered pull replications does not update checkpoint if the result of _changes is empty.

Fixes #599 

## Approach
Change is located in the PullStrategy.java where the case of empty changes result is considered for the checkpoint creation.

## Schema & API Changes
- "No change"

## Security and Privacy
- "No change"

## Testing
- Added new test:
   - PullStrategyMockTest.testSetCheckpointWhenEmptyChanges

## Monitoring and Logging
- "No change"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
 # 2.x.y (Unreleased)
-- [IMPROVED] Record checkpoint on empty `_changes` result in Pull replications. This change optimizes 
+- [IMPROVED] Record checkpoint on empty `_changes` result in pull replications. This change optimizes 
    filtered replications when changes in remote database doesn't match the replication filter. 
 # 2.4.0 (2019-01-15)
 - [NEW] `Database` methods `read`, `contains`, `create`, and `delete` now accept local
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+# 2.x.y (Unreleased)
+- [IMPROVED] Record checkpoint on empty `_changes` result in Pull replications. This change optimizes 
+   filtered replications when changes in remote database doesn't match the replication filter. 
 # 2.4.0 (2019-01-15)
 - [NEW] `Database` methods `read`, `contains`, `create`, and `delete` now accept local
   (non-replicating documents). These documents must have their document ID prefixed with `_local/`
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
@@ -262,6 +262,13 @@ public class PullStrategy implements ReplicationStrategy {
             if (changeFeeds.size() > 0) {
                 batchChangesProcessed = processOneChangesBatch(changeFeeds);
                 state.documentCounter += batchChangesProcessed;
+            } else {
+                try {
+                    this.targetDb.putCheckpoint(this.getReplicationId(), changeFeeds.getLastSeq());
+                } catch (DocumentStoreException e) {
+                    logger.log(Level.WARNING, "Failed to put checkpoint doc, next replication will " +
+                            "start from previous checkpoint", e);
+                }
             }
 
             long batchEndTime = System.currentTimeMillis();
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
@@ -473,7 +473,6 @@ public class PullStrategy implements ReplicationStrategy {
     }
 
     private ChangesResultWrapper nextBatch(final Object lastCheckpoint) {
-//        final Object lastCheckpoint = this.targetDb.getCheckpoint(this.getReplicationId());
         logger.fine("last checkpoint " + lastCheckpoint);
 
         ChangesResult changeFeeds = null;
--- a/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
+++ b/cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
@@ -236,9 +236,10 @@ public class PullStrategy implements ReplicationStrategy {
         }
 
         this.state.documentCounter = 0;
+
         while (!this.state.cancel) {
             this.state.batchCounter++;
-
+            final Object lastKnownCheckpoint = this.targetDb.getCheckpoint(this.getReplicationId());
             String msg = String.format(
                     "Batch %s started (completed %s changes so far)",
                     this.state.batchCounter,
@@ -247,7 +248,7 @@ public class PullStrategy implements ReplicationStrategy {
             logger.info(msg);
             long batchStartTime = System.currentTimeMillis();
 
-            ChangesResultWrapper changeFeeds = this.nextBatch();
+            ChangesResultWrapper changeFeeds = this.nextBatch(lastKnownCheckpoint);
             int batchChangesProcessed = 0;
 
             // So we can check whether all changes were processed during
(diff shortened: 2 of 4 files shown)


fixture/empty_changes.json
```diff
@@ -0,0 +1,5 @@
+{
+    "last_seq": "5-g1AAAAIreJyVkEsKwjAURZ...",
```
emlaver: Let's use an example revision without the `...` like `10-d9e5b0147af143e5b6d1979378ad957b`

fixture/empty_changes.json
```diff
@@ -0,0 +1,5 @@
+{
+    "last_seq": "5-g1AAAAIreJyVkEsKwjAURZ...",
+    "pending": 0,
+    "results": []
+}
```
emlaver: Add newline

CHANGES.md
```diff
@@ -1,3 +1,6 @@
+# 2.x.y (Unreleased)
+- [IMPROVED] Record checkpoint on empty `_changes` result in Pull replications. This change optimizes
```
emlaver: Change `Pull` to lowercase

jjrodrig:
Thanks @emlaver for the review. I just updated the PR.

bessbd:
Thank you for the patch, @jjrodrig ! Unfortunately, master was failing which has been fixed recently. Can you please rebase your PR on master?

jjrodrig:
Thanks @bessbd. Done.

bessbd:
Thank you, @jjrodrig . Looks good to me!

cloudant-sync-datastore-core/src/main/java/com/cloudant/sync/internal/replication/PullStrategy.java
```diff
@@ -262,6 +262,13 @@ private void replicate()
             if (changeFeeds.size() > 0) {
                 batchChangesProcessed = processOneChangesBatch(changeFeeds);
                 state.documentCounter += batchChangesProcessed;
+            } else {
```
ricellis: I'm a little concerned about this being in an `else` block without additional checks. Whilst this works for your golden path case of having some new changes (all filtered) and thus seeing 0 changes, but having a new `lastSeq` I think there are other cases where it is going to force the writing of a checkpoint even though there has been no change to the `lastSeq`.
For example, in cases of 0 changes received (with no filtering) or an integer multiple of `changeLimitPerBatch` changes this else will cause writing the same checkpoint/lastSeq we already know once more again before breaking out of the loop.

I don't think this concern is definitely a blocker for this change, but I think we ought to at least consider the potential impact of writing those redundant checkpoints vs the complexity of moving the retrieval of the current known `lastSeq` out of `processOneChangesBatch` so that we could validate if it changed before writing the checkpoint in something like an `else if (!knownLastSeq.equals(changeFeeds.getLastSeq())) {...}`.

jjrodrig: Thanks @ricellis for your review. 

I've reworked this PR including your proposal. The current checkpoint is retrieved on the main loop and setted back only it changed. I agree with you that it is better to protect the checkpoint from redundant updates. 

I also added some tests related to checkpoint recording. I didn't find many tests related to checkpoints.

ricellis: Thanks, this looks really good.
I'm running our on-device tests then hopefully we'll be able to proceed.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/PullStrategyMockTest.java
```diff
@@ -231,6 +232,97 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
...
+        PullStrategy pullStrategy = super.getPullStrategy();
+        pullStrategy.sourceDb = mockRemoteDb;
+        pullStrategy.targetDb = mockLocalDb;
+        pullStrategy.getEventBus().register(mockListener);
+        pullStrategy.run();
+
+        //make sure the correct events were fired
+        verify(mockListener).complete(any(ReplicationStrategyCompleted.class));
+        verify(mockListener, never()).error(any(ReplicationStrategyErrored.class));
+        verify(mockLocalDb, never()).putCheckpoint(anyString(), any());
```
ricellis: This test failed on device and from some investigations it seems that mocking of a package-private class doesn't work with mockito in Dalvik because of some classloader differences (see e.g. https://github.com/mockito/mockito/issues/796 and https://github.com/janstk/dexmaker/issues/26#issuecomment-142492955).
Given that this is a mock test anyway I don't think we're gaining anything by running it on device since it runs elsewhere anyway, so I suggest just using an assume to skip the test if mocking fails e.g.:
```java
        try {
            DatastoreWrapper mockLocalDb = mock(DatastoreWrapper.class);
            when(mockLocalDb.getCheckpoint(any(String.class))).thenReturn("10" +
                    "-d9e5b0147af143e5b6d1979378ad957b");

            StrategyListener mockListener = mock(StrategyListener.class);
            PullStrategy pullStrategy = super.getPullStrategy();
            pullStrategy.sourceDb = mockRemoteDb;
            pullStrategy.targetDb = mockLocalDb;
            pullStrategy.getEventBus().register(mockListener);
            pullStrategy.run();

            //make sure the correct events were fired
            verify(mockListener).complete(any(ReplicationStrategyCompleted.class));
            verify(mockListener, never()).error(any(ReplicationStrategyErrored.class));
            verify(mockLocalDb, never()).putCheckpoint(anyString(), any());
        } catch(UnsupportedOperationException uoe) {
            assumeNoException("Cannot proxy DatastoreWrapper on Dalvik", uoe);
        }
```

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/PullStrategyMockTest.java
```diff
@@ -231,6 +232,97 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
...
+        PullStrategy pullStrategy = super.getPullStrategy();
+        pullStrategy.sourceDb = mockRemoteDb;
+        pullStrategy.targetDb = mockLocalDb;
+        pullStrategy.getEventBus().register(mockListener);
+        pullStrategy.run();
+
+        //make sure the correct events were fired
+        verify(mockListener).complete(any(ReplicationStrategyCompleted.class));
+        verify(mockListener, never()).error(any(ReplicationStrategyErrored.class));
+        verify(mockLocalDb).putCheckpoint(anyString(), eq("10-d9e5b0147af143e5b6d1979378ad957b"));
```
ricellis: As above, `try`/`catch` with an `assumeNoException` to skip this when the class can't be proxied.

cloudant-sync-datastore-core/src/test/java/com/cloudant/sync/internal/replication/PullStrategyMockTest.java
jjrodrig: Hi @ricellis, I've included your propoal and tested on the emulator, now both tests are skipped.
```
com.cloudant.sync.internal.replication.PullStrategyMockTest > testSetCheckpointWhenModified[Nexus_4_API_23(AVD) - 6.0] SKIPPED 
  org.junit.internal.AssumptionViolatedException: Cannot proxy DatastoreWrapper on Dalvik: got: <java.lang.UnsupportedOperationException: can
not proxy inaccessible class class com.cloudant.sync.internal.replication.DatastoreWrapper>, expected: null
        at org.junit.Assume.assumeThat(Assume.java:118)

com.cloudant.sync.internal.replication.PullStrategyMockTest > testSetCheckpointWhenModified[Nexus_4_API_23(AVD) - 6.0] SKIPPED 
    org.junit.internal.AssumptionViolatedException: Cannot proxy DatastoreWrapper on Dalvik: got: <java.lang.UnsupportedOperationException: cann
ot proxy inaccessible class class com.cloudant.sync.internal.replication.DatastoreWrapper>, expected: null
        at org.junit.Assume.assumeThat(Assume.java:118)
```

Thanks

